PR #4, opened 2016-12-11
Dev Shutter

twistermcneal:
Start shutter movment with bus event implementing callback_after_update

### DIFF
--- a/home-assistant-plugin/custom_components/cover/xknx.py
+++ b/home-assistant-plugin/custom_components/cover/xknx.py
@@ -42,7 +42,12 @@ class XKNX_Cover(CoverDevice):
         self._closing_tilt = True
         self._unsub_listener_cover = None
         self._unsub_listener_cover_tilt = None
+        self.register_callbacks()
 
+    def register_callbacks(self):
+        def after_update_callback(device):
+            self.update()
+        self.device.after_update_callback = after_update_callback
     @property
     def name(self):
         """Return the name of the cover."""
@@ -50,8 +55,8 @@ class XKNX_Cover(CoverDevice):
 
     @property
     def should_poll(self):
-        """No polling needed for a demo cover."""
-        return False
+        """ polling cover."""
+        return True
 
     @property
     def current_cover_position(self):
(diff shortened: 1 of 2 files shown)


Julius2342:
Can you explain what you did? Then I will port your changes in my branch ...

(looks like I dont really understand the report back mechanism)

twistermcneal:
i implementet the callback and used the update funktion to get the bus event with this i trigger the up or down movement... in my case some shutters dont have a position value so i think the time funktion is a way to stop the cover at a position but  in this case we need a config parameter how long the cover runs from open to close

Julius2342:
Ok, i see, so it looks like it gets more complicated ...

(My actuator has an explicit group address for requesting the position)

twistermcneal:
can your actuator also move to a given position?

Julius2342:
Yes, but this is a different group address

Julius2342:
btw, i added you on skype. Dont know if you use this :)

twistermcneal:
havent it installed at the moment but yes sometimes ....

twistermcneal:
I think i go with my own shutter class for the moment, so we can think later about adding both functions !?

Julius2342:
My goal is to write a library which is suitable for all common KNX products.

Which shutter actuator do you have? Doesn't it support an absolute shutter position group address? 

(I run **GIRA KNX/Shutter Binary Actuator 103800**,  in my system the group address is named "Shutter/Awning position")

twistermcneal:
this one is a Busch Jaeger 6196/43-102 4x Shutter Actuator KNX,
no it does not support an absolute shutter position, only Move Up/down and Stop/tilt.

Julius2342:
Hmm, ic ... *think*.

I think we should do somehting like:

````
    class Direction(Enum):
        UP = 1
        DOWN = 2


    def set_position(self, position):
        if self.group_address_position is not None:
            self.send(self.group_address_position, [0x80, position])
        else
            self.set_position_relative(position)

    dev set_position_relative(self, position):
        relative_position = position - self.position
        direction = Direction.UP if relative_position > 0 else Direction.DOWN

        travel_time_full = travel_time_up if direction == Direction.UP else travel_time_up_down # should be taken from configuration ..

        travel_time = travel_time_full * abs(relative_position) / 100

        if direction == Direction.UP:
           self.set_up()
        else
           self.set_down()

       time.sleep(travel_time)

       self.set_short_down()
````

PR #7, opened 2016-12-14
Update address.py

twistermcneal:
also support 1/2/0 address format


twistermcneal:
Ah nice :) i close this

twistermcneal:
i still learning to interact correctly with github ... is it not possible to undo or override a commit with the latest source from your repo

Julius2342:
Well, you can revert commits, but i think the easier way is to delete the branch and re-checkout master.

You you have any other commits in that branch?

PR #13, opened 2016-12-18
Cover without absolute position t5

twistermcneal:
after some debugging  and testing i think i got it...


twistermcneal:
timeing is very good ... good work !!!!
the only thing missing is the trigger from the bus to start/stop the movement.

PR #16, opened 2016-12-22
Update address.py

twistermcneal:
Bit format change to 0mmm msss gggg gggg


Julius2342:
*sigh* you are right :-)

PR #41, opened 2017-06-12
Sanatized all corresponding __str__ to return wellformed XML for bett…

GeorgConradi:
…er extensibility.


Julius2342:
Thank you very much :-)

PR #47, opened 2017-08-01
Logging class

Julius2342:
@elupus , @daBONDi : Is this how you would expect a logging class?


elupus:
Yea something like that. The logging class could have been imported in each file on a global level thou. With same name, but this works fine too.

PR #66, opened 2017-09-22
Add AUTO mode to HVAC

buergi:
According to KNX specification v2.1 3/7/2 Datapoint Types, section 4.3, DPT 20.102 value 0 is "Auto"

This value is used for example when the climate control is able to use a presence detector which governs the HVAC type.

I have however no idea how the AUTO mode is coded in this strange "non-standardised DP type", maybe `if raw[0] & 15 == 0` but I don't know it for sure.

### DIFF
--- a/test/climate_test.py
+++ b/test/climate_test.py
@@ -289,6 +289,7 @@ class TestClimate(unittest.TestCase):
             group_address_controller_status='1/2/4')
 
         for operation_mode in HVACOperationMode:
+            if operation_mode == HVACOperationMode.AUTO: continue
             self.loop.run_until_complete(asyncio.Task(climate.set_operation_mode(operation_mode)))
             self.assertEqual(xknx.telegrams.qsize(), 1)
             telegram = xknx.telegrams.get_nowait()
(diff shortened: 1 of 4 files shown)


resoai:
There is definitely no "auto" mode in Jung controllers.

TinusK:
MDT doesn't have it either :(

buergi:
Yes there is an "Auto" mode, here is extract from the KNX specification, v2.1 3/7/2 Datapoint Types, section 4.3, DPT 20.102
hvac

I use the "Auto" setting with my Busch-Jaeger [6131/21-24](http://www.busch-jaeger-catalogue.com/2CKA006132A0344,artikel.html)

It has a _Working mode_ object (11.19.12 in the manual) which allows to configure the values Comfort/Standby/Economy/Frost protection and a _Superimposed operating mode_ object (11.19.13 in the manual) which additionally allows the Auto (Superimposition inactive) setting.
The idea is to have the presence detector output object coupled (in the same group) with the _Working mode_ object, so that it automatically sets the operation mode to Comfort when I'm at home and back to Eco when I'm gone.
The _Superimposed operating mode_ allows me to override this behavior to a fixed mode, which is independent of by presence, by setting it to anything other than Auto.

So in short: Auto in my setup means comfort when I'm at home, otherwise eco.

buergi:
Regarding the strange non-standardized DPT there is information about the coding in the some manuals, for example [here](http://download.gira.de/data2/20693390.pdf) page 91:
grafik

This value does not hold any information about the "auto" mode, maybe this information is somehow encoded in bit 0 of the additional status telegram, also of non-standardized DPT:
grafik

So I think we have to go the way you proposed by announcing the Auto value only for `operation_mode` not for `controller_status`.

... und, ja bin aus München.

Julius2342:
@buergi : Thank you very much :)

Made a small change: 9fa4b314af3902b9b682c3e24935fd5790dd5464 and merged it.

PR #80, opened 2017-11-09
Rename RemoteValue object to Group

Julius2342:
As discussed yesterday in Discord, RemoteValue is a bad name. Lets rename it to Group.


Julius2342:
@encbladexp : May you review?

PR #83, opened 2017-12-03
Split Address into two groups.

encbladexp:
Currently we use `Address()` for just everything address related. This results in a lot of unstructured code and is not really python like.

This change splits `Address()` in `GroupAddress` and `PhysicalAddress` to honor the specials of these types much better. Later on we should provide more helpers, like `is_valid`.

After all, we should think about uninitialized Addresses in xknx at all, to avoid potential fuckups as 0 is the current default address in some parts.

This changes breaks compatibility with existing third party software, if they use `Address`. I think its time to create some CHANGELOG.md for the next release.

### DIFF
--- a/examples/example_tunnel.py
+++ b/examples/example_tunnel.py
@@ -3,7 +3,7 @@ import asyncio
 
 from xknx import XKNX
 from xknx.io import GatewayScanner, Tunnel
-from xknx.knx import Address, DPTBinary, Telegram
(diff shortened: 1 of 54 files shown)


Julius2342:
great! thank you! i will review asap!

Julius2342:
I merged it to splitaddresses, will do some tests and then merge to dev/master...

PR #85, opened 2018-01-02
Requirements Update 2018-01-02

encbladexp:
Some small updates on our requirements.

### DIFF
--- a/xknx/knx/address_filter.py
+++ b/xknx/knx/address_filter.py
@@ -48,12 +48,15 @@ class AddressFilter:
         """Test if provided address matches Addressfilter."""
         if isinstance(address, str):
             address = Address(address)
+
         if len(self.level_filters) == 3:
             return self._match_level3(address)
         elif len(self.level_filters) == 2:
             return self._match_level2(address)
         elif len(self.level_filters) == 1:
             return self._match_free(address)
+
+        return False
 
     def _match_level3(self, address):
         return (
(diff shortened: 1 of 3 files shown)


xknx/knx/dpt_hvac_mode.py
```diff
@@ -37,6 +37,8 @@ def from_knx(cls, raw):
         elif raw[0] == 0x00:
             return HVACOperationMode.AUTO
 
+        return None
```
Julius2342: Shouldn't we throw an exception here if we can't parse the KNX value? e.g. `raise CouldNotParseKNXIP("Could not parse HVACOperationMode")`

xknx/knx/address_filter.py
```diff
@@ -48,13 +48,16 @@ def match(self, address):
...
             address = Address(address)
+
         if len(self.level_filters) == 3:
             return self._match_level3(address)
         elif len(self.level_filters) == 2:
             return self._match_level2(address)
         elif len(self.level_filters) == 1:
             return self._match_free(address)
 
+        return None
```
Julius2342: Why not a `return false` bc we can't match ...

xknx/knx/dpt_hvac_mode.py
```diff
@@ -76,6 +80,8 @@ def from_knx(cls, raw):
         elif raw[0] & 1 > 0:
             return HVACOperationMode.COMFORT
 
+        return None
```
Julius2342: same here, why not raise an Exception ...

encbladexp:
I updated this PR.

PR #86, opened 2018-01-02
Implemented datetime and date, renamed day to weekday in Time, renamed Time to DateTime

Roemer:


xknx/knx/dpt_datetime.py
```diff
@@ -0,0 +1,103 @@
...
+            return False
+        if day < 1 or day > 31:
+            return False
+        if weekday < 0 or weekday > 7:
+            return False
+        if hours < 0 or hours > 23:
+            return False
+        if minutes < 0 or minutes > 59:
+            return False
+        if seconds < 0 or seconds > 59:
```
encbladexp: Keep in mind that leap seconds exists.

xknx.yaml
```diff
@@ -86,9 +86,9 @@ groups:
         Attic.Climate: {group_address_temperature: '1/7/7', group_address_operation_mode_protection: '1/7/8', group_address_operation_mode_night: '1/7/9', group_address_operation_mode_comfort: '1/7/10'}
         Cellar.Climate: {group_address_temperature: '1/7/11', group_address_controller_status: '1/7/12', group_address_controller_status_state: '1/7/13'}
 
-    # Cyclic sending of time to the KNX bus
-    time:
-        General.Time: {group_address: '2/1/2'}
+    # Cyclic sending of datetime to the KNX bus
```
encbladexp: Date and Time are different datatypes in the knx world. Most sensors/actors have two groups for date and time, i never seen a combined datetime group in the real world.

xknx/knx/dpt_datetime.py
Roemer: According to the KNX Specification, a valid range excludes an eventuel leap second. See:
image

xknx.yaml
Roemer: there are actually three types:
* Time (10.001, 3 octets)
* Date (11.001, 3 octets)
* DateTime (19.001, 8 octets)
With my change, you are able to choose which of those you need.
In real world, the [MDT Glastaster II Smart](http://www.mdt.de/Glastaster_Smart.html) is one of the KNX devices that needs a DateTime to work.

Roemer:
I rebased to dev, maybe I should have merged instead because of the review. But the comments are still visible so it should be ok :)

encbladexp:
Rebasing an PR is most times besser than merging from other branches. Nice to learn this new Datapoint Type.

xknx/knx/dpt_datetime.py
encbladexp: Nice to know, so we can skip these too.

PR #88, opened 2018-01-08
Add electric sensors

JohanElmis:
Work on more DPT to support Electrical Sensors made by user Straeng.
This included 4-byte singed and unsigned values, as well as 4-byte floats.

I added some more DPT's and documentation.
We have also added some generic types that supports the "conversion"  - but doesn't necessary have the proper Min/Max limits - and are missing Unit.
Get it merged in so we don't get to far away from the Dev-branch.


Julius2342:
Thank you very much!

PR #96, opened 2018-01-20
Example of KNX to MQTT forwarder.

JohanElmis:


JohanElmis:
Fixed the PEP warnings. But I think it misinterpreted some of my variables. Complains about Contants (which they are not) - and globals. This is not fully working code, but we have had several requests for this kind of example for the last few days.

examples/example_powermeter_mqtt.py
```diff
@@ -0,0 +1,193 @@
...
+    import sys
+    import re
+    # import time
+    import paho.mqtt.client as mqtt
+    # The following library is not included.
+    from myhouse_sensors_mqtt import SensorClientMqtt
+    from myhouse_sensors_mqtt import MetricType
+    # import pprint
+except ImportError as import_err:
+    err_list = str(import_err).split(' ')
```
Julius2342: I would add a `# pylint disable=invalid-name` here. Pylint thinks it is a constant, but it obviously isn't...

examples/example_powermeter_mqtt.py
```diff
@@ -0,0 +1,193 @@
...
+# It also triggers and send values to the MQTT bus if the values has changed.
+# With no changes it can go up to max_interval_seconds before it's sent.
+# This allows me to fetch fast changes without storing all data at a 15s interval.
+mh_sensor = SensorClientMqtt(change_trigger_percent=5,
+                             max_interval_seconds=600,
+                             metric_class='sensor',
+                             debug=True)
+
+# Pre-compile some regexp filters that will be used to catch the different types.
+re_meter_reading = re.compile('MeterReading_')
```
Julius2342: Isn't this a global variable and should be renamed to RE_METER_READING ?

Julius2342:
Thank you! 🐬

Issue #105, opened 2018-01-28
position_state_address fails with old KNX devices


MrChrisCool:
Hi !

My Knx cover device is [Hager TS225](https://www.hager.it/files/download/0/954_1/0/TS225_TS226.pdf) with application program [TK338](https://www.hager.pt/files/download/0/2901_1/0/TS223_225_ETS_FR.pdf)

I have the following error in home-assistant.log

_2018-01-28 18:06:47 ERROR (MainThread) [xknx.log] Exception while processing telegram: <CouldNotParseTelegram description="" />
Traceback (most recent call last):
  File "/srv/homeassistant/lib/python3.5/site-packages/xknx/core/telegram_queue.py", line 99, in process_telegram
    yield from self.process_telegram_incoming(telegram)
  File "/srv/homeassistant/lib/python3.5/site-packages/xknx/core/telegram_queue.py", line 130, in process_telegram_incoming
    yield from device.process(telegram)
  File "/srv/homeassistant/lib/python3.5/site-packages/xknx/devices/cover.py", line 229, in process
    position_processed = yield from self.position.process(telegram)
  File "/srv/homeassistant/lib/python3.5/site-packages/xknx/devices/remote_value.py", line 76, in process
    raise CouldNotParseTelegram()
xknx.exceptions.exception.CouldNotParseTelegram: <CouldNotParseTelegram description="" />_

It's because position_state_address in xknx is of type 5.001 and my old knx device send its position state with type 1.001 (open or close...)

The question is :
Is it possible to modify position_state_address to handle type 1.001 and 5.001 ?
for example 00(1.001) -> 0%(5.001) and 01(1.001) -> 100%(5.001)

Many thanks !
Chris

MrChrisCool:
Hmm... I just see that Issue #68 should resolve this one...

jncanches:
Is there any news forthis issue ?
I am in the same case with a cover state type in 1.001

Many thanks !

JN

Julius2342:
Not really, Im not happy with #107 . We should solve this the right way and introduce the correct DPT abstraction to handle these devices.

Julius2342:
@jncanches : may you jump in and implement it? :-)

MrChrisCool:
Hi jncanches, No news for this issues, but you can use the hack #107 to solve your problem, it will not break your covers as a 1.00x can always be converted to a 5.00x value.

But Julius2342 is right, we must provide a correct implement.

[...]

Issue #68, opened 2017-09-25
Support for generic KNX Datapoints


flo-wer:
Hello,

it seems this KNX library doesn't support setting arbitrary datapoints for each sensor but instead uses a fixed mapping between the sensor type and the datapoint. As the KNX standard defines all possible data types, it would be nice to support using just the plain datapoint as type:
```
  - platform: knx
    name: Kitchen.Temperature
    address: '6/2/1'
    type: '9.001'
```

Currently I have sensors with the following types that aren't supported:
7.013 DPT_Brightness
9.006 DPT_Value_Pres
9.008 DPT_Value_AirQuality
9.028 DPT_Value_Wsp_kmh
14.033 DPT_Value_Frequency
14.056 DPT_Value_Power

Also the current defined types aren't unique, e.g. both 9.005 DPT_Value_Wsp and 14.065 DPT_Value_Speed use m/s as unit.

iditude:
+1 on this. I use KNX to monitor my power consumption and it uses a data type 13.010 active power (in W/h)

This would be super helpful!

encbladexp:
Sound like a good idea for me too.

JohanElmis:
For sensors I think we have this covered in a Development fork. In for review and merge with upstream development in the near future.

Julius2342:
Is there anything missing here?

flo-wer:
Mhh... after looking at the patch from @JohanElmis, I'm not really convinced how this changes the way the sensor type is defined. After all, only additional Data Points were added, but I thought of a way to have all Data Points covered.
From looking at the code, I think that using a new class for each Data Point doesn't scale very well. Wouldn't it be better to just use a list for all Data points?

Something I quickly hacked together:
```
class DPT2ByteFloat(DPTBase):

	def __init__(self, value_min = -671088.64, value_max = 670760.96, unit = '', resolution = 1, payload_length = 2):
		self.value_min = value_min
		self.value_max = value_max
		self.unit = unit
		self.resolution = resolution
		self.payload_length = payload_length


DPT_MAP = {
	'9.xxx': DPT2ByteFloat(None,	None,	''),	# generic DPT2ByteFloat
	'9.001': DPT2ByteFloat(-273,	670760,	'°C'),	# DPT_Value_Temp
	'9.002': DPT2ByteFloat(-670760,	670760,	'K'),	# DPT_Value_Tempd
[...]

PR #107, opened 2018-01-31
Solve old KNX Devices sending position state in 1.00x DTP

MrChrisCool:
First time i use Git and python...
Hope i do things well and it will help...

### DIFF
--- a/examples/example_cover.py
+++ b/examples/example_cover.py
@@ -1,0 +1,26 @@
+"""Example for getting position state from cover."""
+import asyncio
+
+from xknx import XKNX
+from xknx.devices import Cover
+
+
+async def main():
+    """Connect to KNX/IP bus, test cover."""
+    xknx = XKNX()
+    await xknx.start()
+    cover = Cover(xknx,
+                  name='TestCover',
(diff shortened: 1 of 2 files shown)


Julius2342:
mhmmm ... this looks like a hack... we should solve it correctly.

MrChrisCool:
All of my knx devices answer a 1.009 (open/close) response when i ask them for a position state.
Issue [#105](https://github.com/XKNX/xknx/issues/105)
Actually xknx raise an error because xknx expect 5.001
I just modify the code in order to work properly for my devices... I am sure there is a lot of knx devices still with this behavior... But i am not a programmer, i ve done it the fast way...

The correct way (and the long one) should be solving the issue [#68](https://github.com/XKNX/xknx/issues/68)

feel free to ask...

Julius2342:
@MrChrisCool I think we should implement this right and implement a RemoteValuexxxx1009. with good unit test coverage bc otherwise I dont know if any future change might break your covers :) 

```
self.position = RemoteValueScaling5001() if xxx else RemoteValuexxxx1009()
```

PR #110, opened 2018-02-15
Fix multicast binding on Darwin/Mac OS

DrMurx:
On Darwin/Mac OS (formerly known as OS X), the default multicast behaviour of the network interfaces is different from Linux. The socket has to bind to the interface's IP instead of the multicast IP. 

I found `UDPClient` already prepared for this case, but the corresponding option isn't passed beyond class `Routing`.

This PR allows `KNXIPInterface` to take this option as part of the `ConnectionConfig` object. When using `ConnectionType.AUTOMATIC`, the system platform is checked and the option is set accordingly.


Julius2342:
Uh nice! Thank you very much!

PR #111, opened 2018-02-19
Scan multiple gateways

DrMurx:
The gateway scanner currently stops as soon as the first gateway responds which prevents scan for a gateway with specific features, or to scan a larger topology.

This PR changes the interface of the gateway scanner to return multiple gateways as a list of GatewayDescriptor objects. A class GatewayScanFilter has been introduced to limit the scan for certain aspects of a gateway, such as name or routing/tunnelling support, and it is possible to stop after a certain number of gateways has been found (usually 1).

For higher level use, a GatewayScanFilter can be passed to KNXIPInterface via ConnectionConfig.

## Notes:

* All newly written code uses PEP 484 and PEP 526 annotations for better IDE support.
* This PR bases on !110 which should be merged first.


Julius2342:
Thank you very mcuh! Sorry for taking so long for reviewing.

PR #122, opened 2018-05-08
Add support for KNX DPT 20.105

cian:
I have added support for DPT 20.105.
This class introduces many more operation modes, therefore I have added a configuration option so that the configuration file can specify a list of supported modes as follows:

```
override_supported_operation_modes:
 - HEAT
 - COOL
 - DRY
 - FAN_ONLY
```


cian:
Python `3.7-dev` and `nightly` are failing as follows:
` Failed building wheel for pyyaml`
which doesn't seem related to my changes.
All other builds and tests are passing.

Julius2342:
thank you! Will you also make a PR within home-assistant?

Issue #94, opened 2018-01-16
XKNX UDP Error killing HASS log


FireFrei:
Hey, I got a problem :-)

I use xknx with Hass and it is filling up my hass logs with the following error message:

```
...
During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "uvloop/cbhandles.pyx", line 52, in uvloop.loop.Handle._run
  File "uvloop/handles/udp.pyx", line 60, in uvloop.loop.UDPTransport._on_read_ready
  File "/usr/local/lib/python3.6/site-packages/xknx/io/udp_client.py", line 59, in error_received
    self.xknx.logger.warning('Error received: %s', exc)
AttributeError: 'UDPClientFactory' object has no attribute 'xknx'
2018-01-16 22:46:20 ERROR (MainThread) [homeassistant.core] Error doing job: Exception in callback UDPTransport._on_read_ready
Traceback (most recent call last):
  File "uvloop/handles/udp.pyx", line 56, in uvloop.loop.UDPTransport._on_read_ready
OSError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:
...
```

I just deleted the hass log file with about 65GB of spam. This filled up my servers hard disk :-D

My xknx config in hass is pretty minimalistic:
```
knx:
  tunneling:
    host: '192.168.1.66'
    port: 3671
    local_ip: '192.168.1.60'
```

I've noticed the problem in Hass Version 0.60 and 0.61 

Can you help me with this issue? Cheers :-)

FireFrei:
I'm using the official HASS docker image that ships Python 3.6.4 and xknx 0.7.18

Julius2342:
Hmm,

what I don't understand, how UDPClientFactory can be initialized without having the `self.xknx` object being initialized:

https://github.com/XKNX/xknx/blob/master/xknx/io/udp_client.py#L42

Any ideas?

(There is also another issue for a better handling of UDP errors ... https://github.com/XKNX/xknx/issues/89 ).

encbladexp:
Maybe, this issue has another source, and not the exception we see now. Maybe this exception is the result of an different issue…

janitza-bjag:
Hi,

i have the same issue with HA 0.62.1 used in docker with https://hub.docker.com/r/homeassistant/raspberrypi3-homeassistant/ image

The "error" seems to appear randomly after a few hours/days of running HA. When it happens the log file size crashed the docker image ...

any hints appreciated!

[...]

PR #131, opened 2018-08-18
check if xknx object is available before accessing it.

firefrei:
This should fix issue #94

I actually don't know if this behaviour also exists in python. But the problem could be, that the function, where the xknx object is used, is already called by the base class during construction of the instance. so before the part of the constructor is reached where the xknx object is copied to the instance, the member function is already called (e.g. by the base class constructor).

a simple check if it already exists solves the problem (tested in my environment)

### DIFF
--- a/xknx/io/udp_client.py
+++ b/xknx/io/udp_client.py
@@ -55,12 +55,14 @@ class UDPClient:
                 self.data_received_callback(data)
 
         def error_received(self, exc):
-            """Handel errors. Callback for error received."""
-            self.xknx.logger.warning('Error received: %s', exc)
+            """Handle errors. Callback for error received."""
+            if 'xknx' in self:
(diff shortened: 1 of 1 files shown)


Julius2342:
Thank you!

rnixx:
@FireFrei 

> so before the part of the constructor is reached where the xknx object is copied to the instance, the member function is already called

There's not even a super() call in the constructor. Even if your assumption is right (I never encountered your described behavior), the correct condition would be ``hasattr(self, 'xknx')``

Julius2342:
I pushed a change, lets see if this works :)

rnixx:
@Julius2342 i bet it works ;)

FireFrei:
Looks good to me, but i wasn't able to test it yet. Sorry!

PR #139, opened 2018-08-30
Add fan device

itineric:
New fan device.
Speed is a percentage value (0 to 100).

### DIFF
--- a/test/fan_test.py
+++ b/test/fan_test.py
@@ -1,0 +1,149 @@
+"""Unit test for Fan objects."""
+import asyncio
+import unittest
+from unittest.mock import Mock, patch
+
+from xknx import XKNX
+from xknx.devices import Fan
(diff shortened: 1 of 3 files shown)


itineric:
What should be done to get the checks passing ?

Julius2342:
@itineric May I ask you to add a unit test for your new device? (esp in this case it is very important bc. i dont have fans and i need to know if i break something when i change something in the underlaying library in the future...

itineric:
Dont understand whats wrong now...

itineric:
I have add the hass component: https://github.com/home-assistant/home-assistant/pull/16434
Now i need to point the xknx version that includes the fan device, how/when will i be able to do that ?

Julius2342:
@itineric we need to push a new version with all the recent changes...

PR #140, opened 2018-09-02
[Request for feedback] [Breaking change] Do not read from write addresses?

Julius2342:
Background
=========

When i implemented XKNX, my first devices only had one address for reading and writing a state. I therefore thought this would be the default behavior for KNX in general. My original approach was, that XKNX took the same address for reading and writing - the state addresses were later introduced.

This causes the problem, that the state updater is trying read addresses which are not readable (and do not have a dedicated read address). This is the case e.g. many binary sensors. This leads to unnecessary traffic on the bus - and ugly log/warn messages in the log file.

```syslog
2018-09-02 22:21:13 WARNING (MainThread) [xknx.log] Could not read value of <BinarySensor group_address="GroupAddress("5/2/7")" name="DiningButton12" state="BinarySensorState.OFF"/> 5/2/7
```

Change
======

My suggested change is, to no longer fallback to the write address, if no state address is specified.

Side effects
=========

This change has the side effect, that all configurations, which rely on this fallback have to be changed. An additional state-address has to be introduced:

xknx.yaml
----------

```yaml
    light-switch-5:
        Children-L2:  {group_address_switch: '1/1/37'}
```

has to be changed to:

```yaml
    light-switch-5:
        Children-L2:  {group_address_switch: '1/1/37',  group_address_switch_state: '1/1/37'}
```

Hass config
------------

```yaml
light:
  - platform: knx
    name: Kitchen-L-1
    address: '1/0/9'
```

has to be changed to:

```yaml
light:
  - platform: knx
    name: Kitchen-L-1
    address: '1/0/9'
    state_address: '1/0/9'
```

Request for feedback
================

I'm uncertain if I should merge this PR, please leave your thumbs-up or thumbs-down. I will decide in a few weeks :-)

### DIFF
--- a/test/binary_sensor_test.py
+++ b/test/binary_sensor_test.py
@@ -220,5 +220,5 @@ class TestBinarySensor(unittest.TestCase):
     def test_state_addresses(self):
         """Test expose sensor returns empty list as state addresses."""
         xknx = XKNX(loop=self.loop)
-        switch = BinarySensor(xknx, 'TestInput', group_address='1/2/3')
-        self.assertEqual(switch.state_addresses(), [GroupAddress('1/2/3')])
+        binary_sensor = BinarySensor(xknx, 'TestInput', group_address='1/2/3', group_address_state='1/2/4')
+        self.assertEqual(binary_sensor.state_addresses(), [GroupAddress('1/2/4')])
--- a/test/cover_test.py
+++ b/test/cover_test.py
@@ -79,7 +79,7 @@ class TestCover(unittest.TestCase):
             'TestCover',
             group_address_long='1/2/1',
             group_address_short='1/2/2',
-            group_address_position='1/2/3')
+            group_address_position_state='1/2/3')
         self.loop.run_until_complete(asyncio.Task(cover.sync(False)))
         self.assertEqual(xknx.telegrams.qsize(), 1)
         telegram1 = xknx.telegrams.get_nowait()
@@ -110,8 +110,8 @@ class TestCover(unittest.TestCase):
             'TestCover',
             group_address_long='1/2/1',
             group_address_short='1/2/2',
-            group_address_position='1/2/3',
-            group_address_angle='1/2/4')
+            group_address_position_state='1/2/3',
+            group_address_angle_state='1/2/4')
         self.loop.run_until_complete(asyncio.Task(cover.sync(False)))
         self.assertEqual(xknx.telegrams.qsize(), 2)
         telegram1 = xknx.telegrams.get_nowait()
--- a/test/light_test.py
+++ b/test/light_test.py
@@ -73,9 +73,9 @@ class TestLight(unittest.TestCase):
         xknx = XKNX(loop=self.loop)
         light = Light(xknx,
                       name="TestLight",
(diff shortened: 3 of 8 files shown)


moustic999:
normally you should not try to read the sensor group-address. that is true.

the way KNX works make sure that an device does not have to ask to a sensor it's status. 

it is up to the sensor to send his status, ad you can choose how : only at change, or periodically.
you should never poll the bus the know the status.
The issue, and it is the case for most of the "visu" is at startup where you don't know the status of the sensor. in that case, it is helpfull to have a state address that you can poll, but you should never poll the "write" address. address and state address should never be the same. That will give issue in a KNX bus in the case where you have 2 toggle switch for the same light (actuator). In that scenario if address and state address is the same, it will not work anymore. because the "write" is issued by the switch device, and the state message is issued by the acutator itself ( which know the state of the light) the toggle switch never knows the state of the lamp, except if it receive it from the actuator.

I hope it is clear, but I can provide more explanation if needed.

in conclusion,
the only change you could do is to not try to read on the "write" address.
but doing it often works, so if you keep it, the only side effect is the line in the log.

rsterrenburg:
Totally agree with this change!

spali:
👍 
Thanks for this change, I always had warnings for central group addresses for which there is no state.
If I'm right, these should now not occur anymore due this change if I don't set a state_address.
Hope 0.9.0 gets soon integrated in HA.

Julius2342:
I made a [PR](https://github.com/home-assistant/home-assistant/pull/17912) within HASS.

resoai:
Perfect! looking forward to trying it out with the next release!

rsterrenburg:
Nice!

resoai:
I am still getting warnings on startup for every climate:

```
- platform: knx
  name: Guest Room
  temperature_address: 4/6/13
  target_temperature_address: 4/5/13
  setpoint_shift_address: 3/3/13
  setpoint_shift_state_address: 4/3/13
  operation_mode_address: 3/1/13
  controller_status_state_address: 4/0/13
```

I've added new line so it would be easier to read:

```
Could not read value of <Climate name="Guest Room" 

temperature="None/GroupAddress("4/6/13")/<DPTArray value="[0x7,0xe7]" />/20.23"  
target_temperature="GroupAddress("4/5/13")/None/<DPTArray value="[0x7,0x6c]" />/19.0"  
setpoint_shift="GroupAddress("3/3/13")/GroupAddress("4/3/13")/<DPTArray value="[0xfc]" />/-4" 
setpoint_shift_step="0.5" 
setpoint_shift_max="6" 
setpoint_shift_min="-6" 
group_address_operation_mode="GroupAddress("3/1/13")" 
group_address_operation_mode_state="None" 
group_address_controller_status="None" 
group_address_controller_status_state="GroupAddress("4/0/13")" 
group_address_controller_mode="None" 
group_address_controller_mode_state="None" 
group_address_on_off="None/None/None/None" /> 3/1/13
```

Julius2342:
@resoai you have some "special" setting. Is it only the climate? (is the number of messages now reduced?)

resoai:
Nope. The number stayed same as before the update and it is only happening to the climate component.

PR #141, opened 2018-09-11
Tunnel: Use the bus address assigned by the server

smurfix:
Connecting to a tunnel server assigns a physical bus address to us. Use it.

There probably is a better way to do this, and the configured address should override the one from the server, but there are enough gateways which limit you to the dynamic address which they assign to force this to be the default.


Julius2342:
Yes, makes sense!

PR #143, opened 2018-09-12
Added 'Voltage' as remote sensor value

magenbrot:


magenbrot:
hm. I got that pr somehow mixed up with the other. sorry for the confusion. I'm mainly an administrator, no developer ;)

Julius2342:
MAy you add a unit test for this? Ideally for the sensor and for hte type ...?

magenbrot:
I'll try :)

magenbrot:
do you have an idea how I can remove the commit that somehow was added in? If not, I would close this PR and start over. (I'm not doing this kind of stuff so it's some trial and error :/ )

Julius2342:
you can do `git revert <commit-hash>`.

Alternatively you can create a new branch from master and cherrypick the commits (in the right order). (A cherry-picked commit does get a new hash-id).

Alternatively you can force push the old commit-hash id :)

Julius2342:
Ah, and as i have merged the other PR already, the errornous commit should not be a problem.

Issue #138, opened 2018-08-30
INFO log level too high for E_NO_ERROR


itineric:
Hi,

I am getting tons of messages like this in my log:
2018-08-30 10:33:13 INFO (MainThread) [xknx.log] Success: received correct answer from KNX bus: ErrorCode.E_NO_ERROR
I think the log level should be DEBUG and not INFO since the message says: "everything is ok"
I get this message every 10 seconds at least, sometimes many times per second (get ~500 per hour)

magenbrot:
I had the same 'problem'. This PR sets the loglevel to DEBUG for success messages #144

PR #144, opened 2018-09-12
lowered loglevel from INFO to DEBUG for 'correct answer from KNX bus'

magenbrot:


magenbrot:
I don't know whats the problem with the check. The change works in my home-assistant setup.

Julius2342:
I would guess you have to change this

`with patch('logging.Logger.info') as mock_info:`

to:

`with patch('logging.Logger.debug') as mock_debug`

magenbrot:
heureka ;)

jetzt erstmal ein magenbrot!

magenbrot:
Closes #138

PR #153, opened 2018-10-06
tunable white function for lights

farmio:
- added tunable white function for setting relative color temperature for lights
- tests for tunable white function
- whitelist 'make' for tox
- fixed docstings in light.py
- removed double whitespace in device.py


farmio:
superseded by #154 as it incorporates absolute color temperature too.

PR #158, opened 2018-12-20
Implemented override features for min and max temperatures.

marvin-w:
Background: Some devices do not have a setpoint shift address and get a flaky frontend because of it
This should also fix issues with increasing temperature in the classic UI if the new properties are set.

* Implemented mapping for operation modes to allow the frontend to set supposingly wrong operation list

Next steps: Implement the new operation modes in the frontend, I can do that after this PR makes it into HA.

Fixes https://github.com/home-assistant/home-assistant/issues/14531

### DIFF
--- a/home-assistant-plugin/custom_components/climate/xknx.py
+++ b/home-assistant-plugin/custom_components/climate/xknx.py
@@ -5,15 +5,14 @@ For more details about this platform, please refer to the documentation at
 https://home-assistant.io/components/climate.knx/
 """
 
+import homeassistant.helpers.config_validation as cv
 import voluptuous as vol
-
+from custom_components.xknx import ATTR_DISCOVER_DEVICES, DATA_XKNX
 from homeassistant.components.climate import (
     PLATFORM_SCHEMA, SUPPORT_OPERATION_MODE, SUPPORT_TARGET_TEMPERATURE,
     SUPPORT_ON_OFF, ClimateDevice)
-from custom_components.xknx import ATTR_DISCOVER_DEVICES, DATA_XKNX
 from homeassistant.const import ATTR_TEMPERATURE, CONF_NAME, TEMP_CELSIUS, STATE_UNKNOWN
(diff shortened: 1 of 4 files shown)


home-assistant-plugin/custom_components/climate/xknx.py
```diff
@@ -128,14 +131,37 @@ def async_add_devices_config(hass, config, async_add_devices):
...
+        override_min_temp=config.get(CONF_OVERRIDE_MIN_TEMP),
+        override_max_temp=config.get(CONF_OVERRIDE_MAX_TEMP))
     hass.data[DATA_XKNX].xknx.devices.add(climate)
     async_add_devices([KNXClimate(hass, climate)])
 
 
 class KNXClimate(ClimateDevice):
     """Representation of a KNX climate device."""
 
+    operationModeMapping = {
```
marvin-w: I don't like this personally but I can't think of another way to make KNX comply with other standard climate devices out there.

home-assistant-plugin/custom_components/climate/xknx.py
```diff
@@ -35,6 +35,8 @@
 CONF_OVERRIDE_SUPPORTED_OPERATION_MODES = 'override_supported_operation_modes'
 CONF_ON_OFF_ADDRESS = 'on_off_address'
 CONF_ON_OFF_STATE_ADDRESS = 'on_off_state_address'
+CONF_OVERRIDE_MIN_TEMP = 'override_min_temp'
+CONF_OVERRIDE_MAX_TEMP = 'override_max_temp'
```
Julius2342: Why not just `CONF_MIN_TEMP` ? If it is specified as parameter, you override it. So it is redundant.

marvin-w: Thanks will change it!

I did it this way initially because it might confuse people who are using the setpoint shift calculations as this will override any of these changes. I might change this a bit.

PR #159, opened 2018-12-22
Split up Climate into Climate and ClimateMode

Julius2342:
The Climate object became to large. The operation-modes and all the temperature handling are related but do not necessarily belong together.


Julius2342:
@marvin-w  : May you have a look?

marvin-w:
@Julius2342: I like the idea of splitting on this abstraction layer because I think we wouldn't be happy if we splitted this by device type.

I will test the HA part later with my devices but it looks good.

PR #160, opened 2018-12-23
Climate improvements

Julius2342:
ClimateMode is now a member of Climate.


marvin-w:
Can we do sth about the coverage? Other than that it looks ok for me. I will test the HA part when I have time, maybe this eve.

PR #165, opened 2019-01-07
HA plugin: reduced calls to color_util

farmio:
use instance variable instead of calculating max and min mired every time its called

### DIFF
--- a/home-assistant-plugin/custom_components/light/xknx.py
+++ b/home-assistant-plugin/custom_components/light/xknx.py
@@ -35,12 +35,10 @@ CONF_MAX_KELVIN = 'max_kelvin'
 DEFAULT_NAME = 'XKNX Light'
 DEFAULT_COLOR = [255, 255, 255]
 DEFAULT_BRIGHTNESS = 255
-DEFAULT_COLOR_TEMPERATURE = 333  # 3000 K
+DEFAULT_COLOR_TEMPERATURE = 333  # mireds ~ 3000 K
 DEFAULT_COLOR_TEMP_MODE = 'absolute'
-DEFAULT_MIN_MIREDS = 166  # 6000 K
-DEFAULT_MAX_MIREDS = 370  # 2700 K
-DEFAULT_MIN_KELVIN = 2700
-DEFAULT_MAX_KELVIN = 6000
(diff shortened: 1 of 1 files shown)


home-assistant-plugin/custom_components/light/xknx.py
```diff
@@ -197,42 +204,12 @@ def color_temp(self):
     @property
     def min_mireds(self):
         """Return the coldest color temp this light supports in mireds."""
-        if self.device.supports_color_temperature or \
-                self.device.supports_tunable_white:
-            kelvin = self.device.max_kelvin
-            return color_util.color_temperature_kelvin_to_mired(kelvin) \
-                if kelvin is not None else DEFAULT_MIN_MIREDS
-        return None
```
farmio: https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/light/__init__.py doesn't return None when color temperature is not supported so I think we're good always returning our default.

marvin-w: Yep this looks good!

home-assistant-plugin/custom_components/light/xknx.py
```diff
@@ -197,42 +204,12 @@ def color_temp(self):
...
         """Return the warmest color temp this light supports in mireds."""
-        if self.device.supports_color_temperature or \
-                self.device.supports_tunable_white:
-            kelvin = self.device.min_kelvin
-            return color_util.color_temperature_kelvin_to_mired(kelvin) \
-                if kelvin is not None else DEFAULT_MAX_MIREDS
-        return None
-
-    @property
-    def min_kelvin(self):
```
marvin-w: I assume we removed this kelvin* properties due to the fact that HA doesn't support it right?

farmio: yes! No need to override. Instance variable is enough since its not going to change at runtime. 
If HA decides to support Kelvin too we can always make it a property then.

marvin-w: Why do we need the kelvin variables at all then? :)

marvin-w:
@Julius2342 : Please merge this

Julius2342:
sorry for the delay.

PR #167, opened 2019-01-11
Added DPT-3s and corresponding sensor types (second try)

buergi:
Added all missing DPT-3 datatypes, DPT 3.007 and 3.008, each in their stepwise and startstop mode, see [ABB Applications Manual](https://library.e.abb.com/public/78c74aa86d4648b7b9d918485cd4621a/2CDC500051M0203_ApplicationHB_Lighting_EN.pdf#page=34) for details.
I'm using them to control my Hues with my Busch Jäger push button-coupling 6108/02.

### DIFF
--- a/xknx/dpt/__init__.py
+++ b/xknx/dpt/__init__.py
@@ -6,6 +6,15 @@ Module for encoding and decoding KNX datatypes.
 """
 # flake8: noqa
 from .dpt import DPTArray, DPTBase, DPTBinary, DPTComparator
+from .dpt_1byte_control import (
+    DPTControl,
+    DPTControlStartStop,
+    DPTControlStartStopBlinds,
+    DPTControlStartStopDimming,
+    DPTControlStepwise,
+    DPTControlStepwiseBlinds,
+    DPTControlStepwiseDimming,
+)
 from .dpt_1byte_signed import DPTPercentV8, DPTSignedRelativeValue, DPTValue1Count
 from .dpt_1byte_uint import (
     DPTDecimalFactor,
--- a/xknx/dpt/__init__.py
+++ b/xknx/dpt/__init__.py
@@ -6,15 +6,6 @@ Module for encoding and decoding KNX datatypes.
 """
 # flake8: noqa
 from .dpt import DPTArray, DPTBase, DPTBinary, DPTComparator
-from .dpt_1byte_control import (
-    DPTControl,
-    DPTControlStartStop,
-    DPTControlStartStopBlinds,
-    DPTControlStartStopDimming,
-    DPTControlStepwise,
-    DPTControlStepwiseBlinds,
-    DPTControlStepwiseDimming,
-)
 from .dpt_1byte_signed import DPTPercentV8, DPTSignedRelativeValue, DPTValue1Count
 from .dpt_1byte_uint import (
     DPTDecimalFactor,
@@ -70,6 +61,15 @@ from .dpt_2byte_uint import (
     DPTTimePeriodMsec,
     DPTTimePeriodSec,
     DPTUElCurrentmA,
+)
+from .dpt_4bit_control import (
+    DPTControlStartStop,
+    DPTControlStartStopBlinds,
+    DPTControlStartStopDimming,
+    DPTControlStepCode,
+    DPTControlStepwise,
+    DPTControlStepwiseBlinds,
+    DPTControlStepwiseDimming,
 )
 from .dpt_4byte_float import (
     DPT4ByteFloat,
--- a/xknx/dpt/dpt_1byte_control.py
+++ b/xknx/dpt/dpt_1byte_control.py
@@ -29,8 +29,9 @@ class DPTControl(DPTBase):
     APCI_STEPCODEMASK = 0x07
     APCI_MAX_VALUE = APCI_CONTROLMASK | APCI_STEPCODEMASK
 
-    payload_length = 1
+    value_type = "control"
     unit = ""
+    payload_length = 1
 
     @classmethod
     def _encode(cls, control, step_code):
@@ -104,6 +105,9 @@ class DPTControl(DPTBase):
 class DPTControlStepwise(DPTControl):
     """Abstraction for KNX DPT 3.xxx in stepwise mode with conversion to an incement value."""
 
+    dpt_main_number = 3
+    dpt_sub_number = None
+    value_type = "stepwise"
     unit = "%"
 
     @staticmethod
@@ -158,9 +162,17 @@ class DPTControlStepwise(DPTControl):
 class DPTControlStepwiseDimming(DPTControlStepwise):
     """Abstraction for KNX DPT 3.007 / DPT_Control_Dimming in stepwise mode."""
 
+    dpt_main_number = 3
+    dpt_sub_number = 7
+    value_type = "stepwise_dimming"
+
 
 class DPTControlStepwiseBlinds(DPTControlStepwise):
     """Abstraction for KNX DPT 3.008 / DPT_Control_Blinds in stepwise mode."""
+
+    dpt_main_number = 3
+    dpt_sub_number = 8
+    value_type = "stepwise_blinds"
 
 
 class TitleEnum(Enum):
@@ -178,6 +190,7 @@ class TitleEnum(Enum):
 class DPTControlStartStop(DPTControl):
     """Abstraction for KNX DPT 3.xxx in start/stop mode."""
 
+    value_type = "startstop"
     unit = ""
 
     class Direction(TitleEnum):
@@ -221,6 +234,8 @@ class DPTControlStartStop(DPTControl):
 class DPTControlStartStopDimming(DPTControlStartStop):
     """Abstraction for KNX DPT 3.007 / DPT_Control_Dimming in start/stop mode."""
 
+    value_type = "startstop_dimming"
+
     # redefining Direction enum ensures proper typing, e.g.
     # DPTControlStartStop.Direction.INCREASE != DPTControlStartStopDimming.Direction.INCREASE
     class Direction(TitleEnum):
@@ -233,6 +248,8 @@ class DPTControlStartStopDimming(DPTControlStartStop):
 
 class DPTControlStartStopBlinds(DPTControlStartStop):
     """Abstraction for KNX DPT 3.008 / DPT_Control_Blinds in start/stop mode."""
+
+    value_type = "startstop_blinds"
 
     class Direction(TitleEnum):
         """Enum for indicating the direction."""
--- a/xknx/dpt/dpt_1byte_control.py
+++ b/xknx/dpt/dpt_1byte_control.py
@@ -1,0 +1,242 @@
+"""Implementation of Basic KNX DPT B1U3 Values (DPT 3.007/3.008).
+
+Very good source of information for interpretation of the standard is
+https://library.e.abb.com/public/78c74aa86d4648b7b9d918485cd4621a/2CDC500051M0203_ApplicationHB_Lighting_EN.pdf#page=34
+
+There are two separate dimming modes sharing the same DPT class:
+
+ * Stepwise dimming
+   The full brightness range is divided into 2^(stepcode-1) intervals.
+   The value is always rounded to full interval boundary, i.e. 30% +25% = 50%, 50% +25% = 75%, 30% -25% = 25%
+
+ * Start-stop dimming
+   Dimming is started with -/+100% (0x1/0x9) and keeps dimming until a STOP diagram (0x0/0x8) is received.
+
+As the same payload in these cases in interpreted completely different it is reasonable to make separate DPT classes.
+"""
+from enum import Enum
+
+from xknx.exceptions import ConversionError
+
+from .dpt import DPTBase
+
+
+class DPTControl(DPTBase):
+    """Abstraction for KNX B1U3 values (DPT 3.007/3.008)."""
+
+    # APCI (application layer control information)
+    APCI_CONTROLMASK = 0x08
+    APCI_STEPCODEMASK = 0x07
+    APCI_MAX_VALUE = APCI_CONTROLMASK | APCI_STEPCODEMASK
+
+    payload_length = 1
+    unit = ""
+
+    @classmethod
+    def _encode(cls, control, step_code):
+        """Encode control-bit with step-code."""
+        value = 1 if control > 0 else 0
+        value = (value << 3) | (step_code & cls.APCI_STEPCODEMASK)
+        return value
+
+    @classmethod
+    def _decode(cls, value):
+        """Decode value into control-bit and step-code."""
+        control = 1 if (value & cls.APCI_CONTROLMASK) != 0 else 0
+        step_code = value & cls.APCI_STEPCODEMASK
+        return control, step_code
+
+    @classmethod
+    def _test_boundaries(cls, raw):
+        """Test if raw KNX data is within defined range for this object."""
+        if isinstance(raw, int):
+            return 0 <= raw <= cls.APCI_MAX_VALUE
+        return False
+
+    @classmethod
+    def _test_values(cls, control, step_code):
+        """Test if input values are valid."""
+        if isinstance(control, int) and isinstance(step_code, int):
+            if control in (0, 1) and 0 <= step_code <= cls.APCI_STEPCODEMASK:
+                return True
+        return False
+
+    @classmethod
+    def to_knx(cls, value, invert=False):
+        """Serialize to KNX/IP raw data."""
+        if not isinstance(value, dict):
+            raise ConversionError(
+                "Cant serialize %s; invalid value type" % cls.__name__, value=value
+            )
+
+        try:
+            control = value["control"]
+            step_code = value["step_code"]
+        except KeyError:
+            raise ConversionError(
+                "Cant serialize %s; invalid keys" % cls.__name__, value=value
+            )
+
+        if not cls._test_values(control, step_code):
+            raise ConversionError(
+                "Cant serialize %s; invalid values" % cls.__name__, value=value
+            )
+
+        if invert:
+            control = 0 if control > 0 else 1
+
+        return cls._encode(control, step_code)
+
+    @classmethod
+    def from_knx(cls, raw, invert=False):
+        """Parse/deserialize from KNX/IP raw data."""
+        if not cls._test_boundaries(raw):
+            raise ConversionError("Cant parse %s" % cls.__name__, raw=raw)
+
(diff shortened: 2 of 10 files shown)


nodefeet:
Thank you very much @buergi , that's what was missing in xknx for me.
I think the relativ dimming should additionally be implemented in the device class Light to cover the basic functionality someone would expect from a KNX Light.

farmio:
@nodefeet could you please explain your usecase for DPT-3 in the Light module? My only usage of xknx is through home-assistant - which has no  concept of holding down buttons / switches. Are you using xknx outside of HA?

nodefeet:
Yes and it's really just that for that: dimming directly with KNX switches outside of the HA interface. A lot of visualizations for smart homes offer the relative dimming, but I would agree that as soon as you have a virtual slider or knob it’s not really necessary. But with KNX compatible switches it is very useful to dim the light directly from your physical switch (the fastest app is still no app).

From my experience the stepwise dimming is rarely used in practice, partly due to that it is swiftly flooding the bus with excess telegrams, but, the start stop dimming is imho part of any basic KNX installation with lights. After the brightness is modified via relative dimming (i. e. after receiving the stop telegram or reaching 0 or 100 % brightness ) KNX dimming actuators typically return the actual brightness to the bus (so that interfaces like HA can update their status). 

Let me know if I can provide any additional input.

nodefeet:
@farmio  I see now that you were specifically asking about the use case of DPT-3 in the **class Light** and not just about the use case of DPT-3 in general.

So let me add that since I think DPT-3 or relative dimming is essential for an abstraction of a KNX Light I would expect to not just find the `group_address_brightness(_absolut)`, but also a `group_address_brightness_relative` and the associated functionality in the device class Light.

farmio:
Adding a stepwise dimming GA to the **Light class** would be no big deal. As far as im concerned it is a stateless action so we wouldn't even need a state GA or a `current_* property`. It would just forward the given value to the GA. 
Functionally this would be no different to adding a *Sensor* with DPT-3 type. I see that for the sake of completeness it could be added to the Light class.
I'd say its @Julius2342 call, since he doesn't want the light module to get too bloated.

buergi:
Which bitshifting do you mean? The few lines inside the to_knx method? After some pondering I couldn't make up any better representation that shows the origin of the values better of course I could encapsulate the bitshifting like this, if you prefer this, but I wouldn't consider this to increase the readability

    MakeDPT3 = lambda sign, stepcode: DPTBinary(sign << 3 | stepcode)

xknx/devices/remote_value_dpt3.py
```diff
@@ -0,0 +1,215 @@
...
+            return DPTBinary(0x01) if not self.invert else DPTBinary(0x09)
+        if value == self.Direction.STOP:
+            return DPTBinary(0x00)
+        raise ConversionError("value invalid", value=value, device_name=self.device_name)
+
+    def from_knx(self, payload):
+        """Convert current payload to value."""
+        if payload.value & ~0x0F != 0:  # more than 4-bit
+            pass  # raises exception below
+        elif payload.value & 0x07 == 0:
```
Julius2342: Just that i understand this right, this evaluates for `value=8` and `value=0` ...?

(I'm not too good in bitwise calculations, bit i think we should be more explicit here:

`if payload.value in [0, 8]:` if this is intended.

xknx/devices/remote_value_dpt3.py
```diff
@@ -0,0 +1,215 @@
...
+            return self.Direction.STOP
+        elif payload.value & 0x08 == 0:
```
Julius2342: `payload.value < 8`

buergi:
Integrated your most if not all of your remarks and ported the remote value to a new dpt class. What do you think?

farmio:
Thank you for updating it. This is very much appreciated. 

I think remote_value_sensor should be used for numeric datatypes only. I would add a new class (maybe "remote_value_control") for this - and maybe DPT2 in the future. This would also enable us to pass tuples holding control and step values so we can use them too.
I'd also create a new device subclass for this. This would be a cleaner imho.

buergi:
I see your point, having a sensor for a control state is a bit weird, especially as you cannot set the value for controlling it.
I have a remote_value_dpt3 implementation in my desk drawer, however, it currently does not use the DPT3 class. However, it would be no problem adapting it.
However, I have no idea, how I'd integrate this remote value class into a usable device in home assistant.
So, what do you suggest, how to go on? I remove the sensor integration, add a remote value class and we then probably need a new device type, e.g. control.

As shortly explained in the docstring, there are two control types stepwise dimming and start/stop dimming. The frontend for the start/stop could probably look similar to the cover control. For the stepwise dimming it would be similar, plus evtl. another slider to allow setting the step size.

For easy use of the control without any further automatisation the control's state would probably include an internally accumulated absolute value of the dimming/cover state.

farmio:
You're right, the remote_value class is really not a big deal - its mostly boilerplate.

I like xknx classes to be explicit. HA is another issue - we have to adapt to the available platforms. We could 
- integrate the Control device in the HA integration 
- or maybe the RemoteValueControl in the xknx Sensor class.

In ExposeSensor we have something similar: https://github.com/XKNX/xknx/blob/213c6fc5c88648ca5565299d02c8693a2989e0a7/xknx/devices/expose_sensor.py#L33

For stepwise dimming it would also be an option to make the step size configurable via yaml. Just like KNX switches where the stepsize is set once via ETS. It could always be bypassed by eg. HA-automations with an optional argument.

recMartin:
I made similar work and only now saw that there exists this still unmerged PR.
Perhaps someone which has a better view of the complete xknx than me can look at both?
See #262 .

buergi:
Thanks @recMartin for your implementation. I looked over the code and like the idea of having DPTControl implemented as "control + stepcode". It is closer to the KNX standard as my "increment" implementation and especially as it is bijective (invertable).

However, from a user perspective I'd expect to work with increment values, e.g. make the light 25% brigher. I'm currently consider moving the increment conversion to a corresponding RemoteValue type and stripping the DPT implementation down to the control+stepcode approach.

As @recMartin said he won't do any changes on his PR at the moment I'd suggest I keep working in this PR and try to merge both PRs.

recMartin:
I had a look in the code, its a different approach than mine (#262), but also good work!
Some comments and suggestions:
- I like that this solution does not need a list (as mine does) but only one value
- Maybe we should make increment_to_stepcode(value) inverse of stepcode_to_increment(stepcode, sign) this means:
  - stepcode_to_increment(stepcode, sign) -> value 
  - increment_to_stepcode(value)          -> (value, sign) , calculate 'sign' in this function
- DPT3 looks like only a base-class, but it is also used for both, stepwise-dimming and controlling blinds
- the derived classes 'DPTControlDimming' and 'DPTControlBlinds' can only be used for start/stop dimming. 
  All step dimmings +/-1 to +/-100% are mapped to +/-100% and output as DECREASE or INCREASE or STOP, so some information is lost in from_knx (not the correct step is output).
- Could we allow/handle both types of values in the DPT classes or handle this only in remote_control?
- Since DPT 3.00x is not for sensors, but for actuators/controls, it should be incorporated in e.g. "remote_value_control" (as farmio suggested) and not in remote_sensor.

buergi:
Thanks @recMartin for your input. I've been working on the code yesterday and a lot has changed. I've not pushed it yet as I have some open points I still have to ponder about. Especially as I already have a future control device implementation in mind when developing.

I currently implemented a base DPTControl class based mostly on your code with a few esthetical changes. From this I derived a DPTControlStepwise which is mostly my code. I renamed the DPTControlDimming/Blinds to DPTControlStartStopDimming/Blinds (which I actually already had in the original code of this PR). I still like to add an intermediate DPTControlStartStop class to reduce the amount of duplicated code, but I'm not done with it yet
The implementation of the RemoteValueControl is very similar to your implementation (with a DPTMap as in RemoteValueSensor).

I already added unit tests for all important cases I could think of.
I guess this should tackle all your valid points of critisism you mentioned, but I'm open to new comments once I pushed the code.

I currently don't plan to add a new Control class in this PR, but I have some ideas and will make another PR when I'm ready. In short I plan to make a Control with an integrated absolute counter which can be configured to stepwise or startstop mode. The first case is simple and the absolute value is changed directly when a step-message is received. For the second case I have an implementation in mind which is similar to the one of the cover device, which uses a method which is added as a job to home assistant (see [cover.py)](https://github.com/XKNX/xknx/blob/b097e1978f7c86084d68c415a646c19a27b7abe7/home-assistant-plugin/custom_components/xknx/cover.py#L211)

buergi:
I again rebased the PR and updated the DPT mapping in the RemoteValueControl to match that of RemoteValueSensor.

There is a slight problem with the `dpt_main_number` and `dpt_sub_number` convention. It is meant to be unique, however, in this case it isn't. There are two different interpretations for DPT3 values, stepwise and start/stop. A +100% message e.g. in the context of light dimming can either mean "turn the brightness to maximum" or "start increasing the brighness (until a 0% message is received)". Therefore both are handled via separate classes so only one can carry the `dpt_*_number` flags. I decided for the stepwise class as it carries more information than the start/stop class.

Currently, this value class cannot be used from within HA. I still plan to add a new kind of device which has an internal value that can be controlled with this type of messages. However, in the meantime, and also for flexibility reasons I would suggest extending the RemoteValueSensor type to allow DPTBinary payloads just by adding `or isinstance(payload, DPTBinary)` to the return statement of its `payload_valid` method and a slight adaption in its `to_knx` method. This makes no sense for stepwise controls but would allow using start/stop switches in custom automations. However, as I thought you might not like this idea I did not include it in the PR.

farmio:
Hi! 
Thanks for your contribution. We have some bugs currently that keep me form reviewing this PR. 

All I can say right now is that I'd not be in favour of adding `or isinstance(payload, DPTBinary)` to RemoteValueSensor.payload_valid() . I'd rather use a new RemoteValue subclass and use it form Sensor() - just as done in ExposeSensor for binary type.

xknx/dpt/dpt_1byte_control.py
```diff
@@ -0,0 +1,259 @@
...
+        """Convert value to payload."""
+        control = 0
+        step_code = 0
+        if value == cls.Direction(1):  # INCREASE/DOWN
+            control = 1
+            step_code = 1
+        elif value == cls.Direction(0):  # DECREASE/UP
+            control = 0
+            step_code = 1
+        elif value == cls.Direction(2):  # STOP
```
farmio, line 214:
```suggestion
        if value == cls.Direction.INCREASE:  # Down
            control = 1
            step_code = 1
        elif value == cls.Direction.DECREASE:  # Up
            control = 0
            step_code = 1
        elif value == cls.Direction.STOP:
```

xknx/dpt/dpt_1byte_control.py
```diff
@@ -0,0 +1,259 @@
...
+class TitleEnum(Enum):
+    """Enum with a descriptive string representation.
+
+    Ensures values are rendered nicely, e.g. in home assistant.
+    """
+
+    def __str__(self):
+        """Return string representation."""
+        # pylint: disable=no-member
+        return self.name.title()
```
farmio, line 187: Why not use the representation string as Enum.value?
`    DECREASE = "Decrease"`

xknx/dpt/dpt_1byte_control.py
```diff
@@ -0,0 +1,259 @@
...
+
+    @classmethod
+    def from_knx(cls, raw, invert=False):
+        """Convert current payload to value."""
+        values = super().from_knx(raw, invert)
+        if values["step_code"] == 0:
+            return cls.Direction(2)  # STOP
+        if values["control"] == 0:
+            return cls.Direction(0)  # DECREASE/UP
+        return cls.Direction(1)  # INCREASE/DOWN
```
farmio, line 231:
```suggestion
        if values["step_code"] == 0:
            return cls.Direction.STOP
        if values["control"] == 0:
            return cls.Direction.DECREASE
        return cls.Direction.INCREASE
```

farmio (changes requested):
I have some readability/style suggestions.
The suggested type annotations are just for better overview.

xknx/dpt/__init__.py
```diff
@@ -6,6 +6,15 @@
 """
 # flake8: noqa
 from .dpt import DPTArray, DPTBase, DPTBinary, DPTComparator
+from .dpt_1byte_control import (
```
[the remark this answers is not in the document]
buergi: hm, you're right... though we name all dpts with <2 byte `dpt_1byte_xxx`, but no problem to rename this

xknx/dpt/dpt_1byte_control.py
```diff
@@ -0,0 +1,259 @@
...
+As the same payload in these cases in interpreted completely different it is reasonable to make separate DPT classes.
+"""
+from enum import Enum
+
+from xknx.exceptions import ConversionError
+
+from .dpt import DPTBase
+
+
+class DPTControl(DPTBase):
```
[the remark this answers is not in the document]
buergi: Your right, I initially planned to make it a base class for all control types, DPT2 and 3, but in the current implementation it simply isn't. I'll rename the class.

xknx/dpt/dpt_1byte_control.py
```diff
@@ -0,0 +1,259 @@
...
+    def _encode(cls, control, step_code):
+        """Encode control-bit with step-code."""
+        value = 1 if control > 0 else 0
+        value = (value << 3) | (step_code & cls.APCI_STEPCODEMASK)
+        return value
+
+    @classmethod
+    def _decode(cls, value):
+        """Decode value into control-bit and step-code."""
+        control = 1 if (value & cls.APCI_CONTROLMASK) != 0 else 0
```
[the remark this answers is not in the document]
buergi: Would be possible, however, I don't like it a lot as it is not directly obvious what possible values `control` could get. One first has to look up the value of the mask and count the bits to understand where the `3` is coming from. Especially having in mind the implementation for the DPT2 class we would only need to change the APCI_xxxMASK values to reuse it for a 2-bit value.

xknx/dpt/dpt_1byte_control.py
```diff
@@ -0,0 +1,259 @@
...
+    value_type = "control"
+    unit = ""
+    payload_length = 1
+
+    @classmethod
+    def _encode(cls, control, step_code):
+        """Encode control-bit with step-code."""
+        value = 1 if control > 0 else 0
+        value = (value << 3) | (step_code & cls.APCI_STEPCODEMASK)
+        return value
```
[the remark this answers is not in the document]
buergi: I actually don't like this at all, as you put in the intrinsic assumption that control is always either 0 or 1. You added the type hints, so the type checker should detect this, but users of the xknx library might not use type checking in their code. Just imagine I would pass control=3.
It would be possible to use use `(control & cls.APIC_CONTROLMASK) << 3`, but as above this diminishs the abstraction, as the `3` contains redundant information which is already contained in `APIC_CONTROLMASK`. So I'll add the type hints, but would prefer to keep the rest.

xknx/dpt/dpt_1byte_control.py
```diff
@@ -0,0 +1,259 @@
...
+        """Parse/deserialize from KNX/IP raw data."""
+        if not cls._test_boundaries(raw):
+            raise ConversionError("Cant parse %s" % cls.__name__, raw=raw)
+
+        control, step_code = cls._decode(raw)
+
+        if invert:
+            control = 0 if control > 0 else 1
+
+        return {"control": control, "step_code": step_code}
```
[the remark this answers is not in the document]
buergi: I guess you meant `control ^ invert`. Would be possible, but do you really consider this easier to read? I find it somewhat crypting whereas the explicit version is more obvious. Plus, it would cause a mess if invert is not 0 or 1, which, however, is safe to assume as long as a boolean is passed.

xknx/dpt/dpt_1byte_control.py
buergi: Nop, not possible. I admit the Enum implementation is a little bit tricky. The idea is to use the 0 / 1 representation in the base class and thus for the encoding. And only in the subclasses I provide the interpretation, whether 0 is now DECREASE or UP.
It is necessary to really use the numeric represenation here. As the concrete interpretation of Direction is overridden when called from a derived class. Consider the following example

```
c = DPTControlStartStopBlinds()
v=c.from_knx(1)
print(v) # => Up
print(c.to_knx(v)) => 1
```

When I follow your suggestion I would get the following exception in the last line

```
  File "test.py", line 242, in <module>
    print(c.to_knx(v))
  File "test.py", line 186, in to_knx
    if value == cls.Direction.INCREASE:  # INCREASE/DOWN
  File "C:\tools\miniconda3\lib\enum.py", line 346, in __getattr__
    raise AttributeError(name) from None
AttributeError: INCREASE
```

`INCREASE` is not defined, as `DPTControlStartStopBlinds`'s `Direction` only knows `UP`, `DOWN` and `STOP`.

xknx/dpt/dpt_1byte_control.py
buergi: This would require reimplementing the `to_knx` and `from_knx` classes in both `DPTControlStartStopDimming` and `DPTControlStartStopBlinds`, and would not allow a generic implementation in `DPTControlStartStop`. See example in the previous comment above.

xknx/dpt/dpt_1byte_control.py
buergi: Not possible, for the same reason as above, `cls.Direction.DECREASE` is not defined when called from the child class `DPTControlStartStopBlinds`.

xknx/remote_value/remote_value_control.py
```diff
@@ -0,0 +1,54 @@
...
+        super().__init__(
+            xknx,
+            group_address,
+            group_address_state,
+            device_name=device_name,
+            after_update_cb=after_update_cb,
+        )
+        self.invert = invert
+        # pylint: disable=too-many-arguments
+        _dpt_class = DPTBase.parse_transcoder(value_type)
```
farmio, line 37:
```suggestion
        _dpt_class = DPTBase.parse_transcoder(value_type)
```
duplicate of line 27

farmio:
Hey! I think this can be merged now, but for some reason this is showing 99 files edited now. There are some commits in the changeset that are already merged a while ago. Can you please clean this up (hopefully) a last time?

buergi:
Upps, what did I do there, git keeps on surprising me. I now cherry-picked all my commits again upon the current head. I'm not sure why the coverage test fails. I don't understand the output of this unit test.

farmio (approved):

farmio:
Thank you very much! 
I'm looking forward to see this integrated to HA! 😃

PR #171, opened 2019-02-03
return None when color temperature is unknown

farmio:
- return None instead of default value when ct is unknown
- remove DEFAULT_COLOR_TEMPERATURE

cc @marvin-w for the first review change request on https://github.com/home-assistant/home-assistant/pull/19699


marvin-w:
Please make sure you are using an up to date version of your base branch.

Simply merge master in your feature branch and the checks should succeed.

I would also suggest to sync those things after HA has merged the PR, I will take care of that and will close this PR in the meantime. Thanks!

PR #175, opened 2019-02-08
Don't use read only GA for setting target temperature

### LINKED ISSUE #169: Don't use a read only address to write new temperatures
We currently use the `target_temperature_address` to query the target temperature but also to write a new value to it if we don't supply a `setpoint_shift_address`.

This behaviour is wrong and should be changes.

Provide an additional `target_temperature_state_address` attribute and use `target_temperature_address` only for writing.

marvin-w:
Please note that this will cause some kind of flickering in the frontend but it at least returns reasonable (correct) values for changed temperatures

This also fixes setting min and max temp through xknx.yaml

closes #169

### DIFF
--- a/docs/climate.md
+++ b/docs/climate.md
@@ -39,7 +39,8 @@ groups:
 
 * **group_address_temperature** KNX address of current room temperature
 * **group_address_setpoint** KNX address of basis setpoint.
-* **group_address_target_temperature** KNX address for reading the target temperature from KNX bus.
+* **group_address_target_temperature** KNX address for setting the target temperature if setpoint shift is unsupported.
+* **group_address_target_temperature_state** KNX address for reading the target temperature from the KNX bus.
 * **group_address_operation_mode** KNX address for operation mode.
 
 * **group_address_operation_mode_protection** KNX address for switching on/off frost/heat protection mode.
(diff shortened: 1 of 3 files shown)


marvin-w:
@Julius2342 Can you please test this with a setpoint shift device?

Julius2342:
I tried out your branch and got this error:

```
2019-02-19 09:51:47 ERROR (MainThread) [homeassistant.core] Error doing job: Task exception was never retrieved
  File "/home/julius/home-assistant/homeassistant/helpers/service.py", line 291, in _handle_service_platform_call
    await func(entity, data)
  File "/home/julius/home-assistant/homeassistant/components/climate/__init__.py", line 564, in async_service_temperature_set
    await entity.async_set_temperature(**kwargs)
  File "/home/julius/home-assistant/homeassistant/components/knx/climate.py", line 238, in async_set_temperature
    await self.device.set_target_temperature(temperature)
  File "/home/julius/xknx/xknx/devices/climate.py", line 180, in set_target_temperature
    await self.set_target_temperature_setpoint_shift(target_temperature)
  File "/home/julius/xknx/xknx/devices/climate.py", line 191, in set_target_temperature_setpoint_shift
    raise DeviceIllegalValue("setpoint_shift_max exceeded", setpoint_shift)
xknx.exceptions.exception.DeviceIllegalValue: <DeviceIllegalValue description="setpoint_shift_max exceeded" value="68" />
```

marvin-w:
@Julius2342  As far as I can tell my changes didn't do anything in that method. Are you sure your error is related to this PR?

PR #176, opened 2019-02-10
config test for light with ct and tw

farmio:
- added some missing tests for color temperature configs
- fixed a small typo


Julius2342:
Looks good! (I think documentation is independent).

marvin-w:
Thanks! I aggree that documentation shouldn't part of this PR but it should have been indeed a part of the original lights PR. Due to the fact they we hadn't the opportunity to manage docs alongside code back then I mentioned it here :).

So feel free to adjust them in a separate PR. 

IMHO a product is only as good as it's documentation but that's just my 2 cts

PR #184, opened 2019-02-23
Auto detection of local ip

farmio:
Make `local_ip` optional for tunneling and routing ConnectionConfig.

**Tunnelling**
- Function: An interface on the same subnet as the configured gateway will be selected. If none was found, an interface on the same subnet as the default gateway will be selected.
- Intention: Environments where multicast doesn't work or ip address of host changes.

**Routing**
- Function: If no `local_ip` is configured KNXIPInterface is started like Automatic with a GatewayScanFilter searching only routing devices. If the filter is set only routing connections will be made.
- Intention: Multiple gateways on the bus; routing shall be preferred to reserve tunneling connections for other applications (eg. ETS).

### DIFF
--- a/xknx/core/config.py
+++ b/xknx/core/config.py
@@ -65,14 +65,12 @@ class Config:
         if hasattr(prefs, '__iter__'):
             for pref, value in prefs.items():
                 try:
-                    if pref.startswith("local_ip"):
-                        conn.local_ip = value
-                    elif pref.startswith("gateway_ip"):
+                    if pref.startswith("gateway_ip"):
                         conn.gateway_ip = value
                     elif pref.startswith("gateway_port"):
-                        # don't overwrite default if None
-                        if value is not None:
-                            conn.gateway_port = value
+                        conn.gateway_port = value
+                    elif pref.startswith("local_ip"):
+                        conn.local_ip = value
                 except XKNXException as ex:
                     self.xknx.logger.error("Error while reading config file: Could not parse %s: %s", pref, ex)
         self.xknx.connection_config = conn
--- a/xknx/core/config.py
+++ b/xknx/core/config.py
@@ -85,7 +85,6 @@ class Config:
     def parse_group(self, doc, group):
         """Parse a group entry of xknx.yaml."""
         try:
-            print(group)
             if group.startswith("light"):
                 self.parse_group_light(doc["groups"][group])
             elif group.startswith("switch"):
--- a/xknx/core/config.py
+++ b/xknx/core/config.py
@@ -51,14 +51,18 @@ class Config:
             for conn, prefs in doc["connection"].items():
                 try:
                     if conn == "tunneling":
-                        conn = ConnectionType.TUNNELING
+                        if prefs is None or \
+                                not "gateway_ip" in prefs:
+                            raise XKNXException("`gateway_ip` is required for tunneling connection.")
+                        conn_type = ConnectionType.TUNNELING
                     elif conn == "routing":
-                        conn = ConnectionType.ROUTING
+                        conn_type = ConnectionType.ROUTING
                     else:
-                        conn = ConnectionType.AUTOMATIC
-                    self._parse_connection_prefs(conn, prefs)
+                        conn_type = ConnectionType.AUTOMATIC
+                    self._parse_connection_prefs(conn_type, prefs)
                 except XKNXException as ex:
                     self.xknx.logger.error("Error while reading config file: Could not parse %s: %s", conn, ex)
+                    raise ex
 
     def _parse_connection_prefs(self, conn_type: ConnectionType, prefs) -> None:
         connection_config = ConnectionConfig(connection_type=conn_type)
--- a/xknx/core/config.py
+++ b/xknx/core/config.py
@@ -50,30 +50,30 @@ class Config:
                 and hasattr(doc["connection"], '__iter__'):
             for conn, prefs in doc["connection"].items():
                 try:
-                    if conn == "auto":
-                        self._parse_connection_prefs(ConnectionType.AUTOMATIC, prefs)
-                    elif conn == "tunneling":
-                        self._parse_connection_prefs(ConnectionType.TUNNELING, prefs)
+                    if conn == "tunneling":
+                        conn = ConnectionType.TUNNELING
                     elif conn == "routing":
-                        self._parse_connection_prefs(ConnectionType.ROUTING, prefs)
+                        conn = ConnectionType.ROUTING
+                    else:
+                        conn = ConnectionType.AUTOMATIC
+                    self._parse_connection_prefs(conn, prefs)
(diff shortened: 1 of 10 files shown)


xknx/io/knxip_interface.py
```diff
@@ -91,10 +86,16 @@ def __init__(self, xknx, connection_config=ConnectionConfig()):
...
                 self.connection_config.scan_filter)
-        elif self.connection_config.connection_type == ConnectionType.ROUTING:
+        elif self.connection_config.connection_type == ConnectionType.ROUTING and \
+                self.connection_config.local_ip is not None:
             await self.start_routing(
                 self.connection_config.local_ip,
                 self.connection_config.bind_to_multicast_addr)
+        elif self.connection_config.connection_type == ConnectionType.ROUTING and \
+                self.connection_config.local_ip is None:
+            self.connection_config.scan_filter.routing = True
```
Julius2342: I think this is a hack :) I think is kind of a `const` object and we should not modify it.

xknx/io/knxip_interface.py
```diff
@@ -162,3 +166,33 @@ def telegram_received(self, telegram):
...
+                        network = ipaddress.IPv4Network((link["addr"],
+                                                         link["netmask"]),
+                                                        strict=False)
+                        if gateway in network:
+                            self.xknx.logger.debug("Using interface: %s", interface)
+                            return link["addr"]
+                except KeyError:
+                    self.xknx.logger.debug("Could not find IPv4 address on interface %s", interface)
+                    continue
+            raise XKNXException("No interface on same subnet as gateway found.")
```
Julius2342: I think we should not raise an exception here - and work with `None`.

xknx/io/knxip_interface.py
```diff
@@ -162,3 +166,33 @@ def telegram_received(self, telegram):
...
+
+        gateway = ipaddress.IPv4Address(gateway_ip)
+        try:
```
Julius2342: If above function returns `None` we could do:

```
local_ip = _scan_interfaces(gateway)
if local_ip is None:
    default_gateway = _find_default_gateway()  # <-- I would move this to a function.
    local_ip = _scan_interfaces(default_gateway)
return local_ip
```

I know that this whole block still may return `None` ... But from my understanding of networking, this should not happen under normal conditions.

xknx/core/config.py
```diff
@@ -65,14 +65,12 @@ def _parse_connection_prefs(self, conn_type, prefs):
...
-                        conn.local_ip = value
-                    elif pref.startswith("gateway_ip"):
+                    if pref.startswith("gateway_ip"):
                         conn.gateway_ip = value
                     elif pref.startswith("gateway_port"):
-                        # don't overwrite default if None
-                        if value is not None:
-                            conn.gateway_port = value
+                        conn.gateway_port = value
+                    elif pref.startswith("local_ip"):
```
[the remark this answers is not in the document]
farmio: No I just copied what was there before.

xknx/io/knxip_interface.py
farmio: The intention was to raise once -> try default gateway. Raise twice (no  default gateway found) -> exit since a bind(None) will for sure fail, but without a good error message.

xknx/io/knxip_interface.py
farmio: Shall we instead instanciate a new `GatewayScanFilter()` ?

xknx/io/knxip_interface.py
Julius2342: :)

xknx/io/knxip_interface.py
farmio: 🤔 now it seems even more hacky. 
Maybe we could define the GatewayScanFilter() when reading the configuration file / HA configuration. I'll look into it.

xknx/core/config.py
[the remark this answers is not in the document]
Julius2342: May you try to change this to an `=`-operator?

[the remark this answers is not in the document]
Julius2342: Ah, read first, comment then. Did not realize that you have already implemented this 👍

xknx/core/config.py
```diff
@@ -85,6 +85,7 @@ def parse_groups(self, doc):
     def parse_group(self, doc, group):
         """Parse a group entry of xknx.yaml."""
         try:
+            print(group)
```
Julius2342: :)

xknx/io/knxip_interface.py
```diff
@@ -162,3 +165,35 @@ def telegram_received(self, telegram):
...
+            gws = netifaces.gateways()
+            return ipaddress.IPv4Address(gws['default'][netifaces.AF_INET][0])
+
+        gateway = ipaddress.IPv4Address(gateway_ip)
+        local_ip = _scan_interfaces(gateway)
+        if local_ip is None:
+            self.xknx.logger.debug(
+                "No interface on same subnet as gateway found. Falling back to default gateway.")
+            default_gateway = _find_default_gateway()
+            local_ip = _scan_interfaces(default_gateway)
```
farmio: Shall we `try: except:` and `raise` here?

xknx/core/config.py
```diff
@@ -50,11 +50,11 @@ def parse_connection(self, doc):
                 and hasattr(doc["connection"], '__iter__'):
             for conn, prefs in doc["connection"].items():
                 try:
-                    if conn.startswith("auto"):
+                    if conn == "auto":
                         self._parse_connection_prefs(ConnectionType.AUTOMATIC, prefs)
-                    elif conn.startswith("tunneling"):
+                    elif conn == "tunneling":
                         self._parse_connection_prefs(ConnectionType.TUNNELING, prefs)
```
farmio: Shall we raise if `tunneling:` is used without `gateway_ip:`? Or maybe just `logger.warning` and fall back to ConnectionType.AUTOMATIC ?

xknx/io/knxip_interface.py
```diff
@@ -64,8 +64,12 @@ def __init__(self,
         self.gateway_port = gateway_port
         self.auto_reconnect = auto_reconnect
         self.auto_reconnect_wait = auto_reconnect_wait
-        self.scan_filter = scan_filter
         self.bind_to_multicast_addr = bind_to_multicast_addr
+        if connection_type == ConnectionType.TUNNELING:
+            scan_filter.tunnelling = True
+        elif connection_type == ConnectionType.ROUTING:
+            scan_filter.routing = True
+        self.scan_filter = scan_filter
```
farmio: GatewayScanFilter now accounting ConnectionType on initialization. I was thinking it wouldn't make sense to scan for routers if a tunneling connection shall be made (and vice versa).

xknx/core/config.py
Julius2342: I think we should raise ...

PR #189, opened 2019-03-08
new Sensor types, ExposeSensor fix

### LINKED ISSUE #114: Inverted DPT 5.001 Percentage values
As per xknx v0.8.2 all DPT 5.001 values are inverted. If GA with DPT 5.001 returns 0% then it is showed in HAss as 100%, 80% is showed as 20%, 100% as 0%. 

Regards,
Martijn

### LINKED ISSUE #116: DPT 16.000 Character String (ASCII)
It would be nice to be able to have a sensor type for strings. I would like to be able to display Diagnosis text messages from, for example, my heating actuators.

Ragards,
Tinus

### LINKED ISSUE #129: Exposing boolean values to the KNX bus
Hi,

I couldn't find a way to expose boolean values to the KNX bus (DPT 1.002). I'm using home-assistant. I would like to have that feature for my door and window contacts (opened=1/closed=0) or to tell my devices if it's day (=1) or night (=0).
Maybe it's already possible, but I couldn't figure it out?

Best regards
Oli

### LINKED ISSUE #193: Errors in home-assistant.log at start
Hi,

First of all, thanks for this great library.

I have a very specific case that makes home assistant to log an error at start.

I compute using HA, an average temperature which I then expose to KNX (for the use of an HVAC unit, but it is not the point here). BUT, on startup (before its default value is affected), this computed temperature have a state of "unknown". The expose to KNX process first try to convert this temperature and it fails (and log into HA.log file).

`2019-03-21 01:21:25 ERROR (MainThread) [homeassistant.core] Error doing job: Task exception was never retrieved
  File "/usr/lib/python3.5/asyncio/tasks.py", line 239, in _step
    result = coro.send(None)
  File "/srv/homeassistant/lib/python3.5/site-packages/homeassistant/components/knx/__init__.py", line 338, in _async_entity_changed
    await self.device.set(float(new_state.state))
ValueError: could not convert string to float: 'unknown'`

Generally, I edit the "homeassistant/components/knx/__init__.py" file and insert this at line 338 as an additional check :

`    if new_state.state == "unknown":
         return
` (sorry return don't want to go to next line in github editor)

I wonder if it's possible to add this check in the standard code ?

Thanks a lot !

### LINKED ISSUE #194: Humidity sensor is not displayed in HomeKit
**Description of problem:**

I have a humidity sensor that sends onto the KNX bus every 10 sec. This works fine and is also picked up fine in Home-Assistant.

However, the sensor is not displayed in HomeKit, possibly because Home-Assistant does not see it as a humidity sensor. The HomeKit documentation says this for humidity sensors:
"All sensors that have % as their unit_of_measurement **and** humidity as their device_class." (https://www.home-assistant.io/components/homekit/)

I also have a temperature sensor on KNX that works fine in HomeKit because the HomeKit documentation says this: "All sensors that have Celsius or Fahrenheit as their unit_of_measurement **or** temperature as their device_class."

Emphasis mine both times.

It seems that xknx is not specifying the `device_class` for its sensors and therefore HomeKit picks up only the temperature sensor (because it's measuring °C) but not the humidity sensor.

- [ ] using xknx standalone

**Version information:**
- xknx / Home-Assistant release with the issue: 0.90.1
- last working xknx / Home-Assistant release (if known): N/A

**KNX installation:**
Mainly ABB and some selfmade KNXnet/IP devices using my ESP8266 library

```
sensors:
  - platform: knx
    name: wznico.temp
    type: temperature
    address: '0/1/1'
  - platform: knx
    name: wznico.hum
    type: humidity
    address: '0/1/2'

```

farmio:
- RemoteValueSensor() (Sensor() and ExposeSensor() ) accepts new value types:
  - `angle` DPT 5.003
  - `binary` mapped to `RemoteValueSwitch`
  - `counter_pulses` DPT 6.010
  - `luminous_flux` DPT 14.042
  - `percentU8`DPT 5.004 (1 byte unscaled)
  - `percentV8` DPT 6.001 (1 byte signed unscaled)
  - `pressure` DPT 14.058
  - `scene_number` DPT 17.001
  - `string` DPT 16.000
- Sensor() value type `percent` is now using RemoteValueSensor() with DPTScaling instead of RemoteValueScaling() and behaves like every other type now (it was previously inverted when exposed 🙃)
- the payload of `RemoteValueSwitch` is boolean instead of Enum
- hass plugin doesn't process values of "unknown" in _async_entity_changed() (this string will not work with the `string` value type)
- hass plugin accepts boolean (string "on"/"off") values in _async_entity_changed()
- hass plugin sets device_class for some supported sensors

Internal changes:
- use more pythonic way of types in DPT calsses: eg: `try: int(value)` instead of `isinstance(value, int)` so types are inferred and HAs way of passing everything as a string works for all expose sensors (previously every value was converted to float - so int and string types raised an error)
- removed dpt_scene_number.py in favor of subclassing DPTValue1UCount
- added dpt_scaling.py with DPTAngle and DPTScaling
- added sensor_expose_loop_test.py for testing Sensors and passing the value to an ExposeSensor again to make sure the telegrams match

This PR is linked to following issues: 
fixes #194 
closes #193  
closes #129  
closes #116 
fixes #114  
kind of resolves home-assistant/home-assistant#21737 (https://github.com/XKNX/xknx/pull/189/commits/988060e3462a1b1e591cf7f473075552fd94b401)
https://github.com/home-assistant/home-assistant/issues/21737


xknx/devices/switch.py
```diff
@@ -50,7 +50,8 @@ def has_group_address(self, group_address):
     @property
     def state(self):
         """Return the current switch state of the device."""
-        return self.switch.value == RemoteValueSwitch.Value.ON
+        # None will return False
+        return bool(self.switch.value)
```
farmio: Usually switch and light.state (and probably climate but there is no such test for the on state) is initialized with `None`. Until now `return self.switch.value == RemoteValueSwitch.Value.ON` has returned `False` for `None`.

Technically I would say the new implementation is correct - we don't know the state so it is None but I have no idea if this breaks things. 

Alternatively we could return `False` for `None` (in the device)
```
def state(self):
    return bool(self.device.value)
```

farmio:
`scene_number` currently returns the raw value (0x01 -> 1). In ETS (and I think in most actor applications) scene numbers are started by 1 (0x00 -> 1). 
I did **not** account for this now for backward compatibility of `DPTSceneNumber` but I'm open for discussion.

home-assistant-plugin/custom_components/xknx/sensor.py
```diff
@@ -103,6 +105,20 @@ def unit_of_measurement(self):
...
+            return DEVICE_CLASS_HUMIDITY
+        if self.device.sensor_value.value_type in \
+                ("illuminance", 'luminous_flux'):
+            return DEVICE_CLASS_ILLUMINANCE
+        if self.device.sensor_value.value_type == "pressure":
+            return DEVICE_CLASS_PRESSURE
+        if self.device.sensor_value.value_type == "temperature":
+            return DEVICE_CLASS_TEMPERATURE
+        return None
+
```
farmio: device_class for some value types. #194 
Since it is only relevant for home assistant I put it here. If we defined it in the corresponding DPTArray subclasses we would have to use strings or enums and match the HA constant here.

home-assistant-plugin/custom_components/xknx/__init__.py
```diff
@@ -341,4 +341,13 @@ def async_register(self):
...
+        if new_state.state == "unknown":
+            return
+
+        if self.type == 'binary':
+            if new_state.state == "on":
+                await self.device.set(True)
+            elif new_state.state == "off":
+                await self.device.set(False)
+        else:
+            await self.device.set(new_state.state)
```
farmio: don't convert every new value to float. DPTArray subclasses expecting `int` didn't work. This way `string` is possible too.

test/remote_value_dpt_2_byte_unsigned_test.py
```diff
@@ -33,13 +33,13 @@ def test_from_knx(self):
...
     def test_to_knx_error(self):
-        """Test to_knx function with wrong parametern."""
+        """Test to_knx function with wrong parameters."""
         xknx = XKNX(loop=self.loop)
         remote_value = RemoteValueDpt2ByteUnsigned(xknx)
         with self.assertRaises(ConversionError):
             remote_value.to_knx(65536)
         with self.assertRaises(ConversionError):
-            remote_value.to_knx("256")
+            remote_value.to_knx("a")
```
farmio: with the new DPT to_knx function `"256"` gets converted to int.

test/str_test.py
```diff
@@ -226,7 +226,7 @@ def test_expose_sensor(self):
         self.loop.run_until_complete(asyncio.Task(sensor.set(25)))
         self.assertEqual(
             str(sensor),
-            '<ExposeSensor name="MeinSensor" sensor="GroupAddress("1/2/3")/None/<DPTArray value="[0xbf]" />/25" value="25" unit="%"/>')
+            '<ExposeSensor name="MeinSensor" sensor="GroupAddress("1/2/3")/None/<DPTArray value="[0x40]" />/25" value="25" unit="%"/>')
```
farmio: percent exposed sensors were inverted - and so were the tests.

xknx/devices/expose_sensor.py
```diff
@@ -30,8 +30,8 @@ def __init__(self,
         super(ExposeSensor, self).__init__(xknx, name, device_updated_cb)
 
         self.sensor_value = None
-        if value_type == "percent":
-            self.sensor_value = RemoteValueScaling(
+        if value_type == "binary":
+            self.sensor_value = RemoteValueSwitch(
```
farmio: percent is treated like every other sensor now; binary is extra because of DPTBinary values.

xknx/devices/remote_value_scaling.py
```diff
@@ -17,8 +17,8 @@ def __init__(self,
                  group_address_state=None,
                  device_name=None,
                  after_update_cb=None,
-                 range_from=100,
-                 range_to=0):
+                 range_from=0,
+                 range_to=100):
```
farmio: this was the reason exposed percent values were inverted.

xknx/knx/dpt.py
```diff
@@ -64,6 +64,9 @@ class DPTBinary(DPTBase):
 
     def __init__(self, value):
         """Initialize DPTBinary class."""
+        # TODO: do we need this?
+        # if isinstance(value, bool):
+        #     value = int(value)
```
farmio: I guess not...
```
>>> a = True
>>> isinstance(a, int)
True
>>> a = False
>>> isinstance(a, int)
True
```

xknx/knx/dpt_float.py
```diff
@@ -153,6 +156,7 @@ class DPTTemperature(DPT2ByteFloat):
     value_min = -273
     value_max = 670760
     unit = "°C"
+    ha_device_class = "temperature"
```
farmio: here the home assistant device_class strings are assigned  for corresponding DPT  #194

xknx/knx/dpt_2byte.py
```diff
@@ -29,11 +29,13 @@ def from_knx(cls, raw):
...
-        if not cls._test_boundaries(value):
-            raise ConversionError("Cant serialize DPT2ByteUnsigned", value=value)
-        return value >> 8, value & 0xff
+        try:
+            knx_value = int(value)
+            if not cls._test_boundaries(knx_value):
+                raise ValueError
+            return knx_value >> 8, knx_value & 0xff
+        except ValueError:
+            raise ConversionError("Cant serialize %s" % cls.__name__, value=value)
```
farmio: This (and the same pattern in the other dpt_*.py) is the heart of the PR ❤️

xknx/devices/remote_value_scaling.py
marvin-w: nice catch, that really did hurt :/

marvin-w:
This looks good to me. I'll merge it.

PR #190, opened 2019-03-09
Climate setpoint shift fix

### LINKED ISSUE #148: Startup Warnings  with Climate component on HA
It look  like every time HA starts, xknx raises a warning for each climate entity as follows:

```
Could not read value of <Climate name="Master Bathroom" temperature="GroupAddress("4/6/10")/None/<DPTArray value="[0xd,0x8]" />/25.76"  target_temperature="GroupAddress("4/5/10")/None/<DPTArray value="[0xc,0x1a]" />/21.0"  setpoint_shift="GroupAddress("3/3/10")/GroupAddress("4/3/10")/<DPTArray value="[0x0]" />/0" setpoint_shift_step="0.5" setpoint_shift_max="6" setpoint_shift_min="-6" group_address_operation_mode="GroupAddress("3/1/10")" group_address_operation_mode_state="None" group_address_controller_status="None" group_address_controller_status_state="GroupAddress("4/0/10")" /> 3/1/10
```

It looks like we need to skip checking operation_mode_state if controller_status_state is defined.

### LINKED ISSUE #183: Climate devices with setpoint_shift
While testing #175 @Julius2342 encountered below error which needs to be analysed and properly fixed.

```
2019-02-19 09:51:47 ERROR (MainThread) [homeassistant.core] Error doing job: Task exception was never retrieved
  File "/home/julius/home-assistant/homeassistant/helpers/service.py", line 291, in _handle_service_platform_call
    await func(entity, data)
  File "/home/julius/home-assistant/homeassistant/components/climate/__init__.py", line 564, in async_service_temperature_set
    await entity.async_set_temperature(**kwargs)
  File "/home/julius/home-assistant/homeassistant/components/knx/climate.py", line 238, in async_set_temperature
    await self.device.set_target_temperature(temperature)
  File "/home/julius/xknx/xknx/devices/climate.py", line 180, in set_target_temperature
    await self.set_target_temperature_setpoint_shift(target_temperature)
  File "/home/julius/xknx/xknx/devices/climate.py", line 191, in set_target_temperature_setpoint_shift
    raise DeviceIllegalValue("setpoint_shift_max exceeded", setpoint_shift)
xknx.exceptions.exception.DeviceIllegalValue: <DeviceIllegalValue description="setpoint_shift_max exceeded" value="68" />

```

farmio:
- Climate().target_temperature is now a property. Its implementation abstracts the usage of a writeable _target_temperature RemoteValueTemp() or a read only one with setpoint_shift.
- ~_base_temperature RemoteValueTemp() is introduced to reflect an immutable base temperature used together with setpoint_shift. It can be read from the bus if the actor supports it.~ (removed - no  known actor supports this)
- Climate().base_temperature property reflects the default temperature (setpoint-shift=0) for the active climate mode.
 
This PR should 
- resolve #183
- resolve #148 

In HA target_temperature_state_address is needed for this PR to work properly. 
Also the corresponding state addresses for climate-modes need to be set now.
https://github.com/home-assistant/home-assistant/pull/21541

### DIFF
--- a/home-assistant-plugin/custom_components/xknx/climate.py
+++ b/home-assistant-plugin/custom_components/xknx/climate.py
@@ -57,8 +57,8 @@ OPERATION_MODES_INV = dict((
 PLATFORM_SCHEMA = PLATFORM_SCHEMA.extend({
     vol.Optional(CONF_NAME, default=DEFAULT_NAME): cv.string,
     vol.Required(CONF_TEMPERATURE_ADDRESS): cv.string,
-    vol.Required(CONF_TARGET_TEMPERATURE_STATE_ADDRESS): cv.string,
     vol.Optional(CONF_TARGET_TEMPERATURE_ADDRESS): cv.string,
+    vol.Required(CONF_TARGET_TEMPERATURE_STATE_ADDRESS): cv.string,
     vol.Optional(CONF_SETPOINT_SHIFT_ADDRESS): cv.string,
     vol.Optional(CONF_SETPOINT_SHIFT_STATE_ADDRESS): cv.string,
     vol.Optional(CONF_SETPOINT_SHIFT_STEP,
@@ -211,12 +211,12 @@ class KNXClimate(ClimateDevice):
     @property
     def target_temperature_step(self):
         """Return the supported step of target temperature."""
-        return self.device.setpoint_shift_step
+        return self.device.temperature_step
 
     @property
     def target_temperature(self):
         """Return the temperature we try to reach."""
-        return self.device.target_temperature
+        return self.device.target_temperature.value
 
     @property
     def min_temp(self):
--- a/home-assistant-plugin/custom_components/xknx/climate.py
+++ b/home-assistant-plugin/custom_components/xknx/climate.py
@@ -246,9 +246,10 @@ class KNXClimate(ClimateDevice):
     @property
     def operation_list(self):
         """Return the list of available operation modes."""
-        return [OPERATION_MODES.get(operation_mode.value) for
-                operation_mode in
-                self.device.mode.operation_modes]
+        _operations = [OPERATION_MODES.get(operation_mode.value) for
+                       operation_mode in
+                       self.device.mode.operation_modes]
+        return list(filter(None, _operations))
 
     async def async_set_operation_mode(self, operation_mode):
         """Set operation mode."""
--- a/home-assistant-plugin/custom_components/xknx/climate.py
+++ b/home-assistant-plugin/custom_components/xknx/climate.py
@@ -216,7 +216,7 @@ class KNXClimate(ClimateDevice):
     @property
     def target_temperature(self):
         """Return the temperature we try to reach."""
-        return self.device.target_temperature.value
+        return self.device.target_temperature
 
     @property
     def min_temp(self):
--- a/home-assistant-plugin/custom_components/xknx/climate.py
+++ b/home-assistant-plugin/custom_components/xknx/climate.py
@@ -56,11 +56,6 @@ OPERATION_MODES_INV = dict((
 
 PLATFORM_SCHEMA = PLATFORM_SCHEMA.extend({
     vol.Optional(CONF_NAME, default=DEFAULT_NAME): cv.string,
-    vol.Required(CONF_TEMPERATURE_ADDRESS): cv.string,
-    vol.Optional(CONF_TARGET_TEMPERATURE_ADDRESS): cv.string,
-    vol.Required(CONF_TARGET_TEMPERATURE_STATE_ADDRESS): cv.string,
-    vol.Optional(CONF_SETPOINT_SHIFT_ADDRESS): cv.string,
-    vol.Optional(CONF_SETPOINT_SHIFT_STATE_ADDRESS): cv.string,
     vol.Optional(CONF_SETPOINT_SHIFT_STEP,
                  default=DEFAULT_SETPOINT_SHIFT_STEP): vol.All(
                      float, vol.Range(min=0, max=2)),
@@ -68,6 +63,11 @@ PLATFORM_SCHEMA = PLATFORM_SCHEMA.extend({
         vol.All(int, vol.Range(min=0, max=32)),
     vol.Optional(CONF_SETPOINT_SHIFT_MIN, default=DEFAULT_SETPOINT_SHIFT_MIN):
         vol.All(int, vol.Range(min=-32, max=0)),
+    vol.Required(CONF_TEMPERATURE_ADDRESS): cv.string,
+    vol.Required(CONF_TARGET_TEMPERATURE_STATE_ADDRESS): cv.string,
+    vol.Optional(CONF_TARGET_TEMPERATURE_ADDRESS): cv.string,
+    vol.Optional(CONF_SETPOINT_SHIFT_ADDRESS): cv.string,
+    vol.Optional(CONF_SETPOINT_SHIFT_STATE_ADDRESS): cv.string,
     vol.Optional(CONF_OPERATION_MODE_ADDRESS): cv.string,
     vol.Optional(CONF_OPERATION_MODE_STATE_ADDRESS): cv.string,
     vol.Optional(CONF_CONTROLLER_STATUS_ADDRESS): cv.string,
--- a/xknx/devices/climate.py
+++ b/xknx/devices/climate.py
@@ -32,6 +32,12 @@ class SetpointShiftValue(RemoteValue1Count):
                  max_kelvin=DEFAULT_SETPOINT_SHIFT_MAX):
         """Initialize SepointShift class."""
         # pylint: disable=too-many-arguments
+        super().__init__(xknx,
+                         group_address,
+                         group_address_state,
+                         device_name,
+                         after_update_cb)
+
         self.setpoint_shift_step = setpoint_shift_step
         self.min_kelvin = min_kelvin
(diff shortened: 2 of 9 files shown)


xknx/devices/climate.py
```diff
@@ -13,22 +13,65 @@
...
+        """Return current value in Kelvin."""
+        if super().value is not None:
+            return super().value * self.setpoint_shift_step
+        return None
+
+    async def set(self, value):
+        """Set new value from Kelvin."""
+        if value > self.max_kelvin:
+            raise DeviceIllegalValue("setpoint_shift_max exceeded", value)
+            # value = self.max_kelvin
```
farmio: we could log a warning and move on with the maximum value.

xknx/devices/climate.py
```diff
@@ -13,27 +13,78 @@
...
+    """Class for managing setpoint_shift values."""
+
+    def __init__(self,
+                 xknx,
+                 group_address=None,
+                 group_address_state=None,
+                 device_name=None,
+                 after_update_cb=None,
+                 setpoint_shift_step=DEFAULT_SETPOINT_SHIFT_STEP,
+                 min_kelvin=DEFAULT_SETPOINT_SHIFT_MIN,
```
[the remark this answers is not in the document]
farmio: -> setpoint_shift.min_kelvin to differentiate between the actual bus value and the human readable value. (Kelvin is commonly used for temperature changes or differences instead of Celsius)

xknx/devices/climate.py
```diff
@@ -59,22 +110,27 @@ def __init__(self,
...
             after_update_cb=self.after_update)
-        self.target_temperature = RemoteValueTemp(
+        self._target_temperature = RemoteValueTemp(
             xknx,
             group_address_target_temperature,
             group_address_target_temperature_state,
             device_name=self.name,
             after_update_cb=self.after_update)
-        self.setpoint_shift = RemoteValue1Count(
+        self._base_temperature = RemoteValueTemp(
```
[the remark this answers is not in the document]
farmio: > _base_temperature RemoteValueTemp() is introduced to reflect an immutable base temperature used together with setpoint_shift. It can be read from the bus if the actor supports it.

I dont even know if there are actors that provide this value as communication object. I will edit it out again to remove the bloat.
_target_temperature can then be changed to target_temperature again.

xknx/devices/climate.py
Julius2342: I think this makes sense.

xknx/devices/climate.py
```diff
@@ -164,63 +208,74 @@ def is_on(self):
...
+        """Return current offset from base temperature in Kelvin."""
+        return self._setpoint_shift.value
+
+    async def set_setpoint_shift(self, offset):
+        """Send new temperature offset to KNX bus."""
+        base_temperature = self.base_temperature
+        await self._setpoint_shift.set(offset)
+        # broadcast new target temperature and set internally
+        if base_temperature is not None:
+            await self.target_temperature.set(base_temperature + self.setpoint_shift)
```
farmio: this yields 
`[xknx.log] Error while processing telegram <ConversionError description="dst_add not set" />`
when using group_address_target_temperature_state and no  group_address_target_temperature.

marvin-w: The actor will send an update if the value changes thus I don't think setting it there is neccessary

Julius2342: In my system the update from the bus is send immediately, so I tend to say an update is not necessary. 

But if we do so, we should only try to broadcast the target temperature if a set address is set.

farmio: I was thinking about defining a `writeable` property in RemoteValue() tocheck this.

xknx/devices/climate.py
```diff
@@ -164,63 +208,74 @@ def is_on(self):
...
-        """Calculate setpoint shift shift and send it to  KNX bus."""
+    @property
+    def temperature_step(self):
+        """Return smallest possible temperature step."""
         if self.initialized_for_setpoint_shift_calculations:
-            await self.set_target_temperature_setpoint_shift(target_temperature)
-        # broadcast new target temperature and set internally
-        await self.target_temperature.set(target_temperature)
+            return self._setpoint_shift.setpoint_shift_step
+        return 0.1
```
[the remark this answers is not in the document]
farmio: If we returned None (for temperature_address actors) HA would default to a 0.5 resolution. This could behave awkward if other devices on the KNX bus support 0.1 resolution.
We could either 
- return 0.1 per default (not really elegant)
- check for None in the HA plugin and return 0.1 there (even less elegant) or 
- change `setpoint_shift_step` to something like `temperature_resolution` and use it for temperature_address actors and setpoint_shift actors alike

xknx/devices/climate.py
```diff
@@ -13,27 +13,78 @@
 from .remote_value_switch import RemoteValueSwitch
 from .remote_value_temp import RemoteValueTemp
 
+DEFAULT_SETPOINT_SHIFT_STEP = 0.5
+DEFAULT_SETPOINT_SHIFT_MAX = 6
+DEFAULT_SETPOINT_SHIFT_MIN = -6
+
+
```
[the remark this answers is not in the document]
farmio: I have reduced the size of this class. It doesn't make sense to use this in other modules than climate.py . Shall we still outsource it?

xknx/devices/climate.py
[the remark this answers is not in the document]
farmio: it returns DEFAULT_TEMPERATURE_STEP now. Are you ok with this or shall we move this to the HA component?

xknx/devices/climate.py
```diff
@@ -165,63 +210,81 @@ def is_on(self):
...
+    def setpoint_shift(self):
+        """Return current offset from base temperature in Kelvin."""
+        return self._setpoint_shift.value
+
+    async def set_setpoint_shift(self, offset):
+        """Send new temperature offset to KNX bus."""
+        base_temperature = self.base_temperature
+        await self._setpoint_shift.set(offset)
+        # broadcast new target temperature and set internally
+        if self.target_temperature.writable and \
```
marvin-w: s/writable/writeable/g

xknx/devices/climate.py
marvin-w: I like the idea of the `writeable` property. 

On the other hand it makes the code more complex. Shouldn't this generally be the case? Can we not just check in `RemoteValue.set()` if the there is a GA defined for writing ?

xknx/devices/climate.py
```diff
@@ -13,16 +12,61 @@
 from .remote_value_switch import RemoteValueSwitch
 from .remote_value_temp import RemoteValueTemp
 
+DEFAULT_SETPOINT_SHIFT_MAX = 6
+DEFAULT_SETPOINT_SHIFT_MIN = -6
+DEFAULT_SETPOINT_SHIFT_STEP = 0.5
+DEFAULT_TEMPERATUE_STEP = 0.1
+
+
+class SetpointShiftValue(RemoteValue1Count):
```
marvin-w: good job, I like this!

xknx/devices/climate.py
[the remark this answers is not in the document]
marvin-w: For me `DEFAULT_TEMPERATURE_STEP` would be enough, but lets wait for @Julius2342

xknx/devices/climate.py
farmio: I like the idea of checking for writable in RemoteValue.set(). I would log a warning instead of an info if it is called for a non-writable device. 
To suppress this warning I'd still check for writable in climate.py.

xknx/devices/climate.py
farmio: Thats what I had initially. Then I had a long google for this. Now I'm unsure again.
Maybe its like tunnelling / tunneling

marvin-w: Leave it - both is correct :)

home-assistant-plugin/custom_components/xknx/climate.py
```diff
@@ -57,8 +57,8 @@
 PLATFORM_SCHEMA = PLATFORM_SCHEMA.extend({
     vol.Optional(CONF_NAME, default=DEFAULT_NAME): cv.string,
     vol.Required(CONF_TEMPERATURE_ADDRESS): cv.string,
-    vol.Required(CONF_TARGET_TEMPERATURE_STATE_ADDRESS): cv.string,
     vol.Optional(CONF_TARGET_TEMPERATURE_ADDRESS): cv.string,
+    vol.Required(CONF_TARGET_TEMPERATURE_STATE_ADDRESS): cv.string,
```
marvin-w: can you change the order back please? this is their code style 

1. optionals with default
2. required
3. truely optional

farmio: done

xknx/devices/remote_value.py
```diff
@@ -98,8 +103,12 @@ def value(self):
     async def set(self, value):
         """Set new value."""
         if not self.initialized:
-            self.xknx.logger.info("Setting value of uninitialized device %s (value %s)", self.device_name, value)
+            self.xknx.logger.info("Setting value of uninitialized device: %s (value: %s)", self.device_name, value)
             return
+        if not self.writable:
+            self.xknx.logger.warning("Atempted to set value for non-writable device: %s (value: %s)", self.device_name, value)
```
marvin-w: s/Atempted/Attempted/g

PR #191, opened 2019-03-10
Add support for DPT 251.600 (6 byte RGBW).

phbaer:
Maybe that additional DPT for lights might be useful to others as well. Tested with MDT SCN-DALI16.03.


marvin-w:
LGTM besides the fact that the light class is going to blow up.

Maybe @farmio can have a look.

farmio:
Looks good to me too, but I have no   colored lights so can't test it ¯\_(ツ)_/¯ 
I think we would have all knx light possibilities covered now 🤓

Adding it to the hass-component would be fine.

phbaer:
Blow up: yes, unfortunately. Also though of merging it with the generic colour propery but ... well, it's basically a low-level interface to KNX and hence different DPTs should not be hidden in that case.

The HASS component is on my list. Need to find some spare time later this week.

phbaer:
As the integration in home assistant is easier with a combined RGB/RGBW interface, I changed the xknx light implementation accordingly. This slightly reduces the light.py blowup...

Also the underlying DPT abstraction now only returns the colour+white values and keeps the management bits hidden. If a colour or the white value are invalid, they'll be set to the previously received value or 0.

I'd be done for now.

PR #198, opened 2019-04-18
index scene numbers with 1 instead of 0

farmio:
From the KNX Specifications v2.1

> DPT_SceneControl allows numbering the scene from 0 to 63. KNX Association recommends displaying these scene numbers in ETS(TM), other software and controllers numbered from 1 to 64, this is, with an offset of 1 compared to the actual transmitted value.

It breaks current implementations of `Scene()` devices so I'm not sure if merging this is a good idea or if it would be better to just ignore this. I figured now that the `scene_number` sensor type is merged it is a good time to discuss this.

### DIFF
--- a/test/dpt_value_1_ucount_test.py
+++ b/test/dpt_value_1_ucount_test.py
@@ -2,7 +2,7 @@
 import unittest
 
 from xknx.exceptions import ConversionError
-from xknx.knx import DPTSceneNumber, DPTValue1Ucount
+from xknx.knx import DPTValue1Ucount
 
 
 class TestDPTValue1Ucount(unittest.TestCase):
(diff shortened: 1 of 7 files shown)


test/dpt_value_1_ucount_test.py
```diff
@@ -54,24 +54,3 @@ def test_from_knx_wrong_parameter2(self):
         """Test parsing of DPTValue1Ucount with wrong value (array containing string)."""
         with self.assertRaises(ConversionError):
             DPTValue1Ucount().from_knx(("0x23"))
```
farmio: this is a duplicate of test/dpt_scene_number_test.py

Julius2342: Might have been just a copy & paste error!

Julius2342:
I'm in favor of this change (life is hard). We should just not forget to add a comment to the PR within HASS.

PR #199, opened 2019-04-26
read binary sensors on startup

### LINKED ISSUE #103: There is need for a KNX read-request-service
Hi
Some KNX physical sensors don´t send telegrams as often as you would like. Or, sometimes there is need for requesting a status/telegram to be sent by the physical sensor to the bus. 
The current "payload" request only sends knx_write, not read and the sensors dont respond to write-requests.

I had this problem and I found a smart guy that had shared a custom component that works (I have tested and use this as of now)

https://gist.github.com/ct-23/b707f3c2d8406c8697a25a3ce619f3c2#file-knx_reader-py

It would really improve HA KNX functionality if a similar knx_read-service was implemented. 

Best
/PH

farmio:
- rebased HA plugin from HA 0.92.0 (mainfest.json)
- removed group_address form BinarySensor in favor for group_address_state
- replaced group_address in Sensor with group_address_state

Sensor and BinarySensor now have `group_address_state` as their only group address attributes as they are read only devices. With this change binary sensors are read on startup ( #140 ).

This resolves #103

### DIFF
--- a/docs/home_assistant.md
+++ b/docs/home_assistant.md
@@ -41,19 +41,6 @@ Delete automatically installed version:
 rm .homeassistant/deps/lib/python3.5/site-packages/xknx*
 ```
 
-Comment out dependency from xknx.py, otherwise hass would reinstall the xknx library when startup:
-
-````diff
-julius@xxx ~/xknx:$ git diff -U0 home-assistant-plugin/custom_components/xknx.py 
-diff --git a/home-assistant-plugin/custom_components/xknx.py b/home-assistant-plugin/custom_components/xknx.py
-index 7f26c43..0b1e046 100644
---- a/home-assistant-plugin/custom_components/xknx.py
-+++ b/home-assistant-plugin/custom_components/xknx.py
-@@ -39 +39 @@ _LOGGER = logging.getLogger(__name__)
--REQUIREMENTS = ['xknx==0.7.18']
-+# REQUIREMENTS = ['xknx==0.7.18']
-```
-
 Ideally start hass from command line. Export the environment variable PYTHONPATH to your local xknx checkout:
 
 ```bash
--- a/docs/sensor.md
+++ b/docs/sensor.md
@@ -34,7 +34,7 @@ Sensor objects are usually configured via [`xknx.yaml`](/configuration):
     sensor:
         Heating.Valve1: {group_address_state: '2/0/0', value_type: 'percent'}
         Heating.Valve2: {group_address_state: '2/0/1', value_type: 'percent'}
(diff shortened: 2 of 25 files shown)


xknx/devices/sensor.py
```diff
@@ -16,7 +16,7 @@ class Sensor(Device):
     def __init__(self,
                  xknx,
                  name,
-                 group_address=None,
+                 group_address_state=None,
```
farmio: Should we maybe keep `group_address` for a while and raise some sort of depreciation warning?

docs/home_assistant.md
```diff
@@ -41,19 +41,6 @@ Delete automatically installed version:
...
-julius@xxx ~/xknx:$ git diff -U0 home-assistant-plugin/custom_components/xknx.py 
-diff --git a/home-assistant-plugin/custom_components/xknx.py b/home-assistant-plugin/custom_components/xknx.py
-index 7f26c43..0b1e046 100644
---- a/home-assistant-plugin/custom_components/xknx.py
-+++ b/home-assistant-plugin/custom_components/xknx.py
-@@ -39 +39 @@ _LOGGER = logging.getLogger(__name__)
--REQUIREMENTS = ['xknx==0.7.18']
-+# REQUIREMENTS = ['xknx==0.7.18']
-```
-
```
farmio: This isn't part of home-assistant-plugin/custom_components/xknx.py anymore. 
It loaded my checked out version of xknx without removing the requirement from manifest.json. Can someone confirm this?

xknx/devices/sensor.py
Julius2342: I think this is fine.

docs/sensor.md
```diff
@@ -32,10 +32,10 @@ Sensor objects are usually configured via [`xknx.yaml`](/configuration):
 
 ```yaml
     sensor:
-        Heating.Valve1: {group_address: '2/0/0', value_type: 'percent'}
-        Heating.Valve2: {group_address: '2/0/1', value_type: 'percent'}
-	Kitchen.Temperature: {group_address: '2/0/2', value_type: 'temperature'}
-        Some.Other.Value: {group_address: '2/0/3'}
+        Heating.Valve1: {group_address_state: '2/0/0', value_type: 'percent'}
+        Heating.Valve2: {group_address_state: '2/0/1', value_type: 'percent'}
+	Kitchen.Temperature: {group_address_state: '2/0/2', value_type: 'temperature'}
```
Julius2342: :-O

docs/home_assistant.md
Julius2342: You need a current version of HASS. They now read the dependencies from manifest: https://github.com/home-assistant/home-assistant/pull/22699

PR #203, opened 2019-05-16
Tests and requirement updates

farmio:
- some additional unittests
- testing on Python 3.7 instead of 3.6-dev
- require pyyaml 5.1 to avoid deprecation warnings
- moved tests in separate directories


.gitignore
```diff
@@ -13,4 +13,5 @@ xknx.egg-info/
 venv
 .idea/
 .vscode/
-.sass-cache/
\ No newline at end of file
+.sass-cache/
+.pytest_cache/
```
marvin-w: From personal preference I'd prefer using newlines but its up to you

farmio: oh I'm just using VSCodes default here ¯\_(ツ)_/¯

PR #216, opened 2019-07-04
Optionally disable reading (GroupValueRead) for sensor and binary_sensor

farmio:
- possiblity to disable reading (GroupValueRead) for sensor and binary_sensor
- move from `address` to `state_address` in HA components sensor and binary_sensor

### DIFF
--- a/test/core_tests/config_test.py
+++ b/test/core_tests/config_test.py
@@ -376,6 +376,18 @@ class TestConfig(unittest.TestCase):
                    'Heating.Valve1',
                    group_address_state='2/0/0',
                    value_type='percent',
+                   sync_state=True,
+                   device_updated_cb=TestConfig.xknx.devices.device_updated))
+
+    def test_config_sensor_percent_passive(self):
+        """Test passive percent Sensor from config file."""
+        self.assertEqual(
+            TestConfig.xknx.devices['Heating.Valve2'],
+            Sensor(TestConfig.xknx,
+                   'Heating.Valve2',
+                   group_address_state='2/0/1',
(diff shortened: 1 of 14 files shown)


test/core_tests/config_test.py
```diff
@@ -376,6 +376,18 @@ def test_config_sensor_percent(self):
                    'Heating.Valve1',
                    group_address_state='2/0/0',
                    value_type='percent',
+                   sync_state=True,
```
Julius2342: Isnt this the default value? (feel free to keep or remove it)

xknx/devices/remote_value.py
```diff
@@ -28,16 +29,22 @@ def __init__(self,
...
+        self.after_update_cb = after_update_cb
         self.payload = None
 
     @property
     def initialized(self):
         """Evaluate if remote value is initialized with group address."""
         return bool(self.group_address_state or self.group_address)
 
+    @property
+    def readable(self):
```
Julius2342: Hmm, do we find a more explicit name for this property? Not having any better suggestion, but I think this value is not self explaining.

test/core_tests/config_test.py
farmio: Yes it is.

xknx/devices/remote_value.py
farmio: I don't have a better name too. 
I think in the context of this module - the direct representation of a KNX Communication object - it's not that bad. I think of it as the pendant to the KNX Read Flag of a GA.

PR #223, opened 2019-08-11
added DPT 9.006 as pressure_2Byte sensor.

michelde:
with release 0.11.0 a new type 'pressure' was introduced for 4byte pressure values. But in KNX there are also 2Byte pressure values possible (DPT 9.006). 
Therefore I adjusted the coding and added the type `pressure_2Byte`. Not sure about the naming, maybe also `pressure_short` and `pressure_long` would work.

### DIFF
--- a/xknx/knx/dpt_2byte_float.py
+++ b/xknx/knx/dpt_2byte_float.py
@@ -133,3 +133,9 @@ class DPTEnthalpy(DPT2ByteFloat):
     """DPT 9.* 2-byte float value (with unit)."""
 
     unit = "H"
+
+class DPTPressure2Byte(DPT2ByteFloat):
+    """DPT 9.006 DPT_Value_Pressure."""
+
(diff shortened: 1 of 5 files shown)


xknx/knx/dpt_2byte_float.py
```diff
@@ -133,3 +133,10 @@ class DPTEnthalpy(DPT2ByteFloat):
     """DPT 9.* 2-byte float value (with unit)."""
 
     unit = "H"
+
+
+class DPTPressure2Byte(DPT2ByteFloat):
```
farmio: I'd move this to line 110 - between DPT 9.005 and 9.007 to have it in order.

defl:
Looking forward to this in a release :) Any indication on when this gets merged in?

farmio:
Is was included in the 0.11.2 release in September.

PR #225, opened 2019-08-13
Add new param on_off_invert.

tombbo:
Some KNX Room controller (RCD 20xx, 2178) have inverted power ON/OFF value. ON means disable controller, OFF means enable controller.

### DIFF
--- a/xknx/devices/climate.py
+++ b/xknx/devices/climate.py
@@ -83,7 +83,8 @@ class Climate(Device):
                  min_temp=None,
                  max_temp=None,
                  mode=None,
-                 device_updated_cb=None):
+                 device_updated_cb=None,
+                 on_off_inverted=False):
         """Initialize Climate class."""
         # pylint: disable=too-many-arguments, too-many-locals, too-many-branches, too-many-statements
         super().__init__(xknx, name, device_updated_cb)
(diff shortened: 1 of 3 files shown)


farmio:
Climate.on is a RemoteValueSwitch which has an attribute `invert` that defaults to false. I'd use this instead of changing the implementation of Climate.turn_on() to `self.on.off()`. No   need to touch Climate.is_on property then and better readability.

tombbo:
Changed to use an attribute invert of a RemoteValueSwitch.

xknx/devices/climate.py
```diff
@@ -83,7 +83,8 @@ def __init__(self,
                  min_temp=None,
                  max_temp=None,
                  mode=None,
-                 device_updated_cb=None):
+                 device_updated_cb=None,
+                 on_off_inverted=False):
```
farmio: Please move it to line 83 just after `group_address_on_off_state=None,` to have all on_off things together.

xknx/devices/climate.py
```diff
@@ -130,7 +131,8 @@ def __init__(self,
             group_address_on_off,
             group_address_on_off_state,
             device_name=self.name,
-            after_update_cb=self.after_update)
+            after_update_cb=self.after_update,
+            invert=on_off_inverted)
```
farmio: Consider renaming it to `on_off_invert` to be consistent with the RemoteValueSwitch and Cover attributes.

tombbo:
Done.

PR #227, opened 2019-08-20
Update states only when needed

farmio:
Change the state updater to live in every RemoteValue class.

Every Device should be configurable for one option of state updater:
- disable: never send GroupValueRead requests
- init: only issue GroupValueRead on startup
- expire: issue GroupValueRead on startup and when there was no incoming telegram on the RemoteValue since a configurable time
- periodically: issue GroupValueRead on startup and periodically at a configurable time span (maybe even cron like - every full hour etc.)

depends on #240


farmio:
@marvin-w I was just thinking about 2 possible implementations. 

1) return state_address only when last update is too long ago
```python
    # in RemoteValue
    def state_addresses(self, last_update_min=None):
        """Return group addresses which should be requested to sync state."""
        def _state_expired(minutes):
            return time.time() > (minutes * 60 + self.last_update)
        if self.readable:
            if last_update_min is None or \
                    _state_expired(last_update_min):
                return [self.group_address_state, ]
        return []
```
pro: easy to implement; breaks nothing
con: "state_addresses" probably needs a new name; last_update only checked every tick of StateUpdater

2) move the sync implementation to RemoteValue and create a recursive asyncio task on init that calls it. Maybe like one instance of StateUpdater per RemoteValue
pro: state updtes don't happen all at once for all devices; possibly individual sync timeouts per device; we could be sure no    value/payload is older than our threshold
con: adds a lot of tasks to the loop?

What do you think?

marvin-w:
Hi,

I personally like the second approach more if asyncio doesn't care about the additional tasks. I honestly think it should be fine. Also, it will add more possibilities as you already mentioned, so I personally would go for that. What do you think?

farmio:
I like it better too. It just came to my mind when I did the first approach.

farmio:
Well, I committed a first draft for the new state updater. Let me know what you think before I start to update the tests 🙃 

> 230 failed, 423 passed, 32 error in 38.51 seconds

xknx/devices/state_updater.py
```diff
@@ -0,0 +1,52 @@
...
+        # TODO: should we start from __init__ or manualy?
+        self.initial_start()
+
+    def start(self):
+        """Start StateUpdater."""
+        self._waiting_task = self.xknx.loop.create_task(
+            self._wait_for_expiration())
+
+    def initial_start(self):
+        # TODO: random(1-15) start timeout so not all devices are queried at once?
```
marvin-w: I would at a start timeout.

xknx/devices/state_updater.py
```diff
@@ -0,0 +1,52 @@
...
+                 expiration_min=60,
+                 callback=None):
+        """Initialize StateUpdater class."""
+        self.xknx = xknx
+        self.timeout = expiration_min * 60
+        self._waiting_task = None
+        # TODO: naming? async_callback?
+        # maybe save reference to RemoteValue instead of callback
+        self._callback = callback
+        # TODO: should we start from __init__ or manualy?
```
marvin-w: I would start directly from `__init__`

xknx/devices/state_updater.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self,
+                 xknx,
+                 expiration_min=60,
+                 callback=None):
+        """Initialize StateUpdater class."""
+        self.xknx = xknx
+        self.timeout = expiration_min * 60
+        self._waiting_task = None
+        # TODO: naming? async_callback?
```
marvin-w: I honestly don't mind, async_callback might be good to go

xknx/devices/state_updater.py
```diff
@@ -0,0 +1,52 @@
...
+    async def _wait_for_expiration(self):
+        """Waits for the timeout to expire. Endless loop for updating states."""
+        # on successfull reading the 2nd iteration of the while loop gets canceled when the callback calls reset()
+        # when no telegram was received it will try again endlessly
+        while True:
+            try:
+                await asyncio.sleep(self.timeout)
+            except asyncio.CancelledError:
+                return
+            # TODO: log name of remote value for debugging?
```
marvin-w: yes, please, logs should always contain as much information as possible. Since it's debugging we don't care if it's a lot.

xknx/devices/state_updater.py
```diff
@@ -0,0 +1,52 @@
+"""Module for keeping the value of a RemoteValue from KNX bus up to date."""
+import asyncio
+
+
+class StateUpdater():
+    """Class for reading the value of a RemoteValue from KNX bus."""
+
+    def __init__(self,
+                 xknx,
+                 expiration_min=60,
```
marvin-w: maybe just call it timeout?

xknx/devices/remote_value.py
```diff
@@ -32,14 +34,20 @@ def __init__(self,
...
-        self.sync_state = sync_state
         self.device_name = "Unknown" \
             if device_name is None else device_name
         self.after_update_cb = after_update_cb
 
-        self.last_update = None
         self.payload = None
 
+        # TODO: naming? unclear that it holds minutes
+        if sync_state:
```
marvin-w: name it different, maybe timeout?

xknx/devices/state_updater.py
marvin-w: or update_interval?

xknx/devices/state_updater.py
farmio: I have thought of naming the RemoteValues from Device after their feature (like "Light Kitchen - Brightness"). I think this would be nice in the logs.
Shall we 
- just store this name as an attribute in the StateUpdater or
- store a reference to the RemoteValue in the StateUpdater?
With the second option we could also omit `_callback and reach for `read_state()` directly. Would such a circular reference cause any problems?

xknx/devices/state_updater.py
farmio: 👍 I like update_interval! Very descriptive.

farmio:
Thank you Marvin!

marvin-w:
> Thank you Marvin!

No, thank you!

xknx/devices/state_updater.py
```diff
@@ -0,0 +1,52 @@
...
+    def stop(self):
+        """Stop StateUpdater."""
+        self._waiting_task.cancel()
+
+    def reset(self):
+        """Restart the timer."""
+        self.stop()
+        self.start()
+
+    async def _wait_for_expiration(self):
```
marvin-w: maybe call this update_loop or sth more descriptive?

xknx/devices/state_updater.py
marvin-w: I'd go for the first one. I don't think it's good design having a reference to the RemoteValue in the child object.

I don't think we need that `wait_for_result` parameter, either.

Add a parameter, name looks nice, but the actual group address provides way more information, so I'd go for that. Everyone will find their group address inside their project.

xknx/devices/state_updater.py
Julius2342: `update_interval` 👍

farmio:
I'm sorry, this has become quite big. I have merged #303, #305 and #273 now into this to go on.

Periodical StateUpdaterType isn't implemented yet, and I have to find a way to enable the StateUpdater (eg. XKNX.state_updater = True) by default but disable it for unittests.

farmio:
In the future I'd like to stop the StateUpdater when the connection drops and restart it when a new connection is made - re-initializing all RemoteValues.

marvin-w:
That sounds like a reasonable idea! I have reviewed #273 but I guess we should not merge it if it's included in this PR. 

Changesets that big make it really hard to review and test, potentially causing errors later on. I would generally prefer more smaller changesets but given the topic I don't think there is/was a way to do this.

Is there anything missing here? How can I test this?

farmio:
@marvin-w Thank you for taking a look! I know it's too big and I'm sorry 😐 This is why I did the RemoteValue PRs beforehand. If these were merged the diff would be much smaller here / but they are required for this to work. 
It is testable eh from HA `sensor` or `bibary_sensor` when you set eg `sync_state = expire 10` with this the sensor payload gets requested if it is older than 10 minutes.
other options are `init` this will only read once on init and `every 20` will request the state every 20 minutes from start regardless of in-between telegrams (like it was before).

marvin-w:
@farmio Alright! I can test this in my HA then. I didn't really have a look at all of the other PRs - sorry for that!

I try to give it a go but it can take some time due to the amount of work I have in my dayjob and other projects right now!

PR #232, opened 2019-09-01
module cleanup

farmio:
Introduced and renamed modules for better overview and readability.

- move RemoteValue and child classes out of `devices` to new `remote_value` module
- split `knx` module into `dpt` and `telegram` modules


farmio:
I'd like to make all Devices use RemoteValue classes. 
Currently BinarySensor, ClimateMode, DateTime, Notification use own implementations of some RemoteValue functions. This would make the Device class thinner and better readable. Also #227 would be far less complex to implement.

I think it would be easier to review when this task is split into multiple PRs. So this would be the first step.

Julius2342:
This is a reasonable move!

PR #246, opened 2019-10-04
Support proxied connections (Docker, NAT)

### LINKED ISSUE #215: Can't connect to Knx gateway behind NAT on VM
I am trying to connect to a knx ip gateway from a homeassistant vm but get a connection error. The vm is behind a NAT on the host.

KNX Gateway IP: 192.168.2.191
Host IP: 192.168.2.249
VM IP: 192.168.1.11

My homeassistant configuration
 <pre>Tunneling:
    Host: '192.168.2.191'
    Port: 3671
    Local_ip: '192.168.1.11'</pre>

Iptables configuration on the host
<pre> echo 1 > /proc/sys/net/ipv4/ip_forward
iptables -t nat -A PREROUTING -s '192.168.2.191' -j DNAT --to-destination '192.168.1.11' 
iptables -t nat -A POSTROUTING -d '192.168.2.191' -j SNAT --to-source '192.168.2.249' </pre>

On tcpdump I see an UDP packet going to Knx gateway but there is no reply packet.

farmio:
This PR adds support for forwarded connections like found on eg. Docker. 

KNX gets the source address for a connection from the ConnectRequest and does not use the IP-header. Therefore we have to send the IP address of the proxy in the ConnectRequest instead of the address xknx is using to send.

closes #215


sbias:
is there any update? if i understand correctly that pr would solve the problem to run home-assistent with knx in a docker-container. Is there any alternative solution for this - currently i the exactly this problem :-/

farmio:
You can always run the container with --net=host (network_mode = host in compose) which is the preferred option for the home-assistant Docker Container anyway.

spacegaier:
Is that already in a state to be tested, e.g. in a VirtualBox environment or still WIP? Could it be rebased onto 0.12.0 so that we can test with the same configuration file syntax as the latest HA dev does?

farmio:
Technically it works and should be easy to resolve current merge conflicts, but it is still in a proof of concept state. 
I'll see if I can update it in the next days...

farmio:
Closed in favor for #544

PR #251, opened 2019-11-10
Add support for knx data type DPT 9.002 to climate shiftpoint

arnonym:


home-assistant-plugin/custom_components/xknx/climate.py
```diff
@@ -93,6 +96,9 @@
         vol.Required(CONF_TEMPERATURE_ADDRESS): cv.string,
         vol.Required(CONF_TARGET_TEMPERATURE_STATE_ADDRESS): cv.string,
         vol.Optional(CONF_TARGET_TEMPERATURE_ADDRESS): cv.string,
+        vol.Optional(CONF_SETPOINT_SHIFT_MODE, default=DEFAULT_SETPOINT_SHIFT_MODE): cv.enum(
+            SetpointShiftMode
+        ),
```
farmio: Again HA has very strict styling guidelines. These shall be ordered the following way: Optionals with defaults > Required > Optionals without defaults. So I'd move it to line 87.

home-assistant-plugin/custom_components/xknx/climate.py
```diff
@@ -177,6 +183,7 @@ def async_add_entities_config(hass, config, async_add_entities):
         group_address_setpoint_shift_state=config.get(
             CONF_SETPOINT_SHIFT_STATE_ADDRESS
         ),
+        setpoint_shift_mode=config.get(CONF_SETPOINT_SHIFT_MODE),
```
farmio: Required and default values shall be accessed by `config[]` instead of `config.get()`

xknx/devices/climate.py
```diff
@@ -228,16 +207,19 @@ def initialized_for_setpoint_shift_calculations(self):
     def temperature_step(self):
         """Return smallest possible temperature step."""
         if self._setpoint_shift.initialized:
-            return self._setpoint_shift.setpoint_shift_step
-        return DEFAULT_TEMPERATUE_STEP
+            return self.setpoint_shift_step
```
farmio: I think in SetpointShiftMode.DPT9002 we should use a default of `0.1` not `0.5`
What is the default on MDT actors?

arnonym: The default of the "Glastaster II" is 0.5.

farmio: The heating actuator doesn't allow configuration of step when using 2byte...

arnonym: This is correct. Does it make more sense to use `DEFAULT_TEMPERATURE_STEP` for this mode? I think `DEFAULT_TEMPERATURE_STEP` and `DEFAULT_SETPOINT_SHIFT_STEP` are kinda redundant.

farmio: Yes your right, but 0.1 is required for direct (and in my opinion dpt9 sps). So we would have to change default value for sps dpt6 and this would break some default configuration.

arnonym: I do not understand what you mean with "required" for direct and dpt9 sps. Both methods can send whatever temperature/offset they want, if I understood correctly.

farmio: your right, its not really required... HA thermostat card uses this for their sliders. So when one has set eg. 21.3 from some other knx device it is awkward to have a slider in HA that can be set to 20.3, 20.8, 21.8, 22.3 ... therefore I would prefer 0.1 as a default.

arnonym: But the default of 0.1 will break the DPT6 SPS method. So what are we gonna do? Keep both and use `DEFAULT_TEMPERATURE_STEP` for DPT9 SPS and direct?

farmio: We can keep both until we introduce the next breaking change for Climate and clean this up then.

arnonym: So `DEFAULT_TEMPERATURE_STEP` for DPT9 SPS and direct, the other one for DPT7 SPS?

farmio: Then go for it, its configurable anyway. You could add a note to the documentation.

mqtt-fan:
Hi all,

what is the current issue, why we can't merge this pull request?

farmio:
@mqtt-fan Its still in draft and nobody has left feedback about testing it in a real environment. If you could test it and its working properly I'd be happy to merge it.

mqtt-fan:
I'm currently on it, but I'm facing some issue with the knx settings. I will let you know when it is working.

mqtt-fan:
@farmio  can you please merge it?  It is now working well for the last 16 days

EricReiche:
@farmio hey, can we get this merged please?

farmio:
I have merged this into the `next` branch. f391cc7109352bbb969b8a0a298f00291f7d58b1

PR #253, opened 2019-12-09
[WIP] Add support for other APCI services

basilfx:
This PR adds support for additional APCI services. APCI services are the actual payloads to communicate with actors. These payloads can be used for broadcast messages such as GroupValueWrite, but also for unicast messages such as MemoryRead. I need this to perform low(er)-level operations on my custom actors.

The main focus of this library was GroupValue payloads, so I had to refactor several parts of this library to make it more extensible. Fortunately, most of the higher-level API is still the same. In short:

- `Telegram.group_address` got replaced by `Telegram.address` (APCI supports both types)
- `Telegram.payload` should be a class that extends `knx.telegram.apci.APCI`. For instance, to write a DPT to a group address, you now use `Telegram(.., payload=GroupValueWrite(DPTArray(..)))`.
- `Telegram.telegramtype` is now encoded in `knx.telegram.apci.APCI.code`.

I've added two examples ([example_info.py](https://github.com/basilfx/xknx/blob/feature/apci_payload/examples/example_info.py) and [example_restart.py](https://github.com/basilfx/xknx/blob/feature/apci_payload/examples/example_restart.py)) to demonstate the new features.

Current limitation is that payloads can only be used in connectionless mode. This is not different from before, but several unicast messages are typically used with an active connection (that also supports sequencing and ACK/NACK on transport level). This would require more changes.

Current state is WIP and should not yet be merged. It's here to show and share my interest in this feature, and collect preliminary feedback. I'll continue on this one, but it needs a bit more time to fix the last part.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,8 @@
 
 - renamed "PhysicalAddress" to "IndividualAddress"
 - Telegram: `group_address` renamed to `destination_address`, to prepare support for other APCI services and add `source_address`
+- Telegram: remove `Telegram.telegramtype` and replace with payload object derived from `xknx.telegram.apci.APCI`.
+- CEMIFrame: remove `CEMIFrame.cmd`, which can be derived from `CEMIFrame.payload`.
 - Farewell Travis CI; Welcome Github Actions!
 
 ## 0.15.6 Bugfix for StateUpater 2020-11-26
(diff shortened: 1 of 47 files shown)


basilfx:
The [example_info.py](https://github.com/basilfx/xknx/blob/feature/apci_payload/examples/example_info.py) output looks like below:

image

basilfx:
I've rebased this PR, and split the first part of the commits into #510. If that one gets merged, I'll update this PR accordingly, and open another one for the next changes.

At this point, four unit tests still fail, but the rest seems to work OK.

PR #254, opened 2019-12-13
Update example xknx.yaml file - Climate

JensVanhooydonck:
While setting up a thermostat and using this as an example. The temperature would not load. 
The themperature address should not be under the mode option.


marvin-w:
Thanks!

PR #256, opened 2020-01-21
Corrected binding to multicast address in Windows

FredericMa:
I got the following error in Home Assistant while using the KNX component on Windows.
```
2020-01-21 09:08:42 ERROR (MainThread) [homeassistant.setup] Error during setup of component knx
  File "d:\homeassistant\lib\site-packages\homeassistant\setup.py", line 169, in _async_setup_component
    result = await component.async_setup(  # type: ignore
  File "d:\homeassistant\lib\site-packages\homeassistant\components\knx\__init__.py", line 101, in async_setup
    await hass.data[DATA_KNX].start()
  File "d:\homeassistant\lib\site-packages\homeassistant\components\knx\__init__.py", line 171, in start
    await self.xknx.start(
  File "d:\homeassistant\lib\site-packages\xknx\xknx.py", line 76, in start
    await self.knxip_interface.start()
  File "d:\homeassistant\lib\site-packages\xknx\io\knxip_interface.py", line 93, in start
    await self.start_routing(
  File "d:\homeassistant\lib\site-packages\xknx\io\knxip_interface.py", line 155, in start_routing
    await self.interface.start()
  File "d:\homeassistant\lib\site-packages\xknx\io\routing.py", line 66, in start
    await self.udpclient.connect()
  File "d:\homeassistant\lib\site-packages\xknx\io\udp_client.py", line 164, in connect
    sock = UDPClient.create_multicast_sock(self.local_addr[0], self.remote_addr, self.bind_to_multicast_addr)
  File "d:\homeassistant\lib\site-packages\xknx\io\udp_client.py", line 151, in create_multicast_sock
    sock.bind((remote_addr[0], remote_addr[1]))
OSError: [WinError 10049] The requested address is not valid in its context
```
The way of binding to the multicast address in Windows is different to the way of binding in Linux / Mac OS.

I fixed this with the this patch.

Note: this is my first pull request ever on Github so forgive me if there is something wrong with this PR :grinning:

### DIFF
--- a/xknx/io/udp_client.py
+++ b/xknx/io/udp_client.py
@@ -6,6 +6,7 @@ Due to lame support of UDP multicast within asyncio some special treatment for m
 """
 import asyncio
 import socket
+from sys import platform
 
 from xknx.exceptions import CouldNotParseKNXIP, XKNXException
 from xknx.knxip import KNXIPFrame
@@ -144,7 +145,10 @@ class UDPClient:
         # - bind() with own_ip does not work with ROUTING_INDICATIONS on Gira
         #   knx router - for an unknown reason.
         if bind_to_multicast_addr:
-            sock.bind((remote_addr[0], remote_addr[1]))
+            if platform == "win32":
+                sock.bind(('', remote_addr[1]))
+            else:
+                sock.bind((remote_addr[0], remote_addr[1]))
         else:
             sock.bind((own_ip, 0))
         sock.setsockopt(socket.IPPROTO_IP, socket.IP_MULTICAST_LOOP, 0)


farmio:
Nice, thank you!

What kind of IP-Router have you been testing with? According to the code comments some Gira devices seem to behave differently than others!? 
I can't really test this myself because I have no windows dev machine. 

We already have a OS-gate at https://github.com/XKNX/xknx/blob/3be9993a9abbc75d711a465d4e38150e89950d71/xknx/io/knxip_interface.py#L123 . I don't really like having 2 of these in different places when they are essentially doing the same. 
@Julius2342 do you know the specific reason for bind_to_multicast_addr being in ConnectionConfig() ? https://github.com/XKNX/xknx/blob/3be9993a9abbc75d711a465d4e38150e89950d71/xknx/io/knxip_interface.py#L68
I don't see any other usage than getting routing to work on different OSes. If this is true we could remove it and check for MacOS the same way @FredericMa has done in udp_client.py . Would be cleaner imho.

FredericMa:
> What kind of IP-Router have you been testing with? According to the code comments some Gira devices seem to behave differently than others!?
I can't really test this myself because I have no windows dev machine.

I'm using the MDT IP Router (First generation): https://www.mdt.de/EN_IP_Interface_Router.html

> We already have a OS-gate at
> 
> https://github.com/XKNX/xknx/blob/3be9993a9abbc75d711a465d4e38150e89950d71/xknx/io/knxip_interface.py#L123
> 

I didn't notice that check. So you would actually also move that check to the udp_client.py?
In that case I can make that change in this PR is you guys can confirm.

FredericMa:
I looked at the code and noticed that if we remove it in knxip_interface.py it looks like we can completely remove it from the project. So this means, also remote it from the ConnectionConfig class since binding to the multicast address or not will be determined by the connection_type parameter, right?

farmio:
I think so too. 
It seems that bind_to_multicast_addr was introduced to ConnectionConfig specifically for MacOS Routing.  https://github.com/XKNX/xknx/commit/017511c6ff93e5cbb92596739612a2deb47dbf18
I am not sure if it (still) has any other purpose in UDPClient.create_multicast_sock

FredericMa:
Any news on how to proceed with this one?

farmio:
I have just tested your Win32 `sock.bind(('', remote_addr[1]))` on macOS and funny enough I can  now receive ROUTING_INDICATION from Gira Router. This was not the case with `sock.bind((own_ip, 0))` and also `sock.bind((own_ip, remote_addr[1])). 
(`sock.bind((remote_addr[0], remote_addr[1]))` doesn't work at all on macOS - like Windows)

I will do some more testing - maybe we can even get rid of `bind_to_multicast_addr` and use your way for Win and Mac.

farmio:
We'll address the removal of bind_to_multicast_addr and the macOS issues in another PR.

PR #259, opened 2020-01-27
Change Cover open/close % to Knx default values

### LINKED ISSUE #257: is_open / is_closed not correctly reported if position is inverted
**Description of problem:**
The cover device directs the is_open and is_closed methods to the TravelCalculator. The problem is, that the TravelCalculator doesn't know about inverted position. So, is_open returns false if the position is 0. But with inverted position 0 is open.

I'm not sure where to fix it... should we change the TravelCalculator to correctly report open/closed? Or should we change the cover device which could check itself if current_position == open_position (which is correctly stored in the device).

Please give an advice and I could submit a PR

- [x] using xknx standalone

**Version information:**
- xknx / Home-Assistant release with the issue: 0.104.2

**KNX installation:**
Using a cover actuator from Jung.

- platform: knx
  name: bath
  move_long_address: '8/2/1'
  move_short_address: '8/1/1'
  position_address: '8/3/1'
  position_state_address: '8/4/1'
  invert_position: true

farmio:
- Change Cover and TravelCalculator to
  0% = open
  100% = closed
- Implemented `is_opening` and `is_closing` for Cover() and HA-integration
- remove inversion for 1bit telegrams
  1 bit telegrams have been inverted if `invert_position` was true.
  These are specified in knx as 0="up/open" and 1="down/close".
  I think invert_position shall only be applied to relative position.
- Removed TravelCalculator.time_set_from_outside in favor of unittest.mock.patch

fixes #257 

I've looked through the handbooks of the current blind actuators from ABB, Berker, MDT, Gira and Theben. They all see blinds at 0% as "open/up" and 100% as "closed/down".
HA has it the other way around. 
I propose to set invert_position and  invert_angle to true by default in the HA-integration.

### DIFF
--- a/xknx/devices/cover.py
+++ b/xknx/devices/cover.py
@@ -267,6 +267,14 @@ class Cover(Device):
         """Return if cover is closed."""
         return self.travelcalculator.is_closed()
 
+    def is_opening(self):
+        """Return if the cover is opening or not."""
+        return self.travelcalculator.is_opening()
+
+    def is_closing(self):
+        """Return if the cover is closing or not."""
+        return self.travelcalculator.is_closing()
+
     @property
     def supports_position(self):
         """Return if cover supports direct positioning."""
--- a/xknx/devices/cover.py
+++ b/xknx/devices/cover.py
@@ -159,16 +159,16 @@ class Cover(Device):
 
     async def set_short_down(self):
         """Move cover short down."""
-        await self.step.decrease()
+        await self.step.increase()
 
     async def set_short_up(self):
         """Move cover short up."""
-        await self.step.increase()
+        await self.step.decrease()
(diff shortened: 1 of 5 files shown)


xknx/devices/cover.py
```diff
@@ -161,11 +159,11 @@ def __str__(self):
...
         """Move cover short down."""
-        await self.step.increase()
+        await self.step.decrease()
 
     async def set_short_up(self):
         """Move cover short up."""
-        await self.step.decrease()
+        await self.step.increase()
 
     async def stop(self):
```
farmio: I'm not sure if `await self.step.increase()` shall be replaces with `await self.step.decrease()` for stopping.

PR #260, opened 2020-01-28
1.007 DPT_Step correction

recMartin:
1.007 DPT_Step   set: 0 = Decrease, 1 = Increase
Changed according to KNX Association System Specifications AS v1.5.00.

### DIFF
--- a/xknx/remote_value/remote_value_step.py
+++ b/xknx/remote_value/remote_value_step.py
@@ -40,19 +40,23 @@ class RemoteValueStep(RemoteValue):
         """Test if telegram payload may be parsed."""
         return isinstance(payload, DPTBinary)
 
+    # from KNX Association System Specifications AS v1.5.00:
+    # 1.007 DPT_Step   0 = Decrease 1 = Increase 
+    # 1.008 DPT_UpDown 0 = Up       1 = Down 
+
     def to_knx(self, value):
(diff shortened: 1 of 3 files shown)


farmio:
Good catch! Thanks!
I stumbled upon this issue just yesterday but didn't find this. I switched the call to increase() and decrease() in the Cover device 🤦‍♂️
Upside is that I have already fixed all your breaking tests in #259 . 
I think if i undo the switch in the Cover.set_short_up and Cover.set_short_down and merge both our PRs we would have it all covered.

xknx/remote_value/remote_value_step.py
```diff
@@ -40,19 +40,23 @@ def payload_valid(self, payload):
         """Test if telegram payload may be parsed."""
         return isinstance(payload, DPTBinary)
 
+    # from KNX Association System Specifications AS v1.5.00:
+    # 1.007 DPT_Step   0 = Decrease 1 = Increase
+    # 1.008 DPT_UpDown 0 = Up       1 = Down
```
farmio: 1.008 has its own RemoteValue found in remote_value_updown.py .
It would be nice if we combine them to a single file with these two classes - they are essentially the same and could easily inherit from a RemoteValueBinary base class or just from RemoteValueSwitch().

PR #255, opened 2020-01-15
Added DPTs based on KNX Specifications v2.1

eXtenZy:
For issue #236
Added all DPTs in the 5.*, 7.*, 8.*, 9.*, 13.* and 14.* categories based on the KNX Standard v2.1

PR #261, opened 2020-02-01
Derived DPTs and RemoteDateAndTime

recMartin:
Added more derived DPTs (float and int values) and incorporating into remote_value_sensor. Also done unittest for them.
Added RemoteValueDateAndTime which manages  'date': DPTDate, 'time': DPTTime, 'timeofday': DPTDateTime.
Defined 'payload_length' where missing and used it instead of constant.
Added remarks to input and output of to_knx() and from_knx() in xknx/dpt/__init__.py


farmio:
Bevor you continue please have a look at #255 !

farmio:
Thank you for your commits! We have merged #255 and #273 wich basically implement these features.

PR #264, opened 2020-02-05
Bugfix: finish the last task from the telegram queue

farmio:
When xknx is started and stopped multiple times sequentially
```python
await xknx.start()
await xknx.stop()
await asyncio.sleep(1)
await xknx.start()
await xknx.stop()
```
the second stop() call blocked infinitely https://github.com/XKNX/xknx/blob/ef18a0fa4795655e275721a08ffc5605d4095b02/xknx/xknx.py#L91
because of the `None` had no task_done() .


Julius2342:
That is a valid change :-)

PR #266, opened 2020-02-12
Fix the parsing of operation mode strings to HVACOperationMode

FredericMa:
I tried to use the operation_modes option in HomeAssistant and it appears that there is a bug in XKNX when it tries to parse the provided operation_modes to HVACOperationMode.
This PR fixes that.
The problem was that it tried to match the provided string with the keys of the HVACOpertaionMode enum instead of using it as the value to get the key.

### DIFF
--- a/test/devices_tests/climate_test.py
+++ b/test/devices_tests/climate_test.py
@@ -980,7 +980,7 @@ class TestClimate(unittest.TestCase):
 
     def test_custom_supported_operation_modes_as_str(self):
         """Test get_supported_operation_modes with custom mode as str list."""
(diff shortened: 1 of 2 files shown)


FredericMa:
The valid operation modes in the config of Home Assistant are:
`"Frost Protection", "Night", "Standby", "Comfort", "Auto", "Heat", "Cool", "Off", "Fan only", "Dry"`
It will also require a change in the home assistant KNX Climate integration since at the moment only `"Auto", "Heat", "Cool", "Off", "Fan only", "Dry"` are valid since the preset_modes are not included as valid options. I'll make PR in HA for this.

farmio:
LGTM! Thanks!

PR #267, opened 2020-02-18
Added config option force_hook to binary sensors

andreasnanko:
...to allow enforcing hook call regardless of the current binary sensor state.
If you need an on/off switch on one button, e.g. for a group of lights and want to enforce the state in any case by pressing the button.


Julius2342:
I find the name "force_hook" misleading, bc it does not clearly describe what it is doing. Is there any better name?

What about `ignore_internal_state` ?

It would even be better if we would find a positive name, e.g. `has_state` (If set to `none` the internal state is not saved and therefore not evaluated when changing the state.

andreasnanko:
I see. ignore_internal_state seems a good option, since the sensor can have a state, but it should be ignored under some circumstances. has_state would mislead that the binary sensor does not have a state, it has, but hooks does not fire since the state recieved equals the state it already have.

Julius2342:
@andreasnanko : Looks good. May I also ask you to add a unit test for this functionality?

(We just want to make sure we dont break this functionality some day in the future)

Thank you very much!

Julius2342:
Thank you :)

PR #273, opened 2020-03-29
introduce RemoteValueDateTime

### LINKED ISSUE #243: Unify device class Unify device class
DateTime should use a RemoteValue class for processing telegrams.

This issue is part of #240

farmio:
- use a RemoteValue in DateTime device
- use time.struct_time for internal time and date representation - this takes over some data validation
- enables time, date and datetime payloads that are not the current time (eg. for setting timers)

# Breaking changes:
- DateTime devices are initialized with sting for `broadcast_type`: "time", "date" or "datetime" instead of an Enum value.

closes #243

### DIFF
--- a/xknx/devices/climate.py
+++ b/xknx/devices/climate.py
@@ -48,7 +48,7 @@ class SetpointShiftValue(RemoteValue1Count):
             return None
         return super().value * self.setpoint_shift_step
 
-    async def set(self, value):
+    async def set(self, value, response=False):
         """Set new value from Kelvin."""
         if value > self.max_temp_delta:
             self.xknx.logger.warning("setpoint_shift_max exceeded at %s: %s",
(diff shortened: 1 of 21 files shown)


xknx/devices/climate.py
```diff
@@ -48,7 +48,7 @@ def value(self):
             return None
         return super().value * self.setpoint_shift_step
 
-    async def set(self, value):
+    async def set(self, value, response=False):
```
marvin-w, line 51: Why was this changed?

xknx/remote_value/remote_value_datetime.py
```diff
@@ -0,0 +1,58 @@
+"""
+Module for managing a remote date and time values.
```
marvin-w, line 2: Nice!

marvin-w:
Duplicated by #227.

PR #276, opened 2020-04-05
separate incoming and outgoing telegram queues

farmio:
- Incoming telegrams are processed immediately. Only outgoing telegrams have rate_limit not applied. To achieve this a second Queue is started (`xknx.telegram_queue.outgoing_queue`).

- `xknx.rate_limit` is checked for truthy so we can disable it (by setting it to False or 0).

- `process_all_telegrams` was renamed to `_process_all_telegrams` as it appears to only be used in unit tests. 

- `process_telegram` was removed. Exception handling is done in consumer tasks, logging in `process_telegram_*`

- `queue_stopped` asyncio.Event was removed in favor of just awaiting the consumer tasks.

### DIFF
--- a/test/core_tests/telegram_queue_test.py
+++ b/test/core_tests/telegram_queue_test.py
@@ -37,7 +37,7 @@ class TestTelegramQueue(unittest.TestCase):
 
         self.loop.run_until_complete(xknx.telegram_queue.start())
 
-        self.assertFalse(xknx.telegram_queue.queue_stopped.is_set())
+        self.assertFalse(xknx.telegram_queue._consumer_task.done())
         # queue shall now consume telegrams from xknx.telegrams
         self.assertEqual(xknx.telegrams.qsize(), 0)
         xknx.telegrams.put_nowait(telegram_in)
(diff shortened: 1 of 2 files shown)


farmio:
Note: I have not found a way to unit test the new rate limiting function yet. 

I'd like to put some incoming and outgoing telegrams to the queue and test for the incoming to be processed while the outgoing (except the first) are stuck in asyncio.sleep. But I don't want to actually sleep in the tests so they aren't slowed down for this. Any ideas?

iligiddi:
> Note: I have not found a way to unit test the new rate limiting function yet.
> 
> I'd like to put some incoming and outgoing telegrams to the queue and test for the incoming to be processed while the outgoing (except the first) are stuck in asyncio.sleep. But I don't want to actually sleep in the tests so they aren't slowed down for this. Any ideas?

you may replace the sleep with awaiting an Event with Timeout. So in the unit-test you may just trigger the event to speed up the process.

Julius2342:
Or just do: 

`with mock.patch('asyncio.sleep', new_callable=AsyncMock):`

PR #277, opened 2020-04-10
Automatically publish packages to pypi

Julius2342:


.github/workflows/pythonpublish.yml
```diff
@@ -0,0 +1,31 @@
+# This workflows will upload a Python Package using Twine when a release is created
+# For more information see: https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#publishing-to-package-registries
+
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
```
farmio, line 8: I think we should use `[published]` here: https://github.community/t5/GitHub-Actions/What-is-the-created-type-in-on-release/td-p/41171

farmio:
The tag name, in our case the version number, is stored in the environment variable `GITHUB_REF`. Maybe we could use this to inject it to setup.py?

.github/workflows/pythonpublish.yml
Julius2342: yep, you are right.

PR #286, opened 2020-05-23
Register group

iligiddi:
Allow configuration of custom devices via the main yaml.
Also config.py is fully independent from any specific device implementation now.

Usage example:
        xknx = XKNX(
                loop=self.loop,
                custom_device_classes = { "custom_device" : CustomDevice },
                config= "xknx.yaml")


iligiddi:
need to fix Travis CI reports first

PR #292, opened 2020-05-23
Register group

iligiddi:
Enable registration of customized Device classes.

Customized device classes can also be configured within the same yaml configuration.
Usage example:
xknx = XKNX(
loop=self.loop,
custom_device_classes={"custom_device": CustomDevice},
config="xknx.yaml")
)

config.py does not contain device dependencies any more.

### DIFF
--- a/test/core_tests/config_test.py
+++ b/test/core_tests/config_test.py
@@ -463,7 +463,7 @@ class TestConfig(unittest.TestCase):
     def test_config_file_error(self):
         """Test error message when reading an errornous config file."""
         with patch('logging.Logger.error') as mock_err, \
-                patch('xknx.core.Config.parse_group_light') as mock_parse:
+                patch('xknx.devices.Light.from_config') as mock_parse:
             mock_parse.side_effect = XKNXException()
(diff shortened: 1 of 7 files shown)


iligiddi:
Enable registration of customized Device classes.

Customized device classes can also be configured within the same yaml configuration.
Usage example:
        xknx = XKNX(
                loop=self.loop,
                custom_device_classes={"custom_device": CustomDevice},
                config="xknx.yaml")
            )

config.py does not contain device dependencies any more.

iligiddi:
Enable registration of customized Device classes.

Customized device classes can also be configured within the same yaml configuration.
Usage example:
xknx = XKNX(
loop=self.loop,
custom_device_classes={"custom_device": CustomDevice},
config="xknx.yaml")
)

config.py does not contain device dependencies any more.

marvin-w:
@iligiddi We will no longer support the old config style. Your refactoring makes kind of sense but due to the fact that you've mixed new functionality with a major refactoring this change is really hard to review and stale since march.

I'll close this therefore. If you want to add `CustomDevices` please open a PR for the dedicated device type.

PR #302, opened 2020-06-06
Harmonized HASS spelling and tried to make docu a bit clearer

spacegaier:
Tried to harmonize the spelling a bit (i.e. HASS vs hass), plus make the context of the page clearer.

### DIFF
--- a/docs/home_assistant.md
+++ b/docs/home_assistant.md
@@ -3,10 +3,10 @@ layout: default
 ---
 
 
-Home Assistant Component
+Home Assistant KNX Integration
 ========================
 
-XKNX is shipped within [Home Assistant](https://www.home-assistant.io/), the great platform for home automation, in the form of the included [KNX component](https://www.home-assistant.io/integrations/#search/KNX).
+XKNX is shipped within [Home Assistant (HA)](https://www.home-assistant.io/), the great solution for home automation, in the form of the included [KNX integration](https://www.home-assistant.io/integrations/#search/KNX).
 
 For development and debugging reasons - or to catch up with the newest features - you may still want to use the custom component version of the plugin.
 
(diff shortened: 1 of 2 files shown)


docs/home_assistant.md
```diff
@@ -54,7 +54,7 @@ Starting via service is also possible, but you have to change the configuration
 Configuration:
 --------------
 
-The configuration works as described within [Home Assistant documentation](https://home-assistant.io/components/#search/knx) with the difference that the component is called `xknx` instead of `knx`.
+The configuration for the manually checked out version works the same as described within [Home Assistant KNX documentation](https://home-assistant.io/components/#search/knx) with the difference that the component is called `xknx` instead of `knx` (which is be the HASS default KNX integration name).
```
farmio, line 57: - "with the difference that the component is called `xknx`" should we use "platform" instead of component? (Component is now light or switch - I think the wording changed some time ago)
- "which is be" doesn't sound right. I think this is a typo.

docs/home_assistant.md
```diff
@@ -6,15 +6,15 @@ layout: default
 Home Assistant Component
 ========================
 
-XKNX ships with [Home Assistant](https://home-assistant.io/components/#search/knx), the great platform for home automation!
+XKNX is shipped within [Home Assistant](https://www.home-assistant.io/), the great platform for home automation, in the form of the included [KNX component](https://www.home-assistant.io/integrations/#search/KNX).
 
-For development and debugging reasons - or to catch up with the newest features - you may still use the custom component version of the plugin.
+For development and debugging reasons - or to catch up with the newest features - you may still want to use the custom component version of the plugin.
```
farmio, line 11: Maybe we shouldn't call it plug-in as it could be easily confused with "add-on" which is a different Home-Assistant concept. 
I'd say integration or platform.

docs/home_assistant.md
spacegaier: * The wording in general in HA is quite confusing to me, which is why I triggered a new discussion [here](https://github.com/home-assistant/developers.home-assistant/issues/570) to get more clarity.
* Yes, midnight typo.

spacegaier:
I agree with your comments (esp. regarding German :grin:). I like the `Home Assistant (HA)` approach. I don't think we necessarily need "Core" in there.

PR #303, opened 2020-06-09
BinarySensor: Remote value binary

### LINKED ISSUE #241: Unify device class BinarySensor
BinarySensor should use RemoteValueSwitch class for processing telegrams.

This issue is part of #240

### LINKED ISSUE #298: reset_after for binary_sensor blocks bus writing
**Description of problem:**

- [ ] using xknx standalone

I have an automation to bypass any opened window in the house in order to arm our alarm system if any are open.

Sequence is like this:
- We press a  button on a KNX actuator (see yaml below) that will trigger an automation.
- The automation sends command to the alarm system to bypass every open window via MQTT.
- If all is OK, the alarm sends a "ready to arm" message back via MQTT.
- This triggers another automation that switches a KNX switch to toggle a LED on the KNX actuator.

Although the last step in sequence above is triggered aprox. 3 seconds later, the LED only toggles 10 seconds after the first step. This is exactly the configured "reset_after" delay that is configured for the binary_sensor used in the first step of the sequence describe above.

I supose this line is causing the blocking:
https://github.com/XKNX/xknx/blob/3ab0a53e334e2c55949b57a3e6704466c87e7cf7/xknx/devices/binary_sensor.py#L159

**Version information:**
- xknx / Home-Assistant release with the issue: 0.110.4

**KNX installation:**

```yaml
  - platform: knx
    name: alarm_bypass_windows
    address: '6/3/2'
    reset_after: 10000 # Reset after 10 seconds to ignore double presses.
    sync_state: false
```

farmio:
- use a RemoteValue in BinarySensor device
- use a regular Bool type for state (this also avoids imports outside toplevel)
- reset_after is now implemented as asyncio.Task to prevent blocking the loop
- RemoteValue.process has `always_callback` attribute to run the callbacks on every process even if the payload didn't change

# Breaking changes:

- significant_bit attribute is removed

closes #241 
fixes #298

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,237 +1,194 @@
-Changelog
-=========
+# Changelog
 
-Upcoming version (unreleased)
------------------------------
-
-### New Features
-
-* Added config option ignore_internal_state in binary sensors (@andreasnanko #267)
-
-### Internals
-
-* Automatically publish packages to pypi (@Julius2342 #277)
-* keep xknx version in `xknx/__version__.py` (@farmio #278)
-* add raw_socket logger (@farmio #299)
-
-
-0.11.3 Sensor types galore!  2020-04-28
----------------------------------------
-
-### New Features
-
-* added a lot of DPTs now useable as sensor type (@eXtenZy #255) 
-
-### Bugfixes
-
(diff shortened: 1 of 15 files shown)


farmio:
integrated in #227

PR #304, opened 2020-06-10
[WIP] replace asyncio with anyio

smurfix:
Trio [https://github.com/python-trio/trio] is an alternate async library for Python that follows Structured Concurrency paradigms.
AnyIO [https://github.com/agronholm/anyio] is a compatibility library that allows programs or libraries which use it to work with asyncio or trio, without modification.

This patchset converts the whole of XKNX to use anyio, making xknx useable by programs written for Trio.

Notable side effects:

This PR replaces unittest with pytest (Trio and AnyIO basically require it)

XKNX now supports context managers plus async loops instead of callback registration.

This PR is WIP because it depends on an anyio patch that has not been merged (let alone released) yet.


smurfix:
AnyIO 1.4 is released and works.

I'd like for somebody to test the HASS integration with this branch because, well, I don't use it …

farmio:
Please add anyio (and other requirements if there are any) to `requirements/production.txt`.

xknx/io/udp_client.py
```diff
@@ -111,16 +77,29 @@ def register_callback(self, callback, service_types=None):
...
         self.callbacks.remove(callb)
 
     @staticmethod
-    def create_multicast_sock(own_ip, remote_addr, bind_to_multicast_addr):
+    async def create_multicast_sock(own_ip, remote_addr):
         """Create UDP multicast socket."""
-        sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        sock.setblocking(False)
+        sock = await anyio.create_udp_socket(port=remote_addr[1], address_family=socket.AF_INET)
```
farmio, line 102:
```suggestion
        sock = await anyio.create_udp_socket(port=remote_addr[1], family=socket.AF_INET)
```
`address_family` gets me an unexpected keyword argument error.

Even with this and `reuse_address=True` I can't get this to work at all on macOS. I think anyio is missing an option to set
`sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)`
before bind().

> INFO:xknx.log:XKNX v0.11.3 starting routing connection to KNX bus.
DEBUG:xknx.log:Searching on lo0 / 127.0.0.1
DEBUG:xknx.log:Searching on en1 / 10.1.1.9
ERROR:xknx.log:Could not connect to an KNX/IP device on en1
  File "/Users/meti/Documents/Dev/xknx-xknx/xknx/xknx/io/gateway_scanner.py", line 133, in _search_interface
    await udp_client.connect()
  File "/Users/meti/Documents/Dev/xknx-xknx/xknx/xknx/io/udp_client.py", line 123, in connect
    self.socket = await self.create_multicast_sock(self.local_addr[0], self.remote_addr)
  File "/Users/meti/Documents/Dev/xknx-xknx/xknx/xknx/io/udp_client.py", line 102, in create_multicast_sock
    sock = await anyio.create_udp_socket(port=remote_addr[1], family=socket.AF_INET)
  File "/Users/meti/Documents/Dev/xknx-xknx/xknx/.tox/py37/lib/python3.7/site-packages/anyio/__init__.py", line 555, in create_udp_socket
    await sock.bind((interface or '', port or 0))
  File "/Users/meti/Documents/Dev/xknx-xknx/xknx/.tox/py37/lib/python3.7/site-packages/anyio/_networking.py", line 72, in bind
    await self._run_in_thread(self._raw_socket.bind, address)
  File "/Users/meti/Documents/Dev/xknx-xknx/xknx/.tox/py37/lib/python3.7/site-packages/anyio/_backends/_asyncio.py", line 483, in run_in_thread
    raise exception
  File "/Users/meti/Documents/Dev/xknx-xknx/xknx/.tox/py37/lib/python3.7/site-packages/anyio/_backends/_asyncio.py", line 455, in thread_worker
    result = func(*args)
OSError: [Errno 48] Address already in use

agronholm: The `address_family` was part of @smurfix 's PR but was not in the actual released code.
As for `SO_REUSEPORT`, that was not even part of the PR. I've already added it in my socket/stream refactoring effort for AnyIO v2 though.

You are aware that `SO_REUSEPORT` does not work on Windows?

farmio: Great😃! 

As far as I'm concerned `SO_REUSEPORT` is only required on macOS, but I have no Windows or Linux testing environment (but xknx seems to run on these platforms without problems currently - without `SO_REUSEPORT`).

So to not break xknx compatibility to macOS we would have to wait until anyio v2 is released before merging this?

smurfix: Meh, I seem to have that snippet from my previous PR in my local repository and didn't notice. :-/

I'll update this patch series as soon as AnyIO v2 comes out, let's wait until then.

agronholm: > I'll update this patch series as soon as AnyIO v2 comes out, let's wait until then.

So is this because `SO_REUSEPORT` is missing? Why is it specifically needed on macOS but not other platforms?
As you can see here, the PR did not add that option: https://github.com/agronholm/anyio/pull/115/files

smurfix: > So is this because `SO_REUSEPORT` is missing? Why is it specifically needed on macOS but not other platforms?

Because UDP networking is a grab bag of minor incompatibilities, doubly so when multicast is concerned?

You can use SO_REUSEPORT instead of SO_REUSEADDR on Linux AFAICT, but not on Windows. If I had known that macOS requires that instead of REUSEADDR I'd've sent a more complicated patch …

farmio: I'm not sure macOS needs it instead of REUSEADDR. We are currently running SO_REUSEPORT additionally to SO_REUSEADDR for macOS 🙃

agronholm: Both asyncio and trio have stern warnings against trying to use `SO_REUSEADDR` on Windows in any capacity, so the plan for AnyIO v2 is to not do that either. I'm going to make `SO_REUSEPORT` optional but `SO_REUSEADDR` always enabled on non-Windows platforms (which is what asyncio does too). Is this okay?

farmio: My only knowledge for multicast on windows is from this PR: https://github.com/XKNX/xknx/pull/256/files
Here it seems to work with SO_REUSEADDR but you can't bind the multicast group, but INADDR_ANY. 
Same is true for macOS but it needs SO_REUSEPORT too.

Since it is optional we can always check for platform in our code and pass required arguments to anyio, I guess. Would be nice if anyio would do that for us though ;)

Julius2342:
@smurfix : Am i right, we should keep this branch and MR here and try to keep it up to date?

If we do so, ideally we try to uplift other changes, to keep the diff as small as possible, e.g. the async context manager:

`async with XKNX().run() as xknx:`

marvin-w:
We already have an async context manager

`async with XKNX() as xknx:`

PR #305, opened 2020-06-11
ClimateMode: use RemoteValue subclasses

### LINKED ISSUE #242: Unify device class ClimateMode
ClimateMode should use RemoteValue classes for processing telegrams. Maybe one for HVACMode, ControllerMode and ControllerStatus types.

This issue is part of #240

### LINKED ISSUE #308: Service calls fail (climate.set_hvac_mode & climate.set_preset_mode)
**Description of problem:**

Runtime error occurs when attempting to call the service `climate.set_hvac_mode` from the HA developer tools.

image

The service calls goes through here:
https://github.com/XKNX/xknx/blob/c22e394ca1c7ade54b2cffe8bcedfd782f47bc75/home-assistant-plugin/custom_components/xknx/climate.py#L314

The problem is that `set_operation_mode(self, operation_mode)` is used to update both the hvac operation mode (xknx variable `controller_status_*`) as well as the preset mode (`operation_mode_*`).  

https://github.com/XKNX/xknx/blob/c22e394ca1c7ade54b2cffe8bcedfd782f47bc75/xknx/devices/climate_mode.py#L143

But that is incorrect, because now the method attempts to use my HVAC mode "cool" as an preset, which of course fails.

=> I think the cleanest solution would be to split  `set_operation_mode(self, operation_mode)` into two dedicated methods, one for HVAC mode and one for the presets. 

**BTW**: Call of service `climate.set_preset_mode` works, but only because the corresponding `controller_status_*` parameters are commented out. Otherwise I am quite certain it would fail when attempting to update those values. 

- [ ] using xknx standalone
- [X] using Home-Assistant knx plugin

**Version information:**
- xknx / Home-Assistant release with the issue: 0.11.3 / 0.112.0.dev0

**Problem-relevant `xknx.yaml` or `configuration.yaml` entries:**
```YAML
climate:
  - platform: knx
    name: KNX Office
    temperature_address: '3/1/4'
    target_temperature_address: '3/2/4'
    target_temperature_state_address: '3/6/4'
    operation_mode_address: '3/4/2'
    operation_mode_state_address: '3/4/3'
    # controller_status_address: '3/4/0'
    # controller_status_state_address: '3/4/1'
    min_temp: 7.0
    max_temp: 32.0
    operation_modes:
      - "Off"
      - "Heat"
      - "Cool"
      - "Comfort"
      - "Standby"
      - "Night"
      - "Frost Protection"
```

```Traceback (most recent call last):
  File "/home/pi/git/core/homeassistant/core.py", line 1277, in catch_exceptions
    await coro_or_task
  File "/home/pi/git/core/homeassistant/core.py", line 1296, in _execute_service
    await handler.func(service_call)
  File "/home/pi/git/core/homeassistant/helpers/entity_component.py", line 209, in handle_service
    self._platforms.values(), func, call, required_features
  File "/home/pi/git/core/homeassistant/helpers/service.py", line 454, in entity_service_call
    future.result()  # pop exception if have
  File "/home/pi/git/core/homeassistant/helpers/entity.py", line 605, in async_request_call
    await coro
  File "/home/pi/git/core/homeassistant/helpers/service.py", line 485, in _handle_entity_call
    await result
  File "/home/pi/git/core/homeassistant/components/knx/climate.py", line 314, in async_set_hvac_mode
    await self.device.mode.set_operation_mode(knx_operation_mode)
  File "/home/pi/git/core/venv/lib/python3.7/site-packages/xknx/devices/climate_mode.py", line 150, in set_operation_mode
    DPTArray(DPTHVACMode.to_knx(operation_mode)))
  File "/home/pi/git/core/venv/lib/python3.7/site-packages/xknx/dpt/dpt_hvac_mode.py", line 69, in to_knx
    raise ConversionError("Could not parse HVACOperationMode", value=value)
xknx.exceptions.exception.ConversionError: <ConversionError description="Could not parse HVACOperationMode" value="HVACOperationMode.COOL"/>`

### PR DESCRIPTION
**Pull request by farmio**
- use RemoteValueClimateMode and RemoteValueClimateBinaryMode instead of directly sending from Device
- binary climate modes should be fully functional now (sending, receiving and syncing) - but only on one address
- add `group_address_operation_mode_standby` as binary operation mode

closes #242 
resolves #308

### CODE REVIEW

File: xknx/devices/climate_mode.py [hunk 1]
```diff
@@ -142,35 +165,20 @@ async def _set_internal_operation_mode(self, operation_mode):
...
+
+        for rv in self.__iter_remote_values_climate_mode():
+            if rv.writable and \
+                    operation_mode in rv.supported_operation_modes():
+                await rv.set(operation_mode)
+
+        for rv in self.__iter_remote_values_climate_binary_mode():
+            if rv.writable:
+                # foreign operation modes will set the RemoteValue to False
+                await rv.set(operation_mode)
```
farmio, line 180: Could this lead to problems where splitted up operation_modes and eg. controller_mode is defined? Eg. controller_mode is set to "cool" should operation_mode_comfort be set to `False`? Is this constellation even a thing?
If it was we would need to check against the other 2 possible binary operation modes. This would mean to override `RemoteValue.set()` or check there against `None` from `RemoteValue.to_knx()`.

xknx/devices/climate_mode.py
```diff
@@ -180,103 +188,49 @@ def operation_modes(self):
...
-        if self.group_address_controller_status is not None:
-            return [HVACOperationMode.COMFORT, HVACOperationMode.STANDBY,
-                    HVACOperationMode.NIGHT, HVACOperationMode.FROST_PROTECTION]
-        if self.group_address_controller_mode is not None:
-            return [HVACOperationMode.STANDBY, HVACOperationMode.AUTO, HVACOperationMode.HEAT,
-                    HVACOperationMode.COOL, HVACOperationMode.FAN_ONLY, HVACOperationMode.DRY]
-
-        # Operation modes only supported partially
+    def gather_operation_modes(self):
+        """Gather operation modes from RemoteValues."""
```
farmio, line 192: we aren't guessing 😉

xknx/devices/climate_mode.py
```diff
@@ -180,103 +188,49 @@ def operation_modes(self):
...
-        if self.group_address_operation_mode_night:
-            operation_modes.append(HVACOperationMode.NIGHT)
-        if self.group_address_operation_mode_protection:
-            operation_modes.append(HVACOperationMode.FROST_PROTECTION)
-        return operation_modes
+        for rv in self.__iter_remote_values():
+            if rv.writable:
+                operation_modes.extend(rv.supported_operation_modes())
+        # remove duplicates
+        return list(set(operation_modes))
```
farmio, line 198: Don't really like this, but I haven't found a more elegant way.

xknx/devices/climate_mode.py
spacegaier: I am not sure I get your point here. Controller and operation mode should be set independently. I can be in "cool" and still need to control if I want "comfort", "away", etc. Or am I missing your point here?

xknx/dpt/dpt_hvac_mode.py
```diff
@@ -30,46 +30,70 @@ class HVACOperationMode(Enum):
...
+
+class DPTHVACMode(_DPTClimateMode):
+    """
+    Abstraction for KNX HVAC mode.
+
+    DPT 20.102
+    """
+
+    SUPPORTED_MODES = {
+        0: HVACOperationMode.AUTO,
```
spacegaier, line 87: Wouldn't it make sense to use the official KNX terms here?

Defintion from KNX Standard v2.1:
image

xknx/devices/climate_mode.py
spacegaier: This should also work and would keep the order:
`operation_modes = list(dict.fromkeys(operation_modes))`

xknx/devices/climate_mode.py
farmio: Is it save to assume when eg. only group_address_operation_mode_night is set the supported operation modes are: Night, Standby, Comfort, Frost Protection? Even if we can not set 2 of them?

xknx/dpt/dpt_hvac_mode.py
farmio: good point. I just copied what was already there. I think we should totally use the knx values and just map it to fitting HA values in the HA integration

spacegaier: How about we add a new config option *_eco and mark the existing one as deprecated but keep it around for some time? And also add the option for "standby" which I think is currently missing.

xknx/devices/climate_mode.py
spacegaier: I don't think we can know/say either way for sure. My understanding is, that a device either supports the 20.102 approach where you have one address that can handle all modes or your device splits them up into multiple 1 bit addresses, but it should then still have one address for each mode so it can still offer the same options. My Gira room controller e.g. offers both options, so they are just different ways of handling (e.g. you choose what can be better integrated into your visualization concept).
But on the other hand, there could potentially be devices that only support two modes (not sure if they would be 100% KNX conform then TBH). And even if a device supports all modes, a xknx user might still only provide the 1 bit addresses for a few of them, if they only want to use those few e.g. via HA.

Not sure if that comment helps or not 🤷 ...

xknx/dpt/dpt_hvac_mode.py
farmio: Looking at the MDT actuator "Standby" is assumed when the other 3 are set to 0. There is no communication object for "Standby". Are you aware if this the case for other models / are there even other manufacturers using binary mode objects?

spacegaier: Yes, my Gira room controller has four separate 1bit channels. See page 54 of the manual, functions 82-85: https://download.gira.de/data3/51423490.pdf

farmio: Right, I just had a look at my Jung actuator 2336. It also has a standby CO. Not really sure how to implement this without a new configuration mode. Maybe we could just add a note in the documentation to manually set `operation_modes` when using the MDT type.

Btw. Jung, Gira and also MDT refer to eco mode as "Night" in their (german) manuals.

xknx/devices/climate_mode.py
farmio: Anyway even with 1byte operation mode and controller mode it is currently not possible to set "cool" and "comfort" at the same time because a new value is always stored in `self.operation_mode`. Probably the right way would be to return a list of currently active modes or store the values separately. Then the HA integration can parse it as mode or preset - which don't really fit knx modes.

farmio:
integrated in #227

PR #311, opened 2020-07-16
Timeout-only gateway scanning

jochembroekhoff:
This adds an additional scanning mode to the GatewayScanner, namely timeout-only scanning.
In some cases it is not desirable to limit the amount of gateways that can be scanned.

Passing `None` or an integer less than or equal to zero to the `stop_on_found` parameter now causes the scanner to only stop once the timeout is reached.

### DIFF
--- a/xknx/io/gateway_scanner.py
+++ b/xknx/io/gateway_scanner.py
@@ -94,7 +94,7 @@ class GatewayScanner():
         self.timeout_in_seconds = timeout_in_seconds
         self.stop_on_found = stop_on_found
         self.scan_filter = scan_filter
-        self.found_gateways = []  # List[GatewayDescriptor]
+        self.found_gateways = []  # type: List[GatewayDescriptor]
         self._udp_clients = []
         self._response_received_or_timeout = asyncio.Event()
         self._timeout_handle = None
--- a/xknx/io/gateway_scanner.py
+++ b/xknx/io/gateway_scanner.py
@@ -7,7 +7,7 @@ GatewayScanner is an abstraction for searching for KNX/IP devices on the local n
 """
 
 import asyncio
-from typing import List
+from typing import List, Optional
 
 import netifaces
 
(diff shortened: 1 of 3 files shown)


xknx/io/gateway_scanner.py
```diff
@@ -172,6 +172,8 @@ def _response_rec_callback(self, knx_ip_frame: KNXIPFrame, udp_client: UDPClient
     def _add_found_gateway(self, gateway):
         if self.scan_filter.match(gateway):
             self.found_gateways.append(gateway)
+            if self.stop_on_found is None or self.stop_on_found <= 0:
```
Julius2342, line 175: Isnt the `self.stop_on_found <= 0` a parameter validation which should not happen within this function? (I would place it within the initializer and throw an exception if the value is negative).

It also took my a while to understand the logic and would suggest to combine this into one condition? What do you think?

jochembroekhoff: Indeed. That would certainly be better (the existing implementation also never checked the actual value).
However, since this is a public field, it can still legally be changed after the instance has been constructed, so the validation cannot only be done in the constructor.
Even during scanning the value can theoretically be changed (although that is probably not very likely).
What is the general attitude to thinks like this in the project? I wouldn't immediately say it's the user's problem.

Assuming the value does not change after the `scan` method is called, a different way of solving this problem may be just keeping a protected field with a 'clean' upper limit value. Let's say that clean value is stored in `self._count_upper_bound` and is always a valid non-negative integer, then 0 indicates no limit and the rest indicates that there is one.
Then the condition I added to `_add_found_gateway` can be moved to the `scan` method, only to clean the current value, and the condition that compares the length can be changed to `if 0 < self._count_upper_bound <= len(self.found_gateways):`.

To be honest, this still does not feel like the best solution. More refactoring may be necessary.
Alternatively, changing the `stop_on_found` field to be protected would solve this problem immediately (and I don't think API compatibility is an issue yet).

farmio (approved):
Great! 

I would never have thought of stop_on_found being changed after instantiation - thats probably because I've never been a CSE student 🤣

This will come handy when implementing HA config_flow. I'm thinking of optionally presenting the user a list of available gateways to choose from instead of configuring by typing the ip-address. 😃

PR #312, opened 2020-07-17
Multicast address configuration

jochembroekhoff:
This adds support for configuring the previously hard-coded multicast address.

Tested with knxd 0.14.39.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -15,6 +15,8 @@ The configuration file can contain three sections.
 - The `general` section can contain:
   - `own_address` the individual / physical address of the XKNX daemon
   - `rate_limit` a rate limit for telegrams sent to the bus
+  - `multicast_group` to override the default multicast address (`224.0.23.12`)
+  - `multicast_port` to override the default multicast port (`3671`)
 - The `connection` section can be used to specify the connection to the KNX interface.
   - `auto` for automatic discovery of a KNX interface
   - `tunneling` for a UDP unicast connection
--- a/docs/xknx.md
+++ b/docs/xknx.md
@@ -23,7 +23,9 @@ xknx = XKNX(config='xknx.yaml',
             address_format=GroupAddressType.LONG
             telegram_received_cb=None,
             device_updated_cb=None,
-            rate_limit=DEFAULT_RATE_LIMIT)
+            rate_limit=DEFAULT_RATE_LIMIT,
+            multicast_group=DEFAULT_MCAST_GRP,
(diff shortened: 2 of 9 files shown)


farmio:
Wouldn't this be a good fit in the `connection` section of configuration instead of `general`? The attributes would fit nicely in ConnectionConfig() instead of Xknx().

jochembroekhoff:
Good point. I did think of that, but can't really say why I didn't implement it that way.
Maybe it was because the connection config is `None` by default, which is not very helpful when, for example, only constructing the `XKNX` object in order to perform a scan while not starting XKNX itself.
But I think it shouldn't really be a problem to slightly change that.
I would certainly say it's better to place it in the connection config object.

In the case of an explicit routing connection, it would certainly suffice to only have to configure the IP/port once. It wouldn't make sense to perform a scan with a different address used for discovery that would never be able to actually discover that routing interface.
For tunneling however, it may not be that straightforward. In most cases where tunneling is desired, either the automatic mode discovers a tunneling server, or it is explicitly configured.
However, in the latter case it will not be possible anymore to use the GatewayScanner, since that would then start searching on the IP address of the tunneling device, and _not_ on the (almost certainly expected) default multicast address.
One could of course say that this situation is rather uncommon, and that the solution to that is to simply have separate XKNX instances.

Personally I would still leave them separated, as this is probably a feature that's only going to be used by uses who know what they're doing (it's a kind of nonstandard thing, many KNX IP interfaces don't even support using a different multicast address configuration; even ETS appears to have the address hard coded).

farmio:
ConnectionConfig can come from XKNX object, from configuration in XKNX.__init__ or from HomeAssistant passed to XKNX.start(). It is None per default now to check if one of these was manually set later in start(). I think this is quite confusing. If HA would pass its ConnectionConfig when initializing Xknx() it would maybe be clearer. Then a default ConnectionConfig can be initialized directly in Xknx.__init__() and used in gateway scans
  
I don't see a need for a gateway scan if a dedicated tunnelling device is set - but I surely have no good overview of all possible applications for xknx. I agree that it is more straight forward to have separate attributes for it. Otherwise the GatewayScanner could check if the `remote_ip` is part of the multicast address range 224.0.0.0/4 and fall back to default if it is not.
(In ETS you can find this in the properties of the Backbone in Topology tab if you want to test your changes 😉)

farmio (approved):
I've resolved a merge conflict by removing `bind_to_multicast` attribute in xknx/io/routing.py as it was removed in 0.12.0

PR #315, opened 2020-07-25
Update testing requirements and CI

farmio:
Coveralls should work again. Let's see...

### DIFF
--- a/xknx/devices/action.py
+++ b/xknx/devices/action.py
@@ -24,6 +24,7 @@ class ActionBase():
 
     def test_if_applicable(self, state, counter=None):
         """Test if should be executed for this state and this counter number."""
+        # pylint: disable=import-outside-toplevel
         from .binary_sensor import BinarySensorState
         if (state == BinarySensorState.ON) \
                 and (self.hook == "on") \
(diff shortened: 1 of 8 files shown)


farmio:
Well seems like coveralls works again, but doesn't comment the coverage difference anymore. @Julius2342 maybe some settings or permissions have changed there?

xknx/devices/action.py
```diff
@@ -24,6 +24,7 @@ def test_counter(self, counter):
 
     def test_if_applicable(self, state, counter=None):
         """Test if should be executed for this state and this counter number."""
+        # pylint: disable=import-outside-toplevel
```
Julius2342, line 27: Hm, i don't remember the reason why i made this a local import ...

Julius2342:
> Well seems like coveralls works again, but doesn't comment the coverage difference anymore. @Julius2342 maybe some settings or permissions have changed there?

Hmm, i did not change anything ... we could also consider moving to github tests? That probably could make our lives easier?

farmio:
Sure, why not. 
This changes are relevant there too because of a version conflict of coveralls and coverage > 5

xknx/devices/action.py
farmio: It's because of a circular dependency conflict. Im not really sure how to prevent this. New Stateupdater eliminates some of these.

PR #316, opened 2020-07-30
rename DPTBase to DPTTranscoder; update HA knx.send

farmio:
- rename DPTBase to DPTTranscoder
- introduce _DPTPayload as base class for DPTBinary and DPTArray. For future type checking.

I think DPTTranscoder is more self-explanatory than DPTBase.

----------

I was also thinking of changing the DPTTranscoder.to_knx() return values to instances of _DPTPayload. 
Eg. DPTTemperature.to_knx(value) now returns a tuple of `(exponent, significand)`. This is packed into a DPTArray in the RemoteValue using it. I think it would be more straight forward to return a DPTArray from this function. I'm not sure how to do this with RemoteValues that use DPTBinary - maybe just add a .to_knx() and .from_knx() function to it.

I'll add an update to the HA service `knx.send` to soon to take an additional attribute `type`. It'll use `DPTTranscoder.parse_type` to select a DPT and encode the payload accordingly (like RemoteValueSensor). So we don't have to use tuples with raw payloads anymore. This would benefit from DPTTranscoder.to_knx() returning a _DPTPayload.

What do you think??

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,8 @@
 
 ### New Features
 
+- HA integration: `knx.send` service takes `type` attribute to allow sending DPT encoded values like `sensor`
+- HA integration: `sensor` and `expose` accept int and float values for `type` (parsed as DPT numbers)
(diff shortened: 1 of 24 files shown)


farmio:
i messed up my next branch with this 🤦‍♂️

PR #318, opened 2020-08-01
rename DPTBase to DPTTranscoder

farmio:
- rename DPTBase to DPTTranscoder
- introduce _DPTPayload as base class for DPTBinary and DPTArray. For future type checking.

I think DPTTranscoder is more self-explanatory than DPTBase.

----------

I was also thinking of changing the DPTTranscoder.to_knx() return values to instances of _DPTPayload. 
Eg. DPTTemperature.to_knx(value) now returns a tuple of `(exponent, significand)`. This is packed into a DPTArray in the RemoteValue using it. I think it would be more straight forward to return a DPTArray from this function. I'm not sure how to do this with RemoteValues that use DPTBinary - maybe just add a .to_knx() and .from_knx() function to it.

What do you think??

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -27,7 +27,7 @@
 
 ### Internals
 
-- Use new DPTTranscoder class as base for DPT implementations - use DPTBase as base class for DPTArray and DPTBinary
+- Use new DPTTranscoder class as base for DPT implementations - use _DPTPayload as base class for DPTArray and DPTBinary (for future type checking)
 - DPT classes can now be searched via value_type string or dpt number from any parent class (DPTTranscoder for all) to be used in Sensor
 - Use RemoteValue class in BinarySensor, DateTime and ClimateMode device
 - use time.struct_time for internal time and date representation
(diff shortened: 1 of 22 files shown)


marvin-w (approved):
Looks good, but you need to resolve the conflicts again.. sorry!

marvin-w:
@farmio Feel free to merge after resolving the conflicts. Otherwise we will keep having conflicts.

marvin-w:
@farmio 

> rename DPTBase to DPTTranscoder

I think the Transcoder would be better off as a separate class as it doesn't have anything in common with all the DPTs. I'd like if it gets refactored. I think DPTBase as class name can and should stay as its the base class for all DPTs. The fact that the transcoder is in there just means that we broke the design principle separation of concerns.

> introduce _DPTPayload as base class for DPTBinary and DPTArray. For future type checking.

I like this change!

farmio:
I'll give this another shot without the renaming but only
- introduce _DPTPayload as base class for DPTBinary and DPTArray and 
> I was also thinking of changing the DPTBase.to_knx() return values to instances of _DPTPayload.

and hopefully some type checking in a new PR soon.

PR #321, opened 2020-08-03
Passive group addresses

farmio:
This allows configuration for multiple group addresses per RemoteValue. These shall not be queried or written to, but telegrams addressing these shall be processed. (german term: Hörende Gruppenadressen)

Example usage: Covers that don't support position status and are triggered by either a normal GA or a central GA (addressing more than one).
```python
cover = Cover(
    xknx,
    group_address=["1/2/3", "0/0/1"])
```
where "1/2/3" is the normal group address and "0/0/1" is the central GA.

- [ ] maybe add additional attribute (eg. "group_addresses_passive" to RemoteValues and `Devices) instead of checking for a list
- [ ] adapt HA integration config validation


marvin-w:
Closed in favour of #441

PR #322, opened 2020-08-03
Cover update - up/down from bus; stop and movement direction

### LINKED ISSUE #25: Shutter/cover: Implement trigger (up/down/stop) from bus side
Implement trigger (up/down/stop) from bus side

### LINKED ISSUE #105: position_state_address fails with old KNX devices

### LINKED ISSUE #257: is_open / is_closed not correctly reported if position is inverted

### LINKED ISSUE #314: Rework is_open for covers
**Description of problem:**
I already created an issue in the HA repository (https://github.com/home-assistant/core/issues/34330), but I realized the problem lies in XKNX. 

Problem shortcut: My garage door can be controlled with knx. It just supports up/down and **cannot** report the current position of the garage door. In the current implementation, I cannot open / close the garage door via HA, because the reported state is not correct in 50% of the time. Same problem occurs with central covers like in my HA issue.
image

Even for covers which can report the current position, xknx uses the time in the travelcalculator and not the actual position of the cover.

With boils down to:
**The cover is_open and is_closed is not correct for knx covers with and without position state address.**

My proposal:
- knx cover **with** `group_address_position_state`: Use the actual position comparison with 100% or 0% to determine is_open and is_closed

- knx cover **without** `group_address_position_state`: Return `None` for both is_open and is_closed. The travelcalculator is a nice effort, but it doesn't help if the position was changed outside of XKNX. Therefore I believe it can't be used to determine the state at all.

I can provide the implementation if it has a chance to get it pulled ;)

- [ ] using xknx standalone

**Version information:**
- xknx / Home-Assistant release with the issue: 102

**KNX installation:**

farmio:
this is derived from #259 so 
- in xknx 0% is open and 100% is closed now
- adds is_opening and is_closing
- react to up/down from bus and start travel calculator
- add group_address_stop for actuators that have it (or don't support step)

resolves #314 
fixes #257 (I hope)
fixes #105 (without implementing the DPT1 state address - just listening for the normal group_address so initial status may be off)
closes #25

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,28 @@
-sudo: false
+os: linux
+dist: xenial
+language: python
 
-matrix:
+jobs:
   include:
-    - python: "3.6"
+    - python: "3.7"
       env: TOXENV=flake8
-    - python: "3.6"
+    - python: "3.7"
       env: TOXENV=pylint
-    - python: "3.6"
+    - python: "3.7"
       env: TOXENV=pydocstyle
     - python: "3.5"
       env: TOXENV=py35
+      after_success: coveralls
     - python: "3.6"
       env: TOXENV=py36
+      after_success: coveralls
     - python: "3.7"
-      dist: xenial # required for Python >= 3.7
       env: TOXENV=py37
-#    - python: "3.7-dev"
-#      env: TOXENV=py36
-#    - python: "nightly"
-#      env: TOXENV=py36
+      after_success: coveralls
 
 cache:
   directories:
     - $HOME/.cache/pip
-language: python
-script: travis_wait tox
-after_success: coveralls
-
 # command to install dependencies
 install: "pip install -r requirements/testing.txt"
-
-after_success: coveralls
+script: travis_wait tox
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ pydocstyle:
 	 @pydocstyle xknx test/*.py test/*.py *.py examples/*.py
 
 coverage:
-	py.test --cov-report html --cov xknx --verbose
+	pytest --cov-report html --cov xknx --verbose
 
 clean:
 	-rm -rf build dist xknx.egg-info
--- a/changelog.md
+++ b/changelog.md
@@ -15,6 +15,7 @@
 - Added config option ignore_internal_state in binary sensors (@andreasnanko #267)
 - Add support for 2byte float type (DPT 9.002) to climate shiftpoint
 - ClimateMode: add `group_address_operation_mode_standby` as binary operation mode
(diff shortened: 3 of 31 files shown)


farmio:
Up/down from bus seems to have issues with the `auto_stop_if_necessary` function. When I set a down telegram on a tunnel a similar telegram is received as answer (not yet really sure why). This overwrites the `travel_to_position` value with 100% and auto stopping fails. 

Workarounds:
- don't Support absolute positioning without position group address or
- don't process updown when `is_traveling() == True` this could lead to wrong positions when travel direction is changed at travel. Maybe ignore updown for the same direction as current movement.

PR #325, opened 2020-08-13
Version 0.12.0

farmio:
Merge the `next` branch into `master` to be able to do a new release. 

### Breaking changes

- Climate: `setpoint_shift_step` renamed for `temperature_step`. This attribute can be applied to all temperature modes. Default is `0.1`
- Removed significant_bit attribute in BinarySensor
- DateTime devices are initialized with sting for broadcast_type: "time", "date" or "datetime" instead of an Enum value
- Removed `bind_to_multicast` option in ConnectionConfig and UDPClient

### New Features

- Cover: add optional `group_address_stop` for manual stopping
- Cover: start travel calculator when up/down telegram from bus is received
- HA integration: `knx.send` service takes `type` attribute to allow sending DPT encoded values like `sensor`
- HA integration: `sensor` and `expose` accept int and float values for `type` (parsed as DPT numbers)
- new StateUpdater: Devices `sync_state` can be set to `init` to just initialize state on startup, `expire [minutes]` to read the state from the KNX bus when it was not updated for [minutes] or `every [minutes]` to update it regularly every [minutes]
- Sensor and ExposeSensor now also accepts `value_type` of int (generic DPT) or float (specific DPT) if implemented.
- Added config option ignore_internal_state in binary sensors (@andreasnanko #267)
- Add support for 2byte float type (DPT 9.002) to climate shiftpoint
- ClimateMode: add `group_address_operation_mode_standby` as binary operation mode
- ClimateMode: add `group_address_heat_cool` and `group_address_heat_cool_state for switching heating mode / cooling mode with DPT1

### Bugfixes

- Tunneling: don't process incoming L_Data.con confirmation frames. This avoids processing every outgoing telegram twice.
- enable multicast on macOS and fix a bug where unknown cemi frames raise a TypeError on routing connections
- BinarySensor: reset_after is now implemented as asyncio.Task to prevent blocking the loop
- ClimateMode: binary climate modes should be fully functional now (sending, receiving and syncing)
- Cover: position update from bus does update current position, but not target position (when moving)

### Internals

- Cover travelcalculator doesn't start from 0% but is initialized by first movement or status telegram
- Cover uses 0% for open cover and 100% for closed cover now
- DPT classes can now be searched via value_type string or dpt number from any parent class (DPTBase for all) to be used in Sensor
- Use RemoteValue class in BinarySensor, DateTime and ClimateMode device
- use time.struct_time for internal time and date representation
- use a regular Bool type for BinarySensor state representation
- RemoteValue.process has always_callback attribute to run the callbacks on every process even if the payload didn't change
- Separate incoming and outgoing telegram queues; apply rate limit only for outgoing telegrams
- Automatically publish packages to pypi (@Julius2342 #277)
- keep xknx version in `xknx/__version__.py` (@farmio #278)
- add raw_socket logger (@farmio #299)


farmio:
Hey @marvin-w could you please have a look at the Netlify tests failing here? I don't really know what this is doing and why its failing.

marvin-w:
@farmio Can you have a look at #326 please? It should fix it. Not sure why it didn't occur before. I think we only deploy previews when we merge to master.

PR #331, opened 2020-08-22
Migrate config to integration domain key.

marvin-w:
This pull request introduces a breaking change into how we handle the configuration of KNX.

It is required due to https://github.com/home-assistant/architecture/blob/master/adr/0007-integration-config-yaml-structure.md.

I've updated the documentation to include samples on how the new config will look like.


marvin-w:
Will redo after HA changes are merged.

PR #335, opened 2020-08-26
Merge upstream changes from HA

marvin-w:
Merges xknx-refactor-0.12.0-updates branch from HA to XKNX

### DIFF
--- a/home-assistant-plugin/custom_components/xknx/const.py
+++ b/home-assistant-plugin/custom_components/xknx/const.py
@@ -1,0 +1,61 @@
+"""Constants for the KNX integration."""
+from enum import Enum
+
+from homeassistant.components.climate.const import (
+    HVAC_MODE_AUTO,
+    HVAC_MODE_COOL,
+    HVAC_MODE_DRY,
+    HVAC_MODE_FAN_ONLY,
+    HVAC_MODE_HEAT,
+    HVAC_MODE_OFF,
+    PRESET_AWAY,
+    PRESET_COMFORT,
+    PRESET_ECO,
+    PRESET_SLEEP,
+)
+
+DOMAIN = "xknx"
+DATA_XKNX = "data_knx"
+
+CONF_STATE_ADDRESS = "state_address"
+CONF_SYNC_STATE = "sync_state"
+
+
+class ColorTempModes(Enum):
+    """Color temperature modes for config validation."""
+
+    absolute = "DPT-7.600"
+    relative = "DPT-5.001"
+
+
+class DeviceTypes(Enum):
+    """KNX device types."""
+
+    cover = "cover"
+    light = "light"
+    binary_sensor = "binary_sensor"
+    climate = "climate"
+    switch = "switch"
+    notify = "notify"
+    scene = "scene"
+    sensor = "sensor"
+
+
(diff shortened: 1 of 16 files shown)


home-assistant-plugin/custom_components/xknx/manifest.json
```diff
@@ -2,6 +2,6 @@
   "domain": "xknx",
   "name": "KNX",
   "documentation": "https://www.home-assistant.io/integrations/knx",
-  "requirements": [],
-  "codeowners": ["@Julius2342"]
+  "requirements": ["xknx==0.12.0"],
```
farmio, line 5: remove the requirement for the custom component so it always uses the local checked out version

home-assistant-plugin/custom_components/xknx/manifest.json
```diff
@@ -2,6 +2,6 @@
...
+  "codeowners": ["@Julius2342", "@farmio", "@marvin-w"]
```
farmio, line 6: didn't you want to use XKNX/xknx-devs ? Doesn't it work?

home-assistant-plugin/custom_components/xknx/const.py
```diff
@@ -0,0 +1,61 @@
...
+    HVAC_MODE_HEAT,
+    HVAC_MODE_OFF,
+    PRESET_AWAY,
+    PRESET_COMFORT,
+    PRESET_ECO,
+    PRESET_SLEEP,
+)
+
+DOMAIN = "xknx"
+DATA_XKNX = "data_knx"
```
farmio, line 18: should this be "data_xknx" in the cc to not conflict with the HA knx integration?

farmio (commented):
Just some small things considering the custom component.
I don't use it so I'm not sure if we should also change 
https://github.com/XKNX/xknx/blob/1c972e34ad977da14276bde5bf373267e857c904/home-assistant-plugin/custom_components/xknx/__init__.py#L328
to "xknx_event"

home-assistant-plugin/custom_components/xknx/manifest.json
marvin-w: We can do that in the next PR to HA.

home-assistant-plugin/custom_components/xknx/const.py
marvin-w: Good catch, will have a look. I just used the merge_from_upstream - thinking that it would just work :(

marvin-w: Actually... Why would you want to run both at the same time? That doesn't make sense to me

farmio: ¯\_(ツ)_/¯ 
https://community.home-assistant.io/t/knx-with-multiple-interfaces/206120/4

https://community.home-assistant.io/t/how-to-use-two-or-more-ip-interfaces/214822/4

home-assistant-plugin/custom_components/xknx/manifest.json
marvin-w: I added a new version of the scripts that do this stuff automatically.

For backport: simply remove it
For upstream PR: get latest release version from github and use that as requirement

PR #338, opened 2020-08-27
Create new weather entity

marvin-w:
Create dedicated weather device

### DIFF
--- a/home-assistant-plugin/custom_components/xknx/__init__.py
+++ b/home-assistant-plugin/custom_components/xknx/__init__.py
@@ -148,11 +148,13 @@ async def async_setup(hass, config):
                 create_knx_device(
                     hass, platform, hass.data[DATA_XKNX].xknx, device_config
                 )
-            hass.async_create_task(
-                discovery.async_load_platform(
-                    hass, platform.value, DOMAIN, {}, config
-                )
+
+    for platform in SUPPORTED_PLATFORMS:
+        hass.async_create_task(
+            discovery.async_load_platform(
+                hass, platform.value, DOMAIN, {}, config
             )
+        )
 
     hass.services.async_register(
         DOMAIN,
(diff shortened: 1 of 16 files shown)


home-assistant-plugin/custom_components/xknx/__init__.py
```diff
@@ -148,11 +148,13 @@ async def async_setup(hass, config):
                 create_knx_device(
                     hass, platform, hass.data[DATA_XKNX].xknx, device_config
                 )
-            hass.async_create_task(
-                discovery.async_load_platform(
-                    hass, platform.value, DOMAIN, {}, config
-                )
+
```
farmio, line 151:
```suggestion
    # load platforms when creation of all devices is finished (weather sensor exposures would not be loaded)
```
add a comment to remember why we do the same loop twice

home-assistant-plugin/custom_components/xknx/__init__.py
```diff
@@ -148,11 +148,13 @@ async def async_setup(hass, config):
...
+    for platform in SUPPORTED_PLATFORMS:
+        hass.async_create_task(
+            discovery.async_load_platform(
+                hass, platform.value, DOMAIN, {}, config
             )
+        )
```
farmio, line 157: should we filter for only configured platforms? or is it a negligible overhead to just load all platforms even if there are no devices? 

It does create log entries for all supported platforms, even if not configured. Eg. for cover:
> [homeassistant.loader] Loaded cover from homeassistant.components.cover
> [homeassistant.setup] Setting up cover
> [homeassistant.setup] Setup of domain cover took 0.0 seconds
> [homeassistant.components.cover] Setting up cover.xknx

marvin-w: Since it's all running async anyway there is no performance downside to it. I'd just leave it as it is.

PR #340, opened 2020-08-27
Adds the devices to the device list automatically after instantiation

### LINKED ISSUE #337: Add new Device to xknx.devices automatically
Should we add a new Device to xknx.Devices on init? Currently newly created devices have to be added manually to receive incoming telegrams.
https://github.com/XKNX/xknx/blob/fb622f76f2e8b8de0454f94558b16a88c6655a33/xknx/core/telegram_queue.py#L144

```
xknx = XKNX()
sensor = Sensor(xknx, ...)
xknx.devices.add(sensor)
```
I'd like to remove this boilerplate and call `xknx.devices.add(self)` in `Device.__init__()` - and remove it in `__del__()`

### LINKED ISSUE #339: 0.12.0 Duplicate names result in incorrect configuration processing
**Description of problem:**
I adjusted the HA YAML configuration file to the new layout as part of the migration to XKNX 0.12.0. There however seems to be a bug with the configuration parsing. 

I have two entries with the same name "KNX Office" one being a light, one being a cover which is throwing off the logic since it is now attempting to interpret the objects as the wrong type (see log output below). I assume somewhere it is attempting to look up the content of the configuration file only via the name instead of platform and name which would then be unique.

Once I change one of the names to "KNX Office 2" everything works fine, so my defined values/parameters themselves are correct.

- [ ] using xknx standalone
- [X] using Home-Assistant knx plugin

**Version information:**
- xknx / Home-Assistant release with the issue: 0.12.0 / HA latest dev

**KNX installation:**

```YAML
knx:
  tunneling:
    host: "192.168.0.222"
  cover:
    - name: KNX Office
      [...]
  light:
    - name: KNX Office
      address: "1/1/29"
      state_address: "1/4/28"
      brightness_address: "1/5/18"
      brightness_state_address: "1/3/15"
```

```
2020-08-27 11:35:36 ERROR (MainThread) [homeassistant.components.light] Error while setting up knx platform for light
  File "/home/pi/git/core/homeassistant/helpers/entity_platform.py", line 193, in _async_setup_platform
    await asyncio.shield(task)
  File "/home/pi/git/core/homeassistant/components/knx/light.py", line 28, in async_setup_platform
    async_add_entities_discovery(hass, discovery_info, async_add_entities)
  File "/home/pi/git/core/homeassistant/components/knx/light.py", line 37, in async_add_entities_discovery
    entities.append(KNXLight(device))
  File "/home/pi/git/core/homeassistant/components/knx/light.py", line 48, in __init__
    self._min_kelvin = device.min_kelvin
AttributeError: 'Cover' object has no attribute 'min_kelvin'
```

marvin-w:
Before this change all devices had to be manually added to the xknx.devices List.
This was cumbersome and not needed, thats why it was removed.

Additionally, the device class now has a unique_id property that is used within the HA
integration in order to avoid errors where multiple devices share the same name within
different platforms.

Fixes #339, fixes #337

### DIFF
--- a/home-assistant-plugin/custom_components/xknx/__init__.py
+++ b/home-assistant-plugin/custom_components/xknx/__init__.py
@@ -24,7 +24,7 @@ from homeassistant.helpers import discovery
 import homeassistant.helpers.config_validation as cv
 from homeassistant.helpers.event import async_track_state_change_event
 
-from .const import DATA_XKNX, DOMAIN, DeviceTypes
+from .const import DATA_XKNX, DOMAIN, SupportedPlatforms
 from .factory import create_knx_device
 from .schema import (
     BinarySensorSchema,
@@ -51,21 +51,10 @@ CONF_XKNX_STATE_UPDATER = "state_updater"
 CONF_XKNX_RATE_LIMIT = "rate_limit"
 CONF_XKNX_EXPOSE = "expose"
 
-CONF_XKNX_LIGHT = "light"
-CONF_XKNX_COVER = "cover"
-CONF_XKNX_BINARY_SENSOR = "binary_sensor"
-CONF_XKNX_SCENE = "scene"
-CONF_XKNX_SENSOR = "sensor"
-CONF_XKNX_SWITCH = "switch"
-CONF_XKNX_NOTIFY = "notify"
-CONF_XKNX_CLIMATE = "climate"
-
 SERVICE_XKNX_SEND = "send"
 SERVICE_XKNX_ATTR_ADDRESS = "address"
 SERVICE_XKNX_ATTR_PAYLOAD = "payload"
 SERVICE_XKNX_ATTR_TYPE = "type"
-
-ATTR_DISCOVER_DEVICES = "devices"
 
 CONFIG_SCHEMA = vol.Schema(
     {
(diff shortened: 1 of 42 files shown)


home-assistant-plugin/custom_components/xknx/__init__.py
```diff
@@ -156,26 +154,23 @@ async def async_setup(hass, config):
...
+        "switch",
+        "climate",
+        "cover",
+        "light",
+        "sensor",
+        "binary_sensor",
+        "scene",
+        "notify",
     ):
-        found_devices = _get_devices(hass, discovery_type)
```
farmio, line 175: now a `discovery.async_load_platform` Task is started for every platform we support. It is not filtered for only configured / used platforms. Is this a problem?

home-assistant-plugin/custom_components/xknx/__init__.py
```diff
@@ -156,26 +154,23 @@ async def async_setup(hass, config):
     for platform_config, device_type in KNX_CONFIG_PLATFORM_MAPPING.items():
```
farmio, line 154: what is KNX_CONFIG_PLATFORM_MAPPING for?
{"switch": "switch",
"light": "light",
...} 
platform_config and device_type seems to be equal.

xknx/devices/device.py
```diff
@@ -80,3 +82,19 @@ async def do(self, action):
...
+            if key not in other.__dict__:
+                return False
+            if other.__dict__[key] != value:
+                return False
+        for key, value in other.__dict__.items():
+            if key == "device_updated_cbs":
+                continue
+            if key not in self.__dict__:
+                return False
+        return True
```
farmio, line 100:
```suggestion
    def __eq__(self, other):
        """Compare for equality."""
        return self.__dict__ == other.__dict__
```
should be alright. We will have to find out why the one callback is added twice and thus breaks the config-tests.

.travis.yml
```diff
@@ -10,9 +10,6 @@ jobs:
       env: TOXENV=pylint
     - python: "3.8"
       env: TOXENV=pydocstyle
-    - python: "3.5"
-      env: TOXENV=py35
-      after_success: coveralls
```
farmio, line 15: I think all changes currently are still 3.5 compatible. Maybe we should remove 3.5 in an extra PR?

home-assistant-plugin/custom_components/xknx/__init__.py
marvin-w: No, the filtering happens directly in the platform itself.

xknx/devices/device.py
marvin-w: The object generated from the config only had the device_update cb from devices.py (in the add method).

The other object we instantiated had the device_update_cb from the constructor and the device_update_db from the add method in devicey.py. Removing the constructor argument worked.

home-assistant-plugin/custom_components/xknx/notify.py
```diff
@@ -28,7 +20,7 @@ def async_get_service_discovery(hass, discovery_info):
 class KNXNotificationService(BaseNotificationService):
     """Implement demo notification service."""
 
-    def __init__(self, devices: XknxNotification):
+    def __init__(self, devices: list):
```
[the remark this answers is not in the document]
marvin-w: Done

marvin-w:
I took care of all the review comments. Let's see what Travis says :-D.

marvin-w:
Merged to next - so that I can continue working on the weather device.

Issue #349, opened 2020-08-30
Improve reconnection logic


farmio:
Every now and then a lost connection to KNX bus is not automatically recovered by xknx. 

It is quite hard for me to reproduce this, as it only happens on my system when I update my LAN-routers software. Maybe someone has a hint how to reliably reproduce a connection loss.

I think we should 
- hook into 
https://github.com/XKNX/xknx/blob/dae16a0d353323228f872986b4aedc47ed3eb352/xknx/io/udp_client.py#L59
to reliably track the connection state of our socket and reconnect if there was an error
- maybe add a self increasing retry delay
- stop the StateUpdater and restart it after reconnection to make sure we did not miss any telegrams
- clear the outgoing telegram queue (should we? maybe after x seconds of disconnection)
- on an Error shut down the whole interface (and UDPClient) and create a new one from KNXIPInterface so the connection_config can be read again. This would enable to connect to (or from) a different IP, KNX/IP Interface or even connection type when one interface fails.

farmio:
One possible error that gets passed to https://github.com/XKNX/xknx/blob/dae16a0d353323228f872986b4aedc47ed3eb352/xknx/io/udp_client.py#L59 would be `[Errno 101] Network unreachable`.
https://github.com/XKNX/xknx/blob/dae16a0d353323228f872986b4aedc47ed3eb352/xknx/io/udp_client.py#L64 isn't called in this case.

farmio:
I think #774 closes this issue.

PR #351, opened 2020-08-30
Harmonize log messages + include timeout value

spacegaier:
Some harmonization and spelling corrections plus inclusion of used timeout values in the log messages.

### DIFF
--- a/home-assistant-plugin/custom_components/xknx/__init__.py
+++ b/home-assistant-plugin/custom_components/xknx/__init__.py
@@ -137,9 +137,9 @@ async def async_setup(hass, config):
         hass.data[DATA_XKNX].async_create_exposures()
         await hass.data[DATA_XKNX].start()
     except XKNXException as ex:
-        _LOGGER.warning("Cannot connect to KNX interface: %s", ex)
+        _LOGGER.warning("Could not connect to KNX interface: %s", ex)
         hass.components.persistent_notification.async_create(
-            f"Cannot connect to KNX interface: <br><b>{ex}</b>", title="KNX"
+            f"Could not connect to KNX interface: <br><b>{ex}</b>", title="KNX"
         )
 
     for platform in SUPPORTED_PLATFORMS:
--- a/home-assistant-plugin/custom_components/xknx/__init__.py
+++ b/home-assistant-plugin/custom_components/xknx/__init__.py
@@ -137,9 +137,9 @@ async def async_setup(hass, config):
         hass.data[DATA_XKNX].async_create_exposures()
         await hass.data[DATA_XKNX].start()
     except XKNXException as ex:
-        _LOGGER.warning("Can't connect to KNX interface: %s", ex)
+        _LOGGER.warning("Cannot connect to KNX interface: %s", ex)
         hass.components.persistent_notification.async_create(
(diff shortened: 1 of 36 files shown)


home-assistant-plugin/custom_components/xknx/__init__.py
```diff
@@ -137,9 +137,9 @@ async def async_setup(hass, config):
         hass.data[DATA_XKNX].async_create_exposures()
         await hass.data[DATA_XKNX].start()
     except XKNXException as ex:
-        _LOGGER.warning("Can't connect to KNX interface: %s", ex)
+        _LOGGER.warning("Cannot connect to KNX interface: %s", ex)
```
marvin-w, line 140: Can we use `Could not` or what was the intention of changing this?

marvin-w (changes requested):

xknx/io/udp_client.py
```diff
@@ -64,7 +64,7 @@ def error_received(self, exc):
         def connection_lost(self, exc):
             """Log error. Callback for connection lost."""
             if hasattr(self, 'xknx'):
-                self.xknx.logger.info('closing transport %s', exc)
+                self.xknx.logger.info('Closing transport: %s', exc)
```
marvin-w, line 67: In my logs `exc` is always None. Maybe we can just remove the parameter? Or did you ever had this set to something else?

xknx/io/request_response.py
```diff
@@ -67,8 +67,8 @@ def on_error_hook(self, knxipframe):
 
     def timeout(self):
         """Handle timeout for not having received expected knxipframe."""
-        self.xknx.logger.warning("Error: KNX bus did not respond in time to request of type '%s'",
-                                 self.__class__.__name__)
+        self.xknx.logger.warning("Error: KNX bus did not respond in time (%s secs) to request of type '%s'",
+                                 self.timeout_in_seconds, self.__class__.__name__)
```
marvin-w, line 71: Nice!

xknx/io/udp_client.py
spacegaier: Always "None" for me as well so far. But I wasn't sure if there might be other scenarios, so I did not touch that.

home-assistant-plugin/custom_components/xknx/__init__.py
spacegaier: "Could not" also works. I just wanted to get rid of the informal "Can't" and the typo "Cant" while I was already adjusting other messages.

marvin-w: I personally don't care. If you want to change it again go ahead (I'd like that) but if not that's also no issue for me. In the end its the same.

xknx/io/udp_client.py
marvin-w: You can delete it IMHO

marvin-w (commented):
Looks good! Thanks for your contribution!

xknx/io/udp_client.py
farmio: It is `None` when we close the connection or it reaches EOF. 
If the connection gets terminated from the other side there should be an exception.
We can explore this further when handling #349

PR #354, opened 2020-08-30
Datetime: fix hourly broadcasting localtime

farmio:
This was previously done by the state_updater calling DateTime.sync() once every hour. New state_updater (0.12) doesn't do that.
A Task is introduced to handle hourly broadcasting of localtime in the Device itself.

### DIFF
--- a/home-assistant-plugin/custom_components/xknx/__init__.py
+++ b/home-assistant-plugin/custom_components/xknx/__init__.py
@@ -337,7 +337,7 @@ class KNXExposeTime:
             name=self.expose_type.capitalize(),
             broadcast_type=self.expose_type.upper(),
             localtime=True,
-            group_address=self.address
+            group_address=self.address,
         )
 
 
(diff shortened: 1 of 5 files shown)


home-assistant-plugin/custom_components/xknx/__init__.py
```diff
@@ -267,7 +267,7 @@ def async_create_exposures(self):
             attribute = to_expose.get(ExposeSchema.CONF_XKNX_EXPOSE_ATTRIBUTE)
             default = to_expose.get(ExposeSchema.CONF_XKNX_EXPOSE_DEFAULT)
             address = to_expose.get(ExposeSchema.CONF_XKNX_EXPOSE_ADDRESS)
-            if expose_type in ["time", "date", "datetime"]:
+            if expose_type.lower() in ["time", "date", "datetime"]:
```
marvin-w, line 270: That's a good addition to the logic!

marvin-w (approved):
Looks good, as spoken we will have to take care of the failing test. I'll give it another shot later.

PR #365, opened 2020-09-01
[Refactoring] Migrate binary sensor automation scheme to HA automations

marvin-w:
**This is a breaking change!**

In order to be able to support config flows later we need to streamline the way we work with our xknx devices. With the exception of the binary sensors it was already possible to use this. However, binary_sensors were still relying on scripting inside of HASS which makes it impossible to migrate to the config flow completely.

This PR adds a callback to the BinarySensor within XKNX that gets fired whenever the state changes. HA registers a callback and fires an event that you can use in your automations instead of defining them in the automation schema inside the binary sensor.

Example:

 ```yaml
 -  name: cover_abstell 
    state_address: "2/0/33"
    automation:
      - counter: 1
        hook: 'on'
        action:
          - entity_id: cover.sonne_abstellkammer
            service: cover.open_cover
      - counter: 1
        hook: 'off'
        action:
          - entity_id: cover.sonne_abstellkammer
            service: cover.close_cover

```

will be:

binary sensor:

```yaml
- name: cover_abstell
  state_address: "2/0/33"
```

for the automation part add an automation config with the following content:

```yaml

automation:
  - trigger:
      - platform: event
        event_type: knx_binary_sensor.cover_abstel
        event_data:
            counter: 1
            state: "on"
    action:
      - service: cover.open_cover
        entity_id: cover.sonne_abstellkammer

  - trigger:
     - platform: event
       event_type: knx_binary_sensor.cover_abstell
       event_data:
            counter: 1
            state: "off"
    action:
      - service: cover.close_cover
        entity_id: cover.sonne_abstellkammer
  
```

Execution:

```
2020-09-02 01:20:35 INFO (MainThread) [homeassistant.components.automation.automation_0] automation 0: Running automation actions
2020-09-02 01:20:35 INFO (MainThread) [homeassistant.components.automation.automation_0] automation 0: Executing step call service
2020-09-02 01:20:41 INFO (MainThread) [homeassistant.components.automation.automation_1] automation 1: Running automation actions
2020-09-02 01:20:41 INFO (MainThread) [homeassistant.components.automation.automation_1] automation 1: Executing step call service
2020-09-02 01:20:56 INFO (MainThread) [homeassistant.components.automation.automation_0] automation 0: Running automation actions
2020-09-02 01:20:56 INFO (MainThread) [homeassistant.components.automation.automation_0] automation 0: Executing step call service
```


farmio:
Is there a difference between a BinarySensor with a `fire_event` callback and one having `ignore_internal_state` set. Could these be used in automations as well?

marvin-w:
Yes, they would just trigger on every update not only when the state has changed.

farmio:
This is exactly what `ignore_internal_state` does.

xknx/devices/binary_sensor.py
```diff
@@ -109,6 +109,11 @@ async def _set_internal_state(self, state):
                 if action.test_if_applicable(self.state, counter):
                     await action.execute()
 
+    @property
+    def counter(self):
```
marvin-w, line 113: @farmio If we just use the `self._count_set_on` and the set_off properties we would not need to make a new release in order to ship this to HA. What's you opinion on this?

farmio (commented):
for after_update_callback to get triggered even when "OFF" was sent twice, `ignore_internal_state` has to be set. So to get the count working I think we should default it to True.

I'm by far not an HA automation expert, but I think 
```yaml
automation:
  trigger:
    platform: state
    entity_id: binary_sensor.<entity_name>
```
would do exactly the same, without the need for an additional HA event. The count could be passed by an entity attribute.

marvin-w:
@farmio Thanks for the review. I have worked in almost all your points.

> for after_update_callback to get triggered even when "OFF" was sent twice, ignore_internal_state has to be set. So to get the count working I think we should default it to True.

I've honestly never used the counter, but they will work exactly as they did before. If the user wants this he/she can enable it (as it is now).

Another thing I noticed: When you initialize the binary sensor and the self.state attribute is initialized with False meaning that pressing OFF on your wall switch wouldn't do anything since that's already set (and self.state == state + ignore_internal_state = False). Maybe we should default the state to None instead?

> [..] would do exactly the same, without the need for an additional HA event. And it would tie in to the way other domains are configured. The count could be passed by an entity attribute.

We can only reference state attributes in automations. Binary Sensors only have on/off as state attribute. All other attributes (and also custom ones) would not be part of the state machine and thus never get triggered upon.

test/devices_tests/binary_sensor_test.py
```diff
@@ -11,6 +10,11 @@
 from xknx.telegram import GroupAddress, Telegram
 
 
+class AsyncMock(MagicMock):
```
marvin-w, line 13: AsyncMock was added to `unittest.mock` only after 2.8 and was then made the default. When we drop 3.6.and 3.7 support we can remove this again.

marvin-w:
@farmio I've integrated what we said. The event now contains the entity_id and I moved all the code to our library to keep the integration clean.

Users will also only experience a delay if the opt in for the `ignore_internal_state` attribute - which is still defaulting to False.

I'd like for this to be merged and integrated in the current HA release. Any objections?

marvin-w:
@farmio `context_timeout` is now configurable. The default state for `ignore_internal_state` has changed to `True`. Documentation is updated and migration guide has been adjusted and splitted into the proper HA versions.

I think we can merge this one, what do you say?

home-assistant-plugin/custom_components/xknx/binary_sensor.py
```diff
@@ -27,10 +30,19 @@ def __init__(self, device: XknxBinarySensor):
     def async_register_callbacks(self):
         """Register callbacks to update hass after device was changed."""
 
-        async def after_update_callback(device):
+        async def after_update_callback(device: XknxBinarySensor):
             """Call after device was updated."""
             self.async_write_ha_state()
 
+            self.hass.bus.fire(
+                f"knx_{self.entity_id}",
```
farmio, line 38:
```suggestion
                f"{DOMAIN}_{self.entity_id}",
```
We could avoid constants in code if we import the DOMAIN constant. The custom component would then fire a `xknx_binary_sensor` event instead of a `knx_binary_sensor` event - don't know if this makes sense.
I'm good with either option.

home-assistant-plugin/custom_components/xknx/schema.py
```diff
@@ -113,11 +100,11 @@ class BinarySensorSchema:
                     cv.boolean,
                     cv.string,
                 ),
-                vol.Optional(CONF_IGNORE_INTERNAL_STATE, default=False): cv.boolean,
+                vol.Optional(CONF_IGNORE_INTERNAL_STATE, default=True): cv.boolean,
+                vol.Optional(CONF_CONTEXT_TIMEOUT, default=1): cv.string,
```
farmio, line 104:
```suggestion
                vol.Optional(CONF_CONTEXT_TIMEOUT, default=1): cv.positive_int,
```
or even better with float (0.5 seconds would be a viable option imho)
```suggestion
                vol.Optional(CONF_CONTEXT_TIMEOUT, default=1.0): vol.All(vol.Coerce(float), vol.Range(min=0, max=10)),
```

xknx/devices/binary_sensor.py
```diff
@@ -90,24 +97,47 @@ def from_config(cls, xknx, name, config):
...
+            if self.ignore_internal_state:
+                if self._context_task:
+                    self._context_task.cancel()
+                self._context_task = self.xknx.loop.create_task(
+                    self._counter_task(self._context_timeout)
+                )
+            else:
+                await self._trigger_callbacks()
+
+    async def _counter_task(self, wait_seconds: int):
```
farmio, line 124:
```suggestion
    async def _counter_task(self, wait_seconds: float):
```

farmio (approved):
Great!

PR #372, opened 2020-09-02
HA: configurable IA, Multicast group and port

farmio:
I'm still not happy with having to set `local_ip` manually for routing connections, but I haven't found a way to configure it so that
```yaml
knx:
  routing:
  expose:
    ....
```
or similar would be possible. 

Should we change the configuration to something like that:
```yaml
knx:
  connection_type: routing # or "tunneling"
  local_ip: 123 # optional - would fit for routing and tunnelling
# or
knx:
  connection_type: tunneling
  tunneling:
    host: 10.1.1.1
    port: 1234
# or
knx:
  connection_type: tunneling
  tunneling_host: 10.1.1.1
  tunneling_port: 1234
```
?

### DIFF
--- a/home-assistant-plugin/custom_components/xknx/__init__.py
+++ b/home-assistant-plugin/custom_components/xknx/__init__.py
@@ -6,8 +6,13 @@ from xknx import XKNX
 from xknx.devices import DateTime, ExposeSensor
 from xknx.dpt import DPTArray, DPTBase, DPTBinary
 from xknx.exceptions import XKNXException
-from xknx.io import DEFAULT_MCAST_PORT, ConnectionConfig, ConnectionType
-from xknx.telegram import AddressFilter, GroupAddress, Telegram
+from xknx.io import (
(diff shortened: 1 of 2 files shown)


marvin-w (approved):
Changes look good. I like the last option most.

```
knx:
  connection_type: tunneling
  tunneling_host: 10.1.1.1
  tunneling_port: 1234
```

One thing:

In the tunneling connection schema you could as well define a default port:

```
gateway_port = self.config[DOMAIN][CONF_XKNX_TUNNELING].get(CONF_PORT)
        local_ip = self.config[DOMAIN][CONF_XKNX_TUNNELING].get(
            ConnectionSchema.CONF_XKNX_LOCAL_IP
        )
        if gateway_port is None:
            gateway_port = DEFAULT_MCAST_PORT
```

Just so that it's nice and tidy.

marvin-w:
Looks good now, thanks! I'd suggest opening another PR for changing the connection params.

PR #377, opened 2020-09-05
Moat/Anyio cherry pick

farmio:
I have cherry-picked some commits form #304

### DIFF
--- a/xknx/devices/climate.py
+++ b/xknx/devices/climate.py
@@ -11,7 +11,6 @@ from xknx.remote_value import (
     RemoteValueSwitch,
     RemoteValueTemp,
 )
-from xknx.telegram import GroupAddress
 
 from .climate_mode import ClimateMode
 from .device import Device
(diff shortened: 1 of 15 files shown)


xknx/devices/climate.py
```diff
@@ -58,13 +57,6 @@ def __init__(
         """Initialize Climate class."""
         # pylint: disable=too-many-arguments, too-many-locals, too-many-branches, too-many-statements
         super().__init__(xknx, name, device_updated_cb)
-        if isinstance(group_address_on_off, (str, int)):
```
marvin-w, line 61: Why is this removed?

marvin-w (changes requested):
General idea looks fine. We just have to verify that all the devices still work. Also I think there is one code snippet that should not be deleted?

xknx/devices/climate.py
farmio: now we can pass GroupAddress() objects to all group_address_* attributes. If something unparsable is passed, an error will be raised when instantiating the GroupAddress in RemoteValue.
There should be no need to double-check types here.

marvin-w: Ah, now it makes sense. We only use that in the constructor.

marvin-w (approved):

PR #378, opened 2020-09-05
use the IA provided by a tunnelling server only for the current connection

farmio:
Don't replace `xknx.own_address` with the address assigned by a tunnelling server - just use it for the tunnel.
If the tunnel is closed and a routing connection is opened with the same XKNX object, the configured address shall be used.


marvin-w (approved):
Looks good!

PR #380, opened 2020-09-06
feature names for weather device

### LINKED ISSUE #352: Change feature_names of weather device
In order to have clean error messages we should add the correct feature_names to the `RemoteValueSensors` in the new weather device.

farmio:
closes #352


marvin-w (approved):
Thanks!

Issue #393, opened 2020-09-10
Process our outgoing telegrams


farmio:
Currently we don't seem to process our outgoing telegrams for our devices. This means that devices that don't support dedicated state addresses get out of sync with the bus if we trigger it from a 2nd device.

Example:
```python
    exposure = ExposeSensor(xknx, "out", group_address="1/0/45", value_type="binary")
    light = Light(
        xknx,
        name="TestLight",
        group_address_switch="1/1/45",
        group_address_switch_state="1/0/45",
    )

    await xknx.start(connection_config=connection_config)

    await light.sync(wait_for_result=True)
    print("Light state: ", light.state) # True
    await exposure.set(False)
    print("Light state: ", light.state) # True
    await asyncio.sleep(1)
    print("Light state: ", light.state) # True
```
The state address "1/0/45" has received a `0` from xknx on the bus, thus its state should be considered `False`.
Should we process the telegram for all devices using the GA after sending it to the bus in 
https://github.com/XKNX/xknx/blob/4d2a05066e05fecb1433a16702a810e908dc2069/xknx/core/telegram_queue.py#L130
? 
eg. add 
```python 
            if telegram.telegramtype == TelegramType.GROUP_WRITE:
                for device in self.xknx.devices.devices_by_group_address(
                    telegram.group_address
                ):
                    await device.process(telegram)
```

I think we could then omit
https://github.com/XKNX/xknx/blob/4d2a05066e05fecb1433a16702a810e908dc2069/xknx/remote_value/remote_value.py#L157
and possibly more because the telegram_queue takes care of this.

farmio:
So I was asking myself why this didn't came up before 🕵️‍♂️ It's because we recently removed double processing of outgoing telegrams in #324 - before this we did at least process the L_Data.CON

PR #394, opened 2020-09-10
process outgoing telegrams in devices

farmio:
process outgoing telegrams in all devices - regarding #393 
This enables xknx-devices that share a specific GA (without state_address) to be updated when a new value is sent form xknx.

I'm not sure if this would break anything - I guess it does. On my installation everything seems to work, but I don't use fancy binary_sensors with count and actions etc...

Updating the tests for this change will be non-trivial 😬

### DIFF
--- a/xknx/core/telegram_queue.py
+++ b/xknx/core/telegram_queue.py
@@ -129,10 +129,7 @@ class TelegramQueue:
         if self.xknx.knxip_interface is not None:
             await self.xknx.knxip_interface.send_telegram(telegram)
             if telegram.telegramtype == TelegramType.GROUP_WRITE:
-                for device in self.xknx.devices.devices_by_group_address(
-                    telegram.group_address
-                ):
-                    await device.process(telegram)
+                await self.xknx.devices.process(telegram)
         else:
             self.xknx.logger.warning("No KNXIP interface defined")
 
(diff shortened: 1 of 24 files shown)


xknx/core/telegram_queue.py
```diff
@@ -140,7 +145,7 @@ async def process_telegram_incoming(self, telegram):
                 ret = await telegram_received_cb.callback(telegram)
                 if ret:
                     processed = True
-
+        # TODO: why don't we process every incoming telegram by device?
```
farmio, line 148: The callbacks usually come from value_reader. I think if a knx-device responds, we should update all fo our devices that use the GA. 
Another source for these callbacks would be HAs fire_event where we use 
```python
        # False signals XKNX to proceed with processing telegrams.
        return False
```
so `processed` is False.

marvin-w (commented):
Tested with binary sensor count == 1 and count == 2, with ignore_internal_state on and off. Seems to work fine. I used the same set of tests that I used for the refactoring.

marvin-w (approved):
Tested and works fine.

farmio:
this should
fix https://github.com/home-assistant/core/issues/40351
fix https://github.com/home-assistant/core/issues/40316
fix https://github.com/home-assistant/core/issues/40359
closes #393

PR #413, opened 2020-09-21
Lower KNX log noise level

marvin-w:
This is needed to get the silver quality scale in HA.

### DIFF
--- a/xknx/io/request_response.py
+++ b/xknx/io/request_response.py
@@ -58,14 +58,10 @@ class RequestResponse:
 
     def on_success_hook(self, knxipframe):
         """Do something after having received a valid answer. May be overwritten in derived class."""
-        self.xknx.logger.debug(
-            "Success: Received correct answer from KNX bus: %s",
-            knxipframe.body.status_code,
-        )
 
     def on_error_hook(self, knxipframe):
         """Do something after having received error within given time. May be overwritten in derived class."""
-        self.xknx.logger.warning(
+        self.xknx.logger.debug(
             "Error: KNX bus responded to request of type '%s' with error in '%s': %s",
             self.__class__.__name__,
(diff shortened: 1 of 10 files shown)


xknx/io/udp_client.py
```diff
@@ -57,7 +57,7 @@ def error_received(self, exc):
 
         def connection_lost(self, exc):
             """Log error. Callback for connection lost."""
-            if hasattr(self, "xknx"):
+            if hasattr(self, "xknx") and exc is not None:
```
farmio, line 60: if we test for exc not None we should at least log it too...
or just remove the whole if and `pass`

xknx/io/request_response.py
```diff
@@ -58,14 +58,10 @@ def response_rec_callback(self, knxipframe, _):
...
         """Do something after having received a valid answer. May be overwritten in derived class."""
-        self.xknx.logger.debug(
-            "Success: Received correct answer from KNX bus: %s",
-            knxipframe.body.status_code,
-        )
 
     def on_error_hook(self, knxipframe):
         """Do something after having received error within given time. May be overwritten in derived class."""
-        self.xknx.logger.warning(
+        self.xknx.logger.debug(
```
farmio, line 64: I think this is a noteworthy error message. Why do we have to log this only in debug?

xknx/io/udp_client.py
marvin-w: Will add it, good point.

xknx/io/request_response.py
marvin-w: HA doesn't allow it since this only happens on connection errors. Users will have a warning once disconnected and an info once reconnected. As a user, this would suffice my needs - not only for HA. If they need more info they can still turn on debug logging.

PR #414, opened 2020-09-21
Allow covers to be inverted again

marvin-w:
Allow covers to be inverted again


farmio (approved):
Looks good.
Shall we test this or is it enough that invert is tested in RemoteValue* ?

marvin-w:
I checked again. Since invert_position had absolutely no tests I added some.

PR #418, opened 2020-09-22
Don't set ClimateMode to Standby

farmio:
... if the telegram was not processed in this ClimateMode device.

### DIFF
--- a/xknx/devices/climate_mode.py
+++ b/xknx/devices/climate_mode.py
@@ -251,14 +251,17 @@ class ClimateMode(Device):
     async def process_group_write(self, telegram):
         """Process incoming and outgoing GROUP WRITE telegram."""
         if self.supports_operation_mode:
+            processed = False
             for rv in self._iter_remote_values():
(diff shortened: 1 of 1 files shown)


xknx/devices/climate_mode.py
```diff
@@ -251,14 +251,17 @@ def gather_operation_modes(self):
...
             for rv in self._iter_remote_values():
                 if await rv.process(telegram):
                     # don't set when binary climate mode rv is False
                     if rv.value:
                         await self._set_internal_operation_mode(rv.value)
                         return
+                    processed = True
             # if no operation mode has been set and all binary operation modes are False
-            await self._set_internal_operation_mode(HVACOperationMode.STANDBY)
+            if processed:
```
marvin-w, line 263: why not just without all the processed stuff.

```suggestion
            if self.has_group_address(telegram.group_address):
```

marvin-w (approved):
Tested, works fine. Do you want to add tests for this? :-D

PR #421, opened 2020-09-23
create outgoing KNXIPFrame and CEMIFrame directly

farmio:
- initialize KNXIPBody objects directly, instead of creating empty objects
- create (outgoing) KNXIPFrame from KNXIPBody

- initialize CEMIFrame objects directly, instead of creating empty objects
- create (outgoing) CEMIFrame from Telegram

This saves a couple of function calls and is better readable imho.

### DIFF
--- a/xknx/io/connect.py
+++ b/xknx/io/connect.py
@@ -1,10 +1,10 @@
 """Abstraction to send ConnectRequest and wait for ConnectResponse."""
 from xknx.knxip import (
     HPAI,
+    ConnectRequest,
     ConnectRequestType,
     ConnectResponse,
     KNXIPFrame,
-    KNXIPServiceType,
 )
 
 from .request_response import RequestResponse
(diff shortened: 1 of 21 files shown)


farmio:
If you like it we could also make CEMIFrame initializeable (staticmethod that takes a telegram and some additional information).

xknx/io/connect.py
```diff
@@ -21,17 +21,18 @@ def __init__(self, xknx, udp_client):
         self.communication_channel = 0
         self.identifier = 0
 
-    def create_knxipframe(self):
+    def create_knxipframe(self) -> KNXIPFrame:
```
marvin-w, line 24: This reminds that I wanted to integrate a mypy testrunner...

marvin-w:
I like the idea!

Julius2342:
Yes, that makes sense!

xknx/io/request_response.py
```diff
@@ -41,7 +41,7 @@ async def start(self):
     async def send_request(self):
         """Build knxipframe (within derived class) and send via UDP."""
         knxipframe = self.create_knxipframe()
-        knxipframe.normalize()
+        # knxipframe.normalize()
```
marvin-w, line 44: Typo?

marvin-w (approved):
Have a look at the one possible too, otherwise nice :)

PR #424, opened 2020-09-24
Add mypy testrunner

marvin-w:
Yay!

We should add new modules to tox.ini one by one and fix the upcoming errors by hand or with monkeytype :-D.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,3 +31,21 @@ known_third_party = xknx
 known_first_party = homeassistant,test
 forced_separate = test
 combine_as_imports = true
+
+[mypy]
+python_version = 3.7
+ignore_errors = true
+follow_imports = silent
+ignore_missing_imports = true
+warn_incomplete_stub = true
+warn_redundant_casts = true
+warn_unused_configs = true
+
+# add the modules below once we add typing for them so that we fail the build in the future if someone changes something without updating the typings
+[mypy-xknx.telegram.*]
+strict = true
+ignore_errors = false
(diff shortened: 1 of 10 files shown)


setup.cfg
```diff
@@ -31,3 +31,21 @@ known_third_party = xknx
...
+python_version = 3.7
+ignore_errors = true
+follow_imports = silent
+ignore_missing_imports = true
+warn_incomplete_stub = true
+warn_redundant_casts = true
+warn_unused_configs = true
+
+# add the modules below once we add typing for them so that we fail the build in the future if someone changes something without updating the typings
+[mypy-xknx.telegram.*]
```
marvin-w, line 45: Add new modules here after typing

xknx/telegram/address.py
```diff
@@ -81,29 +80,29 @@ class PhysicalAddress(BaseAddress):
         r"^(?P<area>\d{1,2})\.(?P<main>\d{1,2})\.(?P<line>\d{1,3})$"
     )
 
-    def __init__(self, address):
+    def __init__(self, address: Union[object, str, int, tuple]) -> None:
```
marvin-w, line 83: We can't use the same Type as the current class unfortunetly :/

farmio: this is our excuse to drop 3.6
https://stackoverflow.com/a/49872353/4516732
https://www.python.org/dev/peps/pep-0563/

farmio (approved):
Nice!

PR #428, opened 2020-09-25
Refactor logging in xknx

marvin-w:
This commit aims to remove some of the tight coupling between XKNX and all our subpackages.
We now use a dedicated logger for every module.

Additionally, when we use `XKNX(log_to_file=True)` we log each and everything to a dedicated file
called xknx.log that will be kept for 7 days and will be rotated every night.

The default for this is False.

### DIFF
--- a/docs/xknx.md
+++ b/docs/xknx.md
@@ -28,7 +28,7 @@ xknx = XKNX(config='xknx.yaml',
             rate_limit=DEFAULT_RATE_LIMIT,
             multicast_group=DEFAULT_MCAST_GRP,
             multicast_port=DEFAULT_MCAST_PORT,
-            log_to_file=False)
+            log_directory=None)
 ```
 
 The constructor of the XKNX object takes several parameters:
@@ -45,7 +45,7 @@ The constructor of the XKNX object takes several parameters:
 * `rate_limit` in telegrams per second - can be used to limit the outgoing traffic to the KNX/IP interface. The default value is 20 packets per second.
 * `multicast_group` is the multicast IP address - can be used to override the default multicast address (`224.0.23.12`)
 * `multicast_port` is the multicast port - can be used to override the default multicast port (`3671`)
-* `log_to_file` is a boolean - when set to True we log to a dedicated file `xknx.log`
+* `log_directory` is the path to the log directory - when set to a valid directory we log to a dedicated file in this directory called `xknx.log`. The log files are rotated each night and will exist for 7 days. After that the oldest one will be deleted.
 
 # [](#header-2)Starting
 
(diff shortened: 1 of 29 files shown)


docs/xknx.md
```diff
@@ -44,6 +45,7 @@ The constructor of the XKNX object takes several parameters:
 * `rate_limit` in telegrams per second - can be used to limit the outgoing traffic to the KNX/IP interface. The default value is 20 packets per second.
 * `multicast_group` is the multicast IP address - can be used to override the default multicast address (`224.0.23.12`)
 * `multicast_port` is the multicast port - can be used to override the default multicast port (`3671`)
+* `log_to_file` is a boolean - when set to True we log to a dedicated file `xknx.log`
```
farmio, line 48: We should enable users to pass a directory where to write the loggiles at. 
Especially in Docker environments the Loggien could go lost.

xknx/core/state_updater.py
```diff
@@ -2,13 +2,16 @@
...
 from functools import partial
+import logging
 
 from xknx.exceptions import ConversionError
 from xknx.remote_value import RemoteValue
 from xknx.telegram import GroupAddress
 
 DEFAULT_UPDATE_INTERVAL = 60
 
+logger = logging.getLogger("xknx_log")
```
farmio, line 13: This should have an extra logger imho - it spams a lot.

xknx/devices/climate.py
```diff
@@ -15,6 +16,8 @@
 from .climate_mode import ClimateMode
 from .device import Device
 
+logger = logging.getLogger("xknx_log")
```
farmio, line 19: Shall we use an extra device-logger?

marvin-w: I thought about this too, yes sure. I also saw that the other 3 loggers are only really barely used.

docs/xknx.md
marvin-w: Yea good point. Should we have multiple logfiles or only one? It's only one right now with each type having there own column (%(name) is the logger name in the formatter)

xknx/core/state_updater.py
marvin-w: This wouldn't change anything. It would still get logged, always. In the logfile and in the usual stream logs.

farmio: We can then set individual log levels for each logger.

docs/xknx.md
farmio: I'd prefer one file so the rows are in timed order.

Julius2342:
A nice change :)

PR #432, opened 2020-09-26
change logger names to previous name

farmio:
also state_updater gets own logger


marvin-w (changes requested):
Please change https://github.com/XKNX/xknx/blob/master/xknx/xknx.py#L169 too - otherwise nice!

marvin-w (approved):
Looks fine now - thanks!!

Issue #238, opened 2019-09-20
Configuration via HA user interface


farmio:
👋 Do you want to improve the xknx home-assistant integration? Nice 😍!

Enable configuration via home-assistant ui by defining a config flow.

Let's define some contract details:

- [x] Migrate Binary Sensor automation schema to home assistant events. https://github.com/XKNX/xknx/pull/365
- [x] Use a config flow to set up the base connection to the gateway - https://developers.home-assistant.io/docs/en/config_entries_config_flow_handler.html
- [x] Use `GatewayScanner` to scan for gateways in the network (maybe tell them that "Answer search requests" need to be set to true in the ETS parameters of the gateway.
- [x] Enhance existing xknx.yaml syntax (it's pretty awful right now - breaking change...)
- [x] Write a migration script for home assistant users so that it's easier to handle the breaking change - sorry :100: 
- [x] Make sure config entries support unloading - then they can be reloaded from the UI (you can add new KNX devices in YAML and they will appear upon reloading the integration) - We can use `hass.config_entries.async_reload(config_entry.entry_id)` in HA for that
- [ ] Drop support for YAML based configuration in Home Assistant

marvin-w:
@farmio I personally don't feel like configuring each entity via the UI would provide a huge benefit for the user, as it's probably more painful to configure group addresses in the UI than in HA.

A very welcomed addition would nevertheless be to have the connection part in the config flow.

Optionally, we could think about completely dropping the YAML configuration inside home assistant and force the user to use our xknx.yaml.

When implemented correctly, we could setup some kind of reload check like:

* Check if the defined yaml has changed within the last x minutes
** If yes, drop all devices from `self.xknx.devices` and readd those from the YAML
*** TODO: Check if we would need to do additional checks or remove something
*** We probably need to restart the `StateUpdater` as well

What do you think? I personally don't like the way our xknx.yaml is structured... But I'd also like to avoid another breaking change... so I'm unsure.

marvin-w:
Btw. I tested the GatewayScanner. At least on my network it did not find my router. KNX IP BAOS 777

TinusK:
I think it is a Great idea!

farmio:
>  I personally don't feel like configuring each entity via the UI would provide a huge benefit for the user, as it's probably more painful to configure group addresses in the UI than in a yaml file where you can just copy paste an entire block.

[...]

PR #436, opened 2020-09-28
Adds support for unique id

marvin-w:
This is still a draft as I'm focusing on something else for now. Some tests are still failing, but the idea should become clear.

### DIFF
--- a/xknx/devices/binary_sensor.py
+++ b/xknx/devices/binary_sensor.py
@@ -28,7 +28,7 @@ class BinarySensor(Device):
         self,
         xknx,
         name,
-        group_address_state=None,
+        group_address_state,
         sync_state=True,
         ignore_internal_state=False,
         device_class=None,
@@ -68,6 +68,11 @@ class BinarySensor(Device):
     def _iter_remote_values(self):
         """Iterate the devices RemoteValue classes."""
         yield self.remote_value
+
+    @property
+    def unique_id(self):
+        """Return unique id for this device."""
+        return f"binary_sensor_{hash(self.remote_value.group_address_state)}"
 
     def __del__(self):
         """Destructor. Cleaning up if this was not done before."""
--- a/xknx/devices/switch.py
+++ b/xknx/devices/switch.py
@@ -22,7 +22,7 @@ class Switch(Device):
         self,
         xknx,
         name,
-        group_address=None,
+        group_address,
         group_address_state=None,
         device_updated_cb=None,
     ):
@@ -41,6 +41,11 @@ class Switch(Device):
     def _iter_remote_values(self):
         """Iterate the devices RemoteValue classes."""
         yield self.switch
+
+    @property
+    def unique_id(self):
+        """Return unique id for this device."""
+        return f"switch_{hash(self.switch.group_address)}"
 
     @classmethod
     def from_config(cls, xknx, name, config):
--- a/test/devices_tests/cover_test.py
+++ b/test/devices_tests/cover_test.py
@@ -775,7 +775,7 @@ class TestCover(unittest.TestCase):
             return None
 
         xknx = XKNX()
-        cover = Cover(xknx, "TestCover")
+        cover = Cover(xknx, "TestCover", group_address_long="2/4/5")
         with patch("xknx.devices.Cover.set_up") as mock:
             mock.return_value = asyncio.ensure_future(async_none())
             self.loop.run_until_complete(cover.do("up"))
@@ -800,7 +800,7 @@ class TestCover(unittest.TestCase):
     def test_wrong_do(self):
         """Test wrong do command."""
         xknx = XKNX()
-        cover = Cover(xknx, "TestCover")
+        cover = Cover(xknx, "TestCover", group_address_long="2/3/4")
         with patch("logging.Logger.warning") as mock_warn:
             self.loop.run_until_complete(cover.do("execute"))
             self.assertEqual(xknx.telegrams.qsize(), 0)
--- a/test/devices_tests/light_test.py
+++ b/test/devices_tests/light_test.py
@@ -139,6 +139,7 @@ class TestLight(unittest.TestCase):
         light = Light(
             xknx,
             name="TestLight",
+            group_address_switch="1/3/5",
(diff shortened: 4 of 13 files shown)


xknx/devices/binary_sensor.py
```diff
@@ -69,6 +69,11 @@ def _iter_remote_values(self):
         """Iterate the devices RemoteValue classes."""
         yield self.remote_value
 
+    @property
+    def unique_id(self):
+        """Return unique id for this device."""
+        return f"binary_sensor_{hash(self.remote_value.group_address_state)}"
```
marvin-w, line 75: This is all still untested - not sure what `hash(GroupAddress)` will return and if it can be used.

elupus: Binary sensor prefix is not needed. Unique IDs are scoped by platform

marvin-w: Ah perfect! Thanks.

elupus: Also it's just a string, so no need to hash it all? Or  du we have a tuple of its here?

marvin-w: Well it would look more like an id with a hash - but yea doesn't really matter.

xknx/devices/switch.py
```diff
@@ -42,6 +42,11 @@ def _iter_remote_values(self):
         """Iterate the devices RemoteValue classes."""
         yield self.switch
 
+    @property
+    def unique_id(self):
+        """Return unique id for this device."""
+        return f"switch_{hash(self.switch.group_address)}"
```
elupus, line 48: Is there I understood or good reason not to use state here? I mean all the other use state.

xknx/devices/binary_sensor.py
elupus: I think it's better it is an understandable value.

xknx/devices/switch.py
marvin-w: Not all switches have a state_address. But all switches have a switch_address. (Check the schema of HA - the switch GA is the only required parameter here)

elupus: Right. I just saw comment on discord.

I wonder if we should prepare for a installation prefix to the unique I'd:s. Might not be worth it.

marvin-w: Well right now you cannot have more than one gateway (and therefore installation) in HA if I recall correctly.

xknx/devices/binary_sensor.py
farmio: hash(GroupAddress) returns the raw GA value (int)
str returns "x/y/z" depending on used address Schema 
and __repr__ returns 'GroupAddress("x/y/z")' also depending on Schema.

marvin-w: I guess str would be fine then, thanks for checking ;).

xknx/devices/switch.py
elupus: Right, but we could solve that quite nicely with config entries for the gateway configuration. Obviously need some more work, but it would move us closer to best practice..

Ps.. it would also allow reload of the integration without restart when you add/remove entities in yaml.

marvin-w: https://github.com/XKNX/xknx/issues/238

marvin-w: PRs are always welcome :dancers:

elupus: Will see if I get the itch. I tend to spread my time too thin on the integrations I'm involved in anyway, so why not a other :)

marvin-w: Haha :-D. I saw you have been dealing with config flows in the past, I might come back to you if I need some help on that - or just ask on HA discord. :-D

elupus: Absolutely, it's usually fun untill you get an angry hoard of users complaining about having to adjust to a breaking change. :)

marvin-w: Haha yeah. I got the hang of that already xD.

xknx/devices/binary_sensor.py
Julius2342: I like the idea. But i see problem, that it is not guaranteed, that the unique id stays as is forever. (e.g. python one day might change the `__hash__()` of int. I would therfore use `md5` or any other guaranteed reproducible hash algorithm.

Am i right, it is not possible to add this on `device`-Level bc not all devices have a `group_address_state` ?

marvin-w: We will just use the string of the GA, so no hash or md5 needed :). `Device` doesn't know anything about the `RemoteValues` of the real implemented devices.

Julius2342: And, are we sure that there are installations out there, which use the same `group_address_state` for two devices? Should we guarantee that it is unique within one setup?

Julius2342: And, last but not least. Assuming you have a light with a ga `1/2/3`. And you change this to a switch with the same address. What would be the expected behavior? The same or a different address?

marvin-w: > And, are we sure that there are installations out there, which use the same group_address_state for two devices? Should we guarantee that it is unique within one setup?

There can be devices that use the same state GA but the unique_id is scoped to the platform - and I highly doubt that there are too binary sensors or anything with two equal state GAs

> And, last but not least. Assuming you have a light with a ga 1/2/3. And you change this to a switch with the same address. What would be the expected behavior? The same or a different address?

Since switch and light are operating on a different platform they will have a different address in HA.

elupus: they will get same unique id scoped to platform. But since they are different platform, they will not retain any friendly name set from GUI, you have to re-set it. But switching would anyway have changed the prefix of the entity_id from switch.* to light.*

There is possibility that people may have used the same control address for a light switch with dimming and a light switch without dimming. This would then not be allowed. Also do we support light switches without on/off state, just a brightness level? It's somewhat tricky to decide which address is the base for unique id, the brightness level or the on/off command?

marvin-w:
Will redo this.

PR #437, opened 2020-09-28
Adds support for new config schema

marvin-w:
This PR will add a new config version to XKNX that can be used to configure it on a deeper level and it also looks cleaner and has built in validation and data type conversion thanks to voloptuous.


home-assistant-plugin/custom_components/xknx/schema.py
```diff
@@ -327,7 +327,9 @@ class SceneSchema:
         {
             vol.Optional(CONF_NAME, default=DEFAULT_NAME): cv.string,
             vol.Required(CONF_ADDRESS): cv.string,
-            vol.Required(CONF_SCENE_NUMBER): cv.positive_int,
+            vol.Required(CONF_SCENE_NUMBER): vol.All(
+                vol.Coerce(int), vol.Range(min=1, max=64)
```
marvin-w, line 331: Remove

marvin-w:
@farmio I'd say we merge this before I continue my work here. It's already _really_ big so I think we should split this in pieces. Users will not notice any difference as we do not yet use the new schema for anything.

xknx/config/schema.py
```diff
@@ -0,0 +1,606 @@
...
+    CONF_DATETIME = "datetime"
+    CONF_NOTIFICATION = "notification"
+    CONF_SCENE = "scene"
+    CONF_EXPOSE = "expose_sensor"
+    CONF_SENSOR = "sensor"
+
+    SCHEMA = vol.Schema(
+        {
+            vol.Optional(CONF_RATE_LIMIT, default=20): vol.All(
+                vol.Coerce(int), vol.Range(min=1, max=100)
```
farmio, line 577:
```suggestion
            vol.Optional(CONF_RATE_LIMIT, default=20): vol.All(
                vol.Coerce(int), vol.Range(min=0, max=100)
```
we can use 0 to disable rate limiter.

xknx/config/schema.py
```diff
@@ -0,0 +1,606 @@
...
+            ),
+            vol.Optional(CONF_VERSION, default=2): string,
+            vol.Optional(CONF_OWN_ADDRESS): string,
```
farmio, line 580: we could use 
https://github.com/XKNX/xknx/blob/552852752ed87963deb192e045277db5e59ee97b/xknx/telegram/address.py#L79
just like for group addresses

xknx/config/schema.py
```diff
@@ -0,0 +1,606 @@
...
+            vol.Optional(CONF_LOG_DIRECTORY): isdir,
+            vol.Inclusive(CONF_FIRE_EVENT, "fire_ev"): boolean,
+            vol.Inclusive(CONF_FIRE_EVENT_FILTER, "fire_ev"): vol.All(
+                ensure_list, [string]
+            ),
+            vol.Optional(CONF_MULTICAST_GROUP, default=DEFAULT_MCAST_GRP): string,
```
farmio, line 586: Is there an easy way to test for an IPv4 address?

farmio (commented):
puh this is huge... 😅

xknx/config/schema.py
marvin-w: Not in voloptuous. HA doesn't have this either. Also, not all IP addresses are valid multicast IPs.

PR #441, opened 2020-09-30
Add support for listening group addresses to RemoteValue

marvin-w:
Based on #321.

The only difference is that we use a dedicated `listening_group_addresses` setting instead of mixing it with the main GA.

The integration into the devices is not yet part of this PR but will be possible once the new config schema is ready.  (#437)

### DIFF
--- a/xknx/remote_value/remote_value.py
+++ b/xknx/remote_value/remote_value.py
@@ -66,7 +66,7 @@ class RemoteValue:
         return bool(
             self.group_address_state
             or self.group_address
-            or len(self.listening_group_addresses)
+            or self.listening_group_addresses
         )
 
     @property
@@ -240,5 +240,5 @@ class RemoteValue:
     def get_listening_group_addresses(passive_group_addresses: List[str]) -> List:
         """Obtain passive state group addresses."""
         if passive_group_addresses is None:
-            passive_group_addresses = []
+            return []
         return [GroupAddress(ga) for ga in passive_group_addresses]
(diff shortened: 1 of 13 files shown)


xknx/remote_value/remote_value.py
```diff
@@ -27,10 +28,14 @@ def __init__(
         device_name=None,
         feature_name=None,
         after_update_cb=None,
+        listening_group_addresses: List[str] = None,
```
farmio, line 31: in your config PR they are called passive_group_addresses if I recall correctly - which I personally prefer. it implies no action. "listening" is more confusable with "state" imho.

xknx/remote_value/remote_value.py
```diff
@@ -58,7 +63,11 @@ def __del__(self):
     @property
     def initialized(self):
         """Evaluate if remote value is initialized with group address."""
-        return bool(self.group_address_state or self.group_address)
+        return bool(
+            self.group_address_state
+            or self.group_address
+            or len(self.listening_group_addresses)
```
farmio, line 69:
```suggestion
            or self.listening_group_addresses
```
an empty list is falsey

xknx/remote_value/remote_value.py
```diff
@@ -219,3 +235,10 @@ def __eq__(self, other):
             if key not in self.__dict__:
                 return False
         return True
+
+    @staticmethod
+    def get_listening_group_addresses(passive_group_addresses: List[str]) -> List:
+        """Obtain passive state group addresses."""
+        if passive_group_addresses is None:
+            passive_group_addresses = []
```
farmio, line 243:
```suggestion
            return []
```

farmio (commented):
Cool!

marvin-w:
Thanks for the review :) Should be fine now.

PR #443, opened 2020-10-01
Tunnel Refactoring

### LINKED ISSUE #89: Catch network errors when setting up tunnel
Setup
------

```
knx:
  tunneling:
    host: 192.168.0.132  # <- non existing IP
    local_ip: 192.168.0.134
```

Current behaviour
------------------

Cant assign requested network address:

```
    sock.bind(local_address)
OSError: [Errno 49] Can't assign requested address
```

Expected behaviour:
----------------------

XKNXException is thrown.

farmio:
- Tunnel connections process DisconnectRequest now and closes/reconnects the tunnel when the other side closes gracefully
- XKNX.connected Event can be used in future to await for a working connection or stop/relaunch tasks if the connection is lost (I'd like to use this eg. for StateUpdater and DateTime broadcast)
- Lower heartbeat rate from every 15sec to every 70 sec and raise ConnectionstateRequest timeout from 1 to 10sec (3/8/1 KNXip Overview §5.8 Timeout Constants). There are 3 retrys so 90 + 10 (initial connectionstate request) + 3x 10 (retrys) = 80 - 10 seconds spare = 80 . We will probably not need to rely on the heartbeat to detect lost connections anymore because we process DisconnectRequests now.
- clean up Tunnel class 

closes #89 

TODO:

- [ ] on Error shut down the whole interface (and UDPClient) and create a new one from KNXIPInterface so the connection_config can be read again. This would enable to connect to (or from) a different IP, KNX/IP Interface or even connection type when one interface fails.

- [ ] CONNECT_REQUEST_TIMEOUT should be 10 seconds instead of 1 according to the specs

- [ ] call Tunnel._tunnel_lost() from UDPClient error_received ( callback / raise / contextmanager ...) - same for Routing

- [ ] recreate log messages and errors 

- [ ] write tests!

Known issues:

- [ ] When using daemon_mode and auto_reconnect=False (which is a very unlikely scenario imho) the XKNX context manager does not return on a lost connection. It hangs waiting for SIGINT. On SIGINT it tries to call Tunnel.stop() which fails.

- [ ] This probably needs a lot of testing under rough conditions 😬


xknx/io/tunnel.py
```diff
@@ -185,112 +231,96 @@ async def _send_telegram_impl(self, telegram):
...
-    async def connectionstate(self):
-        """Return state of tunnel. True if tunnel is in good shape."""
-        conn_state = ConnectionState(
+    ####################
+    #
+    # INCOMING REQUESTS
+    #
+    ####################
+
+    def _reqest_received(self, knxipframe, _udp_client):
```
marvin-w, line 246: s/_reqest/_request/g

xknx/xknx.py
```diff
@@ -53,6 +53,7 @@ def __init__(
         self.state_updater = StateUpdater(self)
         self.knxip_interface = None
         self.started = asyncio.Event()
+        self.connected = asyncio.Event()
```
marvin-w, line 56: I like this. We can then probably remove the one in the HA plugin (it also has a connected flag in the `KNXModule`) that works different

xknx/io/tunnel.py
```diff
@@ -71,77 +74,117 @@ def init_udp_client(self):
...
-            sequence_counter=sequence_counter,
-        )
-        self.udp_client.send(KNXIPFrame.init_from_body(ack))
+    ####################
+    #
+    # CONNECT DISCONNECT
+    #
+    ####################
 
     async def start(self):
```
marvin-w, line 87: I like the idea. Since you now have dedicated start and stop functions it would probably make sense to add an async context manager to it and adjust the examples as well?

examples/example_tunnel.py
```diff
@@ -34,16 +34,14 @@ async def main():
         gateway_port=gateway.port,
     )
 
-    await tunnel.connect_udp()
-    await tunnel.connect()
+    await tunnel.start()
```
marvin-w, line 37: `async with tunnel:` <3

marvin-w (changes requested):
I like the idea, however there are still some open points and I also haven't tested this yet.

It would make sense to add a dedicated context manager.

Also, from experience I can safely say that refactoring a _mostly untested_ class is very highly going to cause integration issues.

xknx/io/tunnel.py
farmio: I sure like the idea. But for now I aim for a stable and reliable connection. Since this is not very likely to be used directly I think syntactic sugar can be added later.

farmio:
I think it works fine now. Maybe someone else can test it for a while?

One thing I still want to change here is:
> on Error shut down the whole interface (and UDPClient) and create a new one from KNXIPInterface so the connection_config can be read again. This would enable to connect to a different interface or even connection type when one interface fails. Also when for some reason the interface gets a new IP address it would be able to connect to it when connection_config is set appropriately.

But this should maybe be done in an extra PR.

xknx/io/tunnel.py
```diff
@@ -71,77 +74,126 @@ def init_udp_client(self):
...
-        """Start tunneling."""
-        await self.connect_udp()
-        await self.connect()
+        """Start tunneling. Returns True on success."""
+        try:
+            await self.udp_client.connect()
+            await self._connect_request()
+        except (OSError, CommunicationError) as ex:
```
farmio, line 92: closes #89

PR #445, opened 2020-10-02
disable counter when context_timeout is 0

farmio:
- enable use of ignore_internal_state without counter function
- don't run callbacks twice (2nd from finishing context_timeout immediately)

> cyberjunky on HA discord:
> I have windows sensors sending their stats every 10 minutes, the state doesn't change but the attribute counter: goes from 1 to 0 in 2 packets, so it triggers automation... and i have to code around it, ending up double automations (for on and off) it should be optional...

### DIFF
--- a/docs/migration_ha_0116.md
+++ b/docs/migration_ha_0116.md
@@ -16,7 +16,7 @@ We've changed the default behavior of the `ignore_internal_state` attribute. It 
 If you encounter issues with your current automations please set it to `False` again. We've analysed the current state
 and it appears that most of the people using it have it set to `True`.
 
-The binary sensor now has an additional `context_timeout` attribute which can be used in order to define the time period
+The binary sensor now has an additional `context_timeout` attribute which allows to define a time period
 in which your clicks should be counted towards the current context (i.e. incrementing the counter variable that you can
 use in your automations).
 
@@ -54,6 +54,7 @@ Your new binary sensor will now look like:
 
 - name: cover_abstell
   state_address: "2/0/33"
+  context_timeout: 1.0
 
 ```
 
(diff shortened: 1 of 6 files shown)


farmio:
I would even say 0 should be default for context_timeout. For most BinarySensors double-tap function is not needed. So we should avoid extra lag it adds and possible confusions from double callbacks.

test/devices_tests/binary_sensor_test.py
```diff
@@ -197,9 +197,42 @@ def test_is_off(self):
...
+    def test_process_callback_ignore_internal_state(self):
+        """Test after_update_callback after state of switch was changed."""
+        # pylint: disable=protected-access
+        xknx = XKNX()
+        switch = BinarySensor(
+            xknx,
+            "TestInput",
+            group_address_state="1/2/3",
+            ignore_internal_state=True,
+            context_timeout=0.01,
```
marvin-w, line 234: Haha lovely :-D

marvin-w (approved):
I'm fine with that :). Looks reasonable. Feel free to change the default as well.

docs/migration_ha_0116.md
```diff
@@ -16,7 +16,7 @@ We've changed the default behavior of the `ignore_internal_state` attribute. It
 If you encounter issues with your current automations please set it to `False` again. We've analysed the current state
```
marvin-w, line 16: We should delete this file. HA doesn't like external links.

farmio: I have linked to this in the forums. It's easier to find than the knx breaking changes section of HAs blog (wich has no anchor links 🤦)

marvin-w: Then create a new file for 0.117.0 instead of changing this one for which the beta has already been released?

PR #447, opened 2020-10-04
Fix parsing floats

jonppe:
Small fix for parsing floating point DPT values and a test.

### DIFF
--- a/test/dpt_tests/dpt_test.py
+++ b/test/dpt_tests/dpt_test.py
@@ -139,3 +139,6 @@ class TestDPTBase(unittest.TestCase):
         dpt6 = DPTBase.parse_transcoder("9.001")
         self.assertEqual(dpt4, dpt5)
         self.assertEqual(dpt5, dpt6)
+        dpt7 = DPTBase.parse_transcoder("active_energy")
+        dpt8 = DPTBase.parse_transcoder(13.010)
+        self.assertEqual(dpt7, dpt8, "parsing float failed")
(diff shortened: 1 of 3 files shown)


farmio:
Thank you for your contribution!

There are also DPTs with 4 digit subtype. I think we currently don't have any implemented, but it would be nice if you had a solution that can work for zero-ending 3 digit and 4 digit subtypes.
If not, I'd be ok to merge this and see again when we implement these subtypes.

jonppe:
For 4 digit subtypes, I might need to understand the requirements a bit better. 
For example, is it possible that both number like 1.2340 and 1.234 are used? 
I guess that would make it impossible to use floats without losing some information.

But if this is not an issue, then a combination of the original and new solution might work:
precision = max(len(str(value_type).split(".")[1]), 3)
main, sub = map(int, f"{value_type:.{precision}f}".split("."))

But that is quite complex already. I wish python had some "minimum precision" formatter but I didn't find that. 
For now, I'd just merge and then wait for the 4 digit subtypes first to be implemented first.

farmio:
Both are used (eg. 12.1200) https://www.knx.org/wAssets/docs/downloads/Certification/Interworking-Datapoint-types/03_07_02-Datapoint-Types-v02.01.02-AS.pdf

marvin-w:
Two things here:

1. DPT 12.1200 is already implemented and can be used for testing :) - see `DPTVolumeLiquidLitre`
2. Neither of the solutions will work because as @jonppe correctly said we will lose some information for floats.

```
13.010 => 13.01 => only works because of the precision of 3 adding the missing 0.
12.1200 => 12.12 => does not work
```

After all, those should be strings and not floats. I'm fine with merging it the way it is now. I also tried your solution locally but it will not work for 12.1200 because we don't know if 12.120 or 12.1200 is meant and we also can't (and do not want to) gather that information. I'll make sure that you can not specify floats in the new config schema as this will break for certain edge cases.

marvin-w (approved):

PR #451, opened 2020-10-06
Support Python 3.9

marvin-w:
Adds CI support for python 3.9


marvin-w:
Still waiting for Travis CI support. (No docker image available yet)

marvin-w:
Waiting for pyenv release that supports python 3.9.

farmio:
I think we could try this again. https://docs.travis-ci.com/user/languages/python/#python-versions indicates 3.9 is supported now.
Maybe we should also remove `dist: xenial` so any dist supporting the python version we need can do the job? I think this could be a reason why we have to wait so long for CI to start recently (3.8.5 on xenial is very specific).

marvin-w:
I removed the dist. Let's wait for the build.

marvin-w:
Well, I guess removing dist didn't help at all. Sad.

PR #454, opened 2020-10-07
Allow resetting switches after x seconds

### LINKED ISSUE #438: Switch: add auto-off function
Sometimes it is required to s always send a "OFF" right after sending a "ON" eg. for door openers. 
I suggest to add a `reset_after` option to Switch, just like BinarySensor, to send the "OFF" telegram after a timer is finished.

marvin-w:
## Description

Allows switches to be turned off after x seconds. Similar as to how binary sensors do it except that switches send the OFF telegram to the bus.

Note: Tests currently fail - not sure why..

Fixes #438 

## Type of change

# Checklist:

- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,24 +2,22 @@
   You are awesome! Thanks for contributing to our project!
   Please, DO NOT DELETE ANY TEXT from this template!.
 -->
-
+## Description
 <!--
 Please include a summary of the change and which issue is fixed.
 Please also include relevant motivation and context.
 List any dependencies that are required for this change.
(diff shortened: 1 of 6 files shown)


.github/pull_request_template.md
```diff
@@ -29,6 +27,8 @@ that your change is to big and it is suggested to split it into several smaller
 # Checklist:
```
marvin-w, line 27: Use second level.

PR #364, opened 2020-09-01
Deprioritize state updater

farmio:
Wait until there is nothing else to send to the bus before sending GroupValueRead requests from StatUpdater.

please "Squash and merge" this if you like it 🤦‍♂️

PR #457, opened 2020-10-11
StateUpdater: only request one group address at once

farmio:
## Description

- Request one group address at a time. StateUpdater now waits for a response (or timeout) and only then triggers the next GroupValueRead request.
- set timeout for GroupValueRead requests back to 1 sec (from 2)
- remove some debug logs

This PR does what #364 was intended for 🙃

## Type of change

## Checklist:

### DIFF
--- a/xknx/core/state_updater.py
+++ b/xknx/core/state_updater.py
@@ -1,14 +1,12 @@
 """Module for keeping the value of a RemoteValue from KNX bus up to date."""
 import asyncio
 from enum import Enum
-from functools import partial
 import logging
 
-from xknx.exceptions import ConversionError
 from xknx.remote_value import RemoteValue
-from xknx.telegram import GroupAddress
 
 DEFAULT_UPDATE_INTERVAL = 60
+MAX_UPDATE_INTERVAL = 1440
 
 logger = logging.getLogger("xknx.state_updater")
 
@@ -21,6 +19,7 @@ class StateUpdater:
         self.xknx = xknx
         self.started = False
         self._workers = {}
+        self._one_by_one = asyncio.Lock()
 
     def register_remote_value(self, remote_value: RemoteValue, tracker_options=True):
         """Register a RemoteValue to initialize its state and/or track for expiration."""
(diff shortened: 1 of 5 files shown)


xknx/core/state_updater.py
```diff
@@ -66,24 +65,41 @@ def parse_tracker_options(tracker_options):
...
+                )
+                update_interval = MAX_UPDATE_INTERVAL
             return (tracker_type, update_interval)
 
-        tracker_type, update_inteval = parse_tracker_options(tracker_options)
+        async def read_state_mutex():
+            """Schedule to read the state from the KNX bus - one at a time."""
+            async with self._one_by_one:
+                # wait until there is nothing else to send to the bus
+                await self.xknx.telegram_queue.outgoing_queue.join()
```
farmio, line 82: It would probably work without this too and we could remove the xknx reference in the class completely. But I think it is ok to be extra safe for now 😉

xknx/core/state_updater.py
```diff
@@ -21,6 +19,7 @@ def __init__(self, xknx):
         self.xknx = xknx
         self.started = False
         self._workers = {}
+        self._one_by_one = asyncio.Lock()
```
marvin-w, line 22: Maybe renaming this to something like `self._mutex` or `self._lock` would be better readable.

farmio: Sure. What about `self._mutex_lock`?

marvin-w: It should be either lock or mutex. A mutex is basically a system wide lock (over multiple processes). I guess in python lock would be the correct wording

marvin-w: I'll test this feature branch and give feedback after :+1:

farmio: I liked the line "async with one_by_one"    😉

marvin-w: Yea also true. Can we use an `asyncio.Semaphore(3)` to update 3 states at the same time? I understand we do not want to spam the bus with all the state updates at once but doing one by one looks kind of unneccessary for me.

xknx/core/state_updater.py
marvin-w: Yea that's fine. I'd still prefer to update more than one state at a time to not have the user wait extra long. See my other comment.

marvin-w (approved):
Tested successfully with my local test HA installation. I'm fine with leaving it like this for now, but maybe we can enhance the logic to update multiple states at the same time as written in the review.

xknx/core/state_updater.py
farmio: Sure, should work. I'll try it later.

PR #460, opened 2020-10-11
Refactor climate modes in operation mode and controller mode.

marvin-w:
## Description

Fixes a bug for binary operation modes where AWAY would always be set initially no matter what value was sent.

Added "controller_modes" parameter to HA configuration.

Fixes https://github.com/home-assistant/core/issues/40768

## Type of change

## Checklist:

### DIFF
--- a/home-assistant-plugin/custom_components/xknx/climate.py
+++ b/home-assistant-plugin/custom_components/xknx/climate.py
@@ -14,10 +14,10 @@ from homeassistant.components.climate.const import (
 )
 from homeassistant.const import ATTR_TEMPERATURE, TEMP_CELSIUS
 
-from .const import DOMAIN, OPERATION_MODES, PRESET_MODES
+from .const import CONTROLLER_MODES, DOMAIN, PRESET_MODES
 from .knx_entity import KnxEntity
 
-OPERATION_MODES_INV = dict(reversed(item) for item in OPERATION_MODES.items())
+CONTROLLER_MODES_INV = dict(reversed(item) for item in CONTROLLER_MODES.items())
 PRESET_MODES_INV = dict(reversed(item) for item in PRESET_MODES.items())
 
 
@@ -92,27 +92,27 @@ class KNXClimate(KnxEntity, ClimateEntity):
         """Return current operation ie. heat, cool, idle."""
         if self._device.supports_on_off and not self._device.is_on:
             return HVAC_MODE_OFF
-        if self._device.mode.supports_operation_mode:
-            return OPERATION_MODES.get(
-                self._device.mode.operation_mode.value, HVAC_MODE_HEAT
+        if self._device.mode.supports_controller_mode:
+            return CONTROLLER_MODES.get(
+                self._device.mode.controller_mode.value, HVAC_MODE_HEAT
             )
         # default to "heat"
         return HVAC_MODE_HEAT
 
     @property
     def hvac_modes(self) -> Optional[List[str]]:
-        """Return the list of available operation modes."""
-        _operations = [
-            OPERATION_MODES.get(operation_mode.value)
-            for operation_mode in self._device.mode.operation_modes
+        """Return the list of available operation/controller modes."""
+        _controller_modes = [
+            CONTROLLER_MODES.get(controller_mode.value)
+            for controller_mode in self._device.mode.controller_modes
         ]
 
         if self._device.supports_on_off:
-            if not _operations:
-                _operations.append(HVAC_MODE_HEAT)
-            _operations.append(HVAC_MODE_OFF)
+            if not _controller_modes:
+                _controller_modes.append(HVAC_MODE_HEAT)
+            _controller_modes.append(HVAC_MODE_OFF)
 
-        _modes = list(set(filter(None, _operations)))
+        _modes = list(set(filter(None, _controller_modes)))
         # default to ["heat"]
         return _modes if _modes else [HVAC_MODE_HEAT]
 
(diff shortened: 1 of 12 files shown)


xknx/devices/climate_mode.py
```diff
@@ -218,55 +253,108 @@ async def _set_internal_operation_mode(self, operation_mode):
...
     def operation_modes(self):
         """Return all configured operation modes."""
         if not self.supports_operation_mode:
             return []
-        return self.operation_modes_
+        return self._operation_modes
+
+    @property
+    def controller_modes(self):
+        """Return all configures controller modes."""
```
farmio, line 305:
```suggestion
        """Return all configured controller modes."""
```

xknx/devices/climate_mode.py
```diff
@@ -218,55 +253,108 @@ async def _set_internal_operation_mode(self, operation_mode):
...
         """Gather operation modes from RemoteValues."""
         operation_modes = []
-        for rv in self._iter_remote_values():
+        for rv in self._iter_binary_remote_values():
+            if rv.writable:
+                operation_modes.extend(rv.supported_operation_modes())
+
+        for rv in self._iter_normal_operation_modes():
+            if rv.writable:
+                operation_modes.extend(rv.supported_operation_modes())
```
farmio, line 319:
```suggestion
        for rv in chain(
             self._iter_normal_operation_modes(), self._iter_binary_remote_values()
         ):
            if rv.writable:
                operation_modes.extend(rv.supported_operation_modes())
```

xknx/devices/climate_mode.py
```diff
@@ -218,55 +253,108 @@ async def _set_internal_operation_mode(self, operation_mode):
...
                 if await rv.process(telegram):
-                    # don't set when binary climate mode rv is False
                     if rv.value:
                         await self._set_internal_operation_mode(rv.value)
-                        return
-                    processed = True
-            # if no operation mode has been set and all binary operation modes are False
-            if processed:
-                await self._set_internal_operation_mode(HVACOperationMode.STANDBY)
+
```
farmio, line 341:
```suggestion
                    return
```
When a telegram is processed we can stop iterating - a GA should not be used in multiple RemoteValues here in any way.

xknx/devices/climate_mode.py
```diff
@@ -218,55 +253,108 @@ async def _set_internal_operation_mode(self, operation_mode):
...
         # remove duplicates
         return list(set(operation_modes))
 
     async def process_group_write(self, telegram):
         """Process incoming and outgoing GROUP WRITE telegram."""
         if self.supports_operation_mode:
-            processed = False
             for rv in self._iter_remote_values():
                 if await rv.process(telegram):
-                    # don't set when binary climate mode rv is False
```
farmio, line 257: I find this comment quite useful. It explains why we need the `if rv.value`here, but not for controller_mode. Could also say 
`# ignore inactive RemoteValueBinaryOperationMode`

xknx/devices/climate_mode.py
```diff
@@ -218,55 +253,108 @@ async def _set_internal_operation_mode(self, operation_mode):
...
-                    processed = True
-            # if no operation mode has been set and all binary operation modes are False
-            if processed:
-                await self._set_internal_operation_mode(HVACOperationMode.STANDBY)
+
+        if self.supports_controller_mode:
+            for rv in self._iter_controller_remote_values():
+                if await rv.process(telegram):
+                    if rv.value:
+                        await self._set_internal_controller_mode(rv.value)
```
farmio, line 346:
```suggestion
                    await self._set_internal_controller_mode(rv.value)
                    return
```
for `return` same as above. 
rv.value will always be True here as RemoteValueBinaryHeatCool returns an operation_mode with payload 1 and 0 RemoteValueClimateMode also. (This is what the comment above meant ;)

xknx/devices/climate_mode.py
```diff
@@ -218,55 +253,108 @@ async def _set_internal_operation_mode(self, operation_mode):
...
         if self.supports_operation_mode:
-            for rv in self._iter_remote_values():
+            for rv in chain(
+                self._iter_normal_operation_modes(), self._iter_binary_remote_values()
+            ):
+                await rv.read_state(wait_for_result=wait_for_result)
+
+        if self.supports_controller_mode:
+            for rv in self._iter_controller_remote_values():
                 await rv.read_state(wait_for_result=wait_for_result)
```
farmio, line 358: supports_operation_mode and supports_controller_mode checks for .initialized so I think it would be save to just iterate through all RVs - rv.read_state checks for .readable first.

We could then even remove the whole sync() function to use the super()-implementation from RemoteValue.

farmio (commented):
nice!

home-assistant-plugin/custom_components/xknx/climate.py
```diff
@@ -92,27 +92,27 @@ def hvac_mode(self) -> Optional[str]:
         """Return current operation ie. heat, cool, idle."""
         if self._device.supports_on_off and not self._device.is_on:
```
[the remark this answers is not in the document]
marvin-w: I'd leave it there for now. But I see your point.

xknx/devices/climate_mode.py
```diff
@@ -125,30 +129,43 @@ def __init__(
...
-            or group_address_heat_cool_state is not None
+                    self._operation_modes.append(mode)
+
+        self._controller_modes = []
+        if controller_modes is None:
+            self._controller_modes = self.gather_controller_modes()
+        else:
+            for mode in controller_modes:
+                if isinstance(mode, str):
+                    self._controller_modes.append(HVACOperationMode(mode))
```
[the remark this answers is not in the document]
marvin-w: I agree that it's not nice and we should change it, but since all DPTs use `HVACOperationMode` I believe we should not change this in the scope of this PR. I'll create follow up issue for this.

PR #472, opened 2020-10-20
Invertable Switch and BinarySensor

### LINKED ISSUE #442: Switch: Add invert function
Sometimes it makes sense to invert a switch. So it would make sense to add an `invert` flag to Switch and pass it to the RemoteValue (that already supports this).

farmio:
## Description

- BinarySensor: added option to invert payloads
- Switch: added option to invert payloads

Bugfix: 
- HA Switch entity: keep state without state_address. Previously `after_update` was called before the internal state of a Switch() was changed. So HA would read the old state, not the updated one. Moving the call to `after_update` from the RemoteValue to Switch.process_group_write() should fix this.

closes #442 

## Type of change

## Checklist:

### DIFF
--- a/xknx/devices/switch.py
+++ b/xknx/devices/switch.py
@@ -8,6 +8,7 @@ It provides functionality for
 """
 import asyncio
 import logging
+from typing import Optional
 
 from xknx.remote_value import RemoteValueSwitch
 
@@ -22,10 +23,11 @@ class Switch(Device):
     def __init__(
         self,
         xknx,
-        name,
-        reset_after=None,
+        name: str,
         group_address=None,
         group_address_state=None,
+        invert: Optional[bool] = False,
+        reset_after: Optional[float] = None,
         device_updated_cb=None,
     ):
         """Initialize Switch class."""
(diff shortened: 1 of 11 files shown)


xknx/remote_value/remote_value_switch.py
```diff
@@ -38,7 +38,7 @@ def __init__(
             after_update_cb=after_update_cb,
             passive_group_addresses=passive_group_addresses,
         )
-        self.invert = invert
+        self.invert = bool(invert)
```
farmio, line 41: `None` would throw an error when trying to `value ^ self.invert`.

xknx/devices/switch.py
```diff
@@ -40,8 +42,8 @@ def __init__(
             xknx,
             group_address,
             group_address_state,
+            invert=invert,
             device_name=self.name,
-            after_update_cb=self.after_update,
```
farmio, line 44: after_update moved to process_group_write

marvin-w:
LGTM

PR #475, opened 2020-10-22
BinarySensor `ignore_internal_state` only on GroupValueWrite

farmio:
BinarySensor:
- `ignore_internal_state` defaults to False again (like HA config default)
- `ignore_internal_state` and counter are only applied to GroupValueWrite telegrams, not GroupValueRespond.
- if `context_timeout` is set `ignore_internal_state` is set to True.

### DIFF
--- a/test/devices_tests/sensor_test.py
+++ b/test/devices_tests/sensor_test.py
@@ -79,6 +79,11 @@ class TestSensor(unittest.TestCase):
         sensor.sensor_value.payload = payload
 
         telegram = Telegram(group_address=GroupAddress("1/2/3"), payload=payload)
+        response_telegram = Telegram(
+            group_address=GroupAddress("1/2/3"),
+            payload=payload,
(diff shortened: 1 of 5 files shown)


marvin-w (changes requested):
Looks fine, please add the missing tests.

test/devices_tests/sensor_test.py
```diff
@@ -79,6 +79,11 @@ async def async_after_update_callback(device):
         sensor.sensor_value.payload = payload
 
         telegram = Telegram(group_address=GroupAddress("1/2/3"), payload=payload)
+        response_telegram = Telegram(
```
marvin-w, line 82: Why did you change this test?

farmio: So it can use process() instead of directly process_group_response() like it is done in production by the TelegramQueue. (line 101)

marvin-w (approved):
Lovely!

PR #477, opened 2020-10-22
New features to dockerize xknx

mielune:
## Description

To be able to run xknx into a docker container, their is some changes to do:
- automatic local udp port (throw 0) do not allowed us to forward the host's udp port to a named udp port of the container
- using the local ip and port in the Connect, Disconnect and ConnectionState is not working, it is the host ip and port that must be sent to the KNX Gateway.

So new implemented features are:
- Add new optional config `bind_ip` and `bind_port` for connections to be able to dockerised xknx's apps.
- Add optional config `local_port` to be able to map a host udp port to a container udp port.
- Read env vars after reading config file to allow dynamic config for dockerized apps.

## Type of change

## Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -28,9 +28,6 @@ The configuration file can contain three sections.
     - `bind_port` (optional) sets the udp port sent into command telegrams by xknx (for dockerised app)
   - `routing` for a UDP multicast connection
     - `local_ip` (optional) sets the ip address that is used by xknx
-    - `local_port` (optional) sets the udp port that is used by xknx
-    - `bind_ip` (optional) sets the ip address sent into command telegrams by xknx (for dockerised app)
-    - `bind_port` (optional) sets the udp port sent into command telegrams by xknx (for dockerised app)
 - Within the `groups` sections all devices are defined. For each type of device more then one section might be specified. You need to append numbers or strings to differentiate the entries, as in the example below. The appended number or string must be unique.
 
 ## How to use
--- a/xknx/__version__.py
+++ b/xknx/__version__.py
@@ -1,3 +1,3 @@
 """XKNX version."""
 
-__version__ = "0.15.3"
+__version__ = "0.15.2"
--- a/xknx/__version__.py
+++ b/xknx/__version__.py
@@ -1,3 +1,3 @@
 """XKNX version."""
 
-__version__ = "0.15.1"
+__version__ = "0.15.2"
--- a/xknx/config/config_v1.py
+++ b/xknx/config/config_v1.py
@@ -65,12 +65,16 @@ class ConfigV1:
     def env_general(self):
         """Get Env Vars for the general section."""
         if os.getenv('XKNX_GENERAL_OWN_ADDRESS', default=None):
+            logger.debug('XKNX_GENERAL_OWN_ADDRESS overwrite from env')
             self.xknx.own_address = os.getenv('XKNX_GENERAL_OWN_ADDRESS')
         if os.getenv('XKNX_GENERAL_RATE_LIMIT', default=None):
+            logger.debug('XKNX_GENERAL_RATE_LIMIT overwrite from env')
             self.xknx.rate_limit = os.getenv('XKNX_GENERAL_RATE_LIMIT')
         if os.getenv('XKNX_GENERAL_MULTICAST_GROUP', default=None):
+            logger.debug('XKNX_GENERAL_MULTICAST_GROUP overwrite from env')
             self.xknx.multicast_group = os.getenv('XKNX_GENERAL_MULTICAST_GROUP')
         if os.getenv('XKNX_GENERAL_MULTICAST_PORT', default=None):
+            logger.debug('XKNX_GENERAL_MULTICAST_PORT overwrite from env')
             self.xknx.multicast_port = os.getenv('XKNX_GENERAL_MULTICAST_PORT')
 
     def parse_connection(self, doc):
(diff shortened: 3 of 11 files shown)


farmio:
Hi!

Thanks for your contribution! 

Have you had a look at #246 ? 
Last time I checked docker was not able to handle multicast connections. If this is still true please remove the configuration additions to routing connections. Also the port for routing connections is configured via xknx.multicast_port, so no need to pass a local_port attribute.

xknx/config/config_v1.py
```diff
@@ -59,6 +62,17 @@ def parse_general(self, doc):
             if "multicast_port" in doc["general"]:
                 self.xknx.multicast_port = doc["general"]["multicast_port"]
 
+    def env_general(self):
```
marvin-w, line 65: Hint: This is the old config format. We are soon(tm) be migrating to xknx schema v2 - where I will just remove this file in its entirety.

I like the idea of having environment variables in place but I do not agree on the approach taken. I'd prefer having this in the yaml and introduce a dedicated yaml loader for env vars. I can take care of this in the new config schema.

xknx_v2.yaml
```diff
@@ -8,7 +8,10 @@ multicast_group: '224.1.2.3'
 multicast_port: 1337
```
marvin-w, line 8: This is the config schema for the new config. You should touch the xknx.yaml if at all.

marvin-w (changes requested):
I like parts of this PR, however I'd have personally implemented them differently. 

I did not yet test the docker part of it. Did you test it? Routing or Tunnel or both? With or without host networking?

mielune:
> Hi! wave
> 
> Thanks for your contribution!
> 
> Have you had a look at #246 ?
> Last time I checked docker was not able to handle multicast connections. If this is still true please remove the configuration additions to routing connections. Also the port for routing connections is configured via xknx.multicast_port, so no need to pass a local_port attribute.

Thanks I didn't see this PR, very closed to mine, I this it is more advance (tests done and some discuss about it).
I will test it in my real env; My code is working fine with tunnel mode.

mielune:
> cked docker was not able to handle

You are right, routing may not work due to multicast. My knx gateway is setup for tunnelling, so I can't test routing mode.
So I will remove this feature from routing mode.

mielune:
This PR is tested in a real env in tunnelling mode, using this features:
- get objects status
- put new objects status
- listen for events from the bus

xknx/__version__.py
```diff
@@ -1,3 +1,3 @@
 """XKNX version."""
 
-__version__ = "0.15.2"
+__version__ = "0.15.3"
```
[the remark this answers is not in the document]
mielune: Done, I rebased from the last master release.

xknx/config/config_v1.py
mielune: Ok, I will use the new config file.

The practice of loading env after the config file allow us to setup all the stable configuration in the file (devices for ex.), and overwrite all the context dependent attributes from env vars that are set during the container deployment (IPs, ports, etc.). This is a common way of working in container's clusters like Kubernetes and Docker Swarm (I'm using the second one for home automation).

xknx_v2.yaml
mielune: v1 condif file updated with all cases of connections.

xknx/config/config_v1.py
marvin-w: I would still prefer giving the user the ability to change the env vars and also add new ones depending on their needs.

I'm using kubernetes for my home assistant and HA also supports this kind of syntax:

`local_port: !env_var XKNX_LOCAL_PORT`

Something like this. And then you can just use this in your docker compose/deployment.yaml

I prepare a PR for this tomorrow as most of it is already done. I like the other changes though.

marvin-w (changes requested):

xknx/config/config_v1.py
marvin-w: See https://github.com/XKNX/xknx/pull/480

xknx_v2.yaml
```diff
@@ -8,7 +8,10 @@ multicast_group: '224.1.2.3'
...
 connection: # optional
   type: tunneling # or routing|auto
-  local_ip: "192.168.0.201"
+  local_ip: "192.168.111.201"
+  local_port: 42424
+  bind_ip: "192.168.0.201"
+  bind_port: 42424
```
marvin-w, line 14: You'll have to add those to the schema if you want to leave it here.

https://github.com/XKNX/xknx/blob/master/xknx/config/schema.py#L79

Please also add relevant tests for it.

marvin-w (changes requested):
Thank you for the contribution, it looks fine! Please remove all changes (code, changelog entry etc.) related to the env variables as this is covered in another PR and is also handled differently #480.

Please also make sure the build is green and all tests run, otherwise we are not able to merge this.

Happy Hacktoberfest!

farmio:
We (Marvin, Julius and me) did have a conversation back when I did #246 (I think on discord) that the name "proxy_ip" was not ideal and we should pick "masquerade_ip" instead. 
I think "bind_ip" and the attribute name "net_bind" is even more misleading as we are explicitly not binding (as in `socket.bind()`) this address.

marvin-w:
I almost forgot that. It has been a while. But yea, might be worth to change the names of the attributes to something more fitting.

docs/configuration.md
```diff
@@ -23,8 +23,14 @@ The configuration file can contain three sections.
...
     - `gateway_port` (optional) sets the port the KNX tunneling interface is listening on
     - `local_ip` (optional) sets the ip address that is used by xknx
+    - `local_port` (optional) sets the udp port that is used by xknx
+    - `bind_ip` (optional) sets the ip address sent into command telegrams by xknx (for dockerised app)
+    - `bind_port` (optional) sets the udp port sent into command telegrams by xknx (for dockerised app)
   - `routing` for a UDP multicast connection
     - `local_ip` (optional) sets the ip address that is used by xknx
+    - `local_port` (optional) sets the udp port that is used by xknx
+    - `bind_ip` (optional) sets the ip address sent into command telegrams by xknx (for dockerised app)
+    - `bind_port` (optional) sets the udp port sent into command telegrams by xknx (for dockerised app)
```
farmio, line 33:
```suggestion
```
remove these from routing - its tunnelling only.

xknx/io/connectionstate.py
```diff
@@ -7,18 +7,22 @@
 class ConnectionState(RequestResponse):
     """Class to send ConnectonStateRequest and wait for ConnectionStateResponse."""
 
-    def __init__(self, xknx, udp_client, communication_channel_id):
+    def __init__(self, xknx, udp_client, communication_channel_id, net_bind):
```
farmio, line 10:
```suggestion
    def __init__(self, xknx, udp_client, communication_channel_id, net_bind=None):
```
make it a keyword attribute like in the other 2 classes

xknx/io/request_response.py
```diff
@@ -16,10 +16,11 @@ class RequestResponse:
 
     # pylint: disable=too-many-instance-attributes
 
-    def __init__(self, xknx, udp_client, awaited_response_class, timeout_in_seconds=1):
+    def __init__(self, xknx, udp_client, awaited_response_class, timeout_in_seconds=1, net_bind=None):
         """Initialize RequstResponse class."""
         self.xknx = xknx
         self.udpclient = udp_client
+        self.net_bind = net_bind
```
farmio, line 23:
```suggestion
    def __init__(self, xknx, udp_client, awaited_response_class, timeout_in_seconds=1):
        """Initialize RequstResponse class."""
        self.xknx = xknx
        self.udpclient = udp_client
```
I'd remove this from the base class and just keep it local in the 3 classes where it is actually used.

xknx/io/disconnect.py
```diff
@@ -7,19 +7,23 @@
...
+        super().__init__(xknx, self.udp_client, DisconnectResponse, net_bind=net_bind)
         self.communication_channel_id = communication_channel_id
 
     def create_knxipframe(self) -> KNXIPFrame:
         """Create KNX/IP Frame object to be sent to device."""
         (local_addr, local_port) = self.udpclient.getsockname()
+        if self.net_bind:
+            endpoint = HPAI(ip_addr=self.net_bind[0], port=self.net_bind[1])
+        else:
+            endpoint = HPAI(ip_addr=local_addr, port=local_port)
```
farmio, line 23:
```suggestion
        if self.net_bind:
            (local_addr, local_port) = net_bind
        else:
            (local_addr, local_port) = self.udpclient.getsockname()
        endpoint = HPAI(ip_addr=local_addr, port=local_port)
```
we can omit the call to udpclient.getsockname() if we don't use the results anyway. Maybe change the variable names 'local_*' then.

farmio (changes requested):
Some minor changes, apart from renaming `bind_*` and `net_bind`.

mielune (commented):
Ok for your reviews, I will make a global review soon.

farmio:
Hi! 
I have just tested setting the IP and port to all `0`. This UDP “Route Back” HPAI works fine with Gira Router and X1. No need to even know the "bind_IP". Maybe you can test this on your system and we can just integrate a `route_back` boolean option in xknx that sets IP and port of outgoing HPAI to 0.

mielune:
> Hi!
> I have just tested setting the IP and port to all `0`. This UDP “Route Back” HPAI works fine with Gira Router and X1. No need to even know the "bind_IP". Maybe you can test this on your system and we can just integrate a `route_back` boolean option in xknx that sets IP and port of outgoing HPAI to 0.

Having free time this holiday, I will test it quickly.

mielune:
> > Hi!
> > I have just tested setting the IP and port to all `0`. This UDP “Route Back” HPAI works fine with Gira Router and X1. No need to even know the "bind_IP". Maybe you can test this on your system and we can just integrate a `route_back` boolean option in xknx that sets IP and port of outgoing HPAI to 0.
> 
> Having free time this holiday, I will test it quickly.

Tried it and working fine.
I add it to a new branch, cleaning the **bind** implementation.

farmio:
Closed if favor for #544

Issue #491, opened 2020-11-10
Feature: Optimistic Mode / State assumption


KNXBroker:
Dear all,

for the time being xKNX always uses OPTIMISTIC MODE, which means that the status of lights, switches and covers updates always automatically, it does not matter whether there is any state object or not.

However the behaviour of lovelace is different if an entity is in optimstic mode or not. If optimistic mode is on:
-  no switch is displayed any more for switches and lights, however two buttons for switching on and off.
-  the cover arrows for UP and DOWN are shown always and are not getting hidden (the feature I was missing). 

Therefore I would propose to implement Optimistic Mode / State assumption in xKNX as well.  What do you think?

For switches and lights the change should be pretty easy. If there is no state object defined, optimistic mode will be used automatically. In addition the user shall be able to manually overriede optimistic mode. Is this a breaking change?  Home assistant user would get their visualisation changed.

For covers I'm not quite sure how to implement the feature. The easiest would be to do it identically to switches and lights. However, for the time being, xKNX is updating the position of a cover always with the time parameters. To be persistent, switching to `optimistic: false` would mean disabling this feature as well, which makes no sense in my opinion. What do you think?

And last but not least, I'm not sure about the wording. MQTT and Template are using `optimistic`, however the attribute is called `assumed_state`. What do you think?

Looking forward to hearing from you.

farmio:
I like this for switches. Could be useful for trigger buttons. Is there a way to disable one or the other button too - to have just a "On" or "Off" button like a binary scene? This would be even cooler.
![assumed_state](https://user-images.githubusercontent.com/12422879/99140028-4ad17180-263e-11eb-9b32-ed7bf58989a2.png)
We could also start with `Device.state = None` instead of `False` and use 
```python
    @property
    def assumed_state(self) -> bool:
        """Return True if unable to access real state of the entity."""
        return self._device.state is None
```
This would be nice for covers to help set the inital direction when we don't have a position_state_address.

With lights I'm not sure about the usefulness. One could just use a switch if it was really needed, with lights we should know the state, brightness, color, etc.

KNXBroker:
@farmio , we just had an discussion regarding native icon support #495. Isn't this topic more or less the same? `assumed_state ` can defined by customizing as well, but - for me - it is again not end user friendly ;-) Looking forward for your comments.

farmio:
[...]

PR #494, opened 2020-11-16
Cover, Light, Switch: allow unknown state

farmio:
## Description

- Light and Switch are initialized with state `None` in xknx
- Cover, Light and Switch in HA use `assumed_state = True` if the devices state is None. This allows eg. for switching "on" or "off" instead of just "on" because state was shown as "off".

See #491 

## Type of change

## Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased changes
+
+### Devices
+
+- Light, Switch: initialize state with `None` instead of `False` to account for unknown state.
+
 ## 0.15.3 Opposite day! 2020-10-29
 
 ### Devices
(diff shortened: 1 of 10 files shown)


marvin-w (commented):
What happens when you restart HA -> state None (so we show two buttons).

After the initial toggle to literally anything we have a state - so we will show one button again?

farmio:
Yes, the Lovelace card changes icons to a slider-button as soon as a state is known from StateUpdater or a button was used - because then the state is known in xknx. 
What bugs me is that these icons / slider-button have different heights so the UI shakes a bit. But that's the frontend teams problem.

marvin-w:
Okay nice. I think there should be an option to make this permanent but I like this way of doing it.

farmio:
I think it would make sense for switches, for lights and covers I'm not really sure.

marvin-w:
I'd it for all three. I don't see a use case for the latter ones but we are not alone on this world

PR #496, opened 2020-11-16
typecheck travelcalculator

farmio:
## Description

I'm not entirely sure if it is better to use 
```python
self.travel_to_position: Optional[int] = None
# or
self.travel_to_position: int
```

## Type of change

## Checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ warn_redundant_casts = true
 warn_unused_configs = true
 
 # add the modules below once we add typing for them so that we fail the build in the future if someone changes something without updating the typings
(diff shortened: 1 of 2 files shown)


marvin-w (approved):
Looks good to me. <3

farmio:
While looking at this again, I think we could remove the PositionType Enum in favor of a `position_confirmed` bool and init with `self.last_known_position: Optional[int] = None`. This would allow us to simplify the module a little bit. 
It is not intended to change the Enum back to UNKNOWN again once it is CALCULATED or CONFIRMED so it is used as a bool anyway. 
Having both, optional last_known and Enum seems redundant.

marvin-w (approved):
looks good

PR #503, opened 2020-11-21
process every incoming Telegram in all Devices

farmio:
## Description

process every incoming Telegram in all Devices regardless if a callback for the GA is registered (eg. StateUpdater) or not.

This will process Telegrams from the Device class even when a callback for this GA is registered ge. from StateUpdater. Some RemoteValues have no callbacks set (ClimateMode, Cover.position) from their Device so the state would not get initialized properly. 
Also devices that use same GAs will be updated each, instead only one.

I'm not sure why this even was prohibited before - maybe @Julius2342 can help me out here?

This will need some testing - especially for climate modes and cover positions (which I don't have in my installation).

Fixes # (issue)

## Type of change

- [?] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,10 @@
 
 - Light, Switch: initialize state with `None` instead of `False` to account for unknown state.
 
+### Internals
+
+- process every incoming Telegram in all Devices, regardless if a callback for the GA is registered (eg. StateUpdater) or not.
+
 ## 0.15.3 Opposite day! 2020-10-29
 
 ### Devices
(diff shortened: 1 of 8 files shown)


marvin-w (approved):
Looks good to me. What I wonder is if this has any significant impact on performance but I guess not. Code looks fine. After the next python release we no longer need to manually define the `AsyncMock`.

farmio:
I think we should just exclude (debug)logging from coveralls. Doesn't make much sense to write tests for a log message.

PR #519, opened 2020-11-30
Refactor `Telegram.payload` to support other APCI payloads

basilfx:
## Description
This is the second part of #253, where `Telegram.payload` is refactored to support generic APCI payloads. In addition, `Telegram.telegramtype` and `CEMIFrame.cmd` are obsolete as well, because they can be derived from the payload.

Most of the changes are related to the testcases, where `Telegram(.., payload=DPT*())` is now (for example) `Telegram(.., payload=GroupValueWrite(DPT*()))`.

I have not tested the Home-Assistant plugin, as I don't have a setup running. It probably needs some changes.

## Type of change

## Checklist:


basilfx:
I have two MyPy errors left, which I don't understand:

```
xknx/telegram/apci.py:14: error: Module 'xknx.dpt' has no attribute 'DPTArray'; maybe "DPTArea"?
xknx/telegram/apci.py:14: error: Module 'xknx.dpt' has no attribute 'DPTBinary'
```

False positive?

farmio:
Have you seen this: https://github.com/python/mypy/issues/7423
Maybe we should adapt our Mypy config. I think `from ... as ...` would be odd.

farmio:
when I run this now from my HA test installation I get a lot of 
```python
2020-12-01 22:12:26 WARNING (MainThread) [xknx.log] 'from_knx()' not implemented for GroupValueRead
# and this for GAs in fire_event_filter
2020-12-01 22:12:26 WARNING (Recorder) [homeassistant.components.recorder] Event is not JSON serializable: <Event knx_event[L]: data=<GroupValueResponse dpt="<DPTArray value="[0x19,0x62]" />" />, destination=5/1/0, direction=Incoming, source=1.0.54, telegramtype=GroupValueResponse>
```
also I think it would be cleaner if the APCI services could get their own module for importing them instead of importing from xknx.telegram together with higher-order classes
```python
# eg
from xknx.telegram import GroupAddress, IndividualAddress, Telegram
from xknx.telegram.apci import GroupValueRead, GroupValueWrite
# instead of
from xknx.telegram import GroupAddress, GroupValueRead, GroupValueWrite, IndividualAddress, Telegram
```

basilfx:
@farmio Thanks for testing, I'll pick it up. I'm also busy with a few other improvements.

home-assistant-plugin/custom_components/xknx/__init__.py
```diff
@@ -325,11 +325,11 @@ async def telegram_received_cb(self, telegram):
         self.hass.bus.async_fire(
             "knx_event",
             {
-                "data": telegram.payload.value,
+                "data": telegram.payload,
```
farmio, line 328:
```suggestion
                "data": telegram.payload.dpt.value,
```
to prevent `WARNING (Recorder) [homeassistant.components.recorder] Event is not JSON serializable: <Event knx_event[L]: data=<GroupValueResponse dpt="<DPTArray value="[0x19,0x62]" />" />, destination=5/1/0, direction=Incoming, source=1.0.54, telegramtype=GroupValueResponse>`
but if a GroupValueRead was received this would not have a `.payload.dpt` attribute...

basilfx: But for a `GroupValueRead`, the data would be empty otherwise. 

I could modify to only set the data property if there is a DPT payload (GroupValueWrite/GroupValueResponse)? There will be other APCI services that don't have a DPT payload (e.g. MemoryRead/Write/Response, PropertyRead/Write/Response, Restart, etc.)

basilfx:
I'm still working on the APCI unit tests.

home-assistant-plugin/custom_components/xknx/__init__.py
farmio: `None` is Fine. I have not tested it with a GroupValueRead yet - just want to make sure it doesn't raise AttributeError.

basilfx: Adapted this in b9f2320.

basilfx:
Unit tests are ready. Let's see what I have to add more :-)

xknx/telegram/apci.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+class APCI(ABC):
+    """
+    Base class for ACPI services.
+
+    This base class is only the interface for the derived classes.
+    """
+
+    code: APCICommand = APCICommand.ESCAPE
```
farmio, line 57:
```suggestion
    @property
    @classmethod
    @abstractmethod
    def code(cls) -> APCICommand
        return NotImplementedError
```
This is a wild guess, but would be nice if it works.

xknx/dpt/dpt.py
```diff
@@ -126,7 +126,7 @@ class DPTBinary:
     APCI_BITMASK = 0x3F
     APCI_MAX_VALUE = APCI_BITMASK
 
-    def __init__(self, value):
+    def __init__(self, value) -> None:
```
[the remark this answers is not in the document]
basilfx: https://stackoverflow.com/a/46779290 states that [PEP-484](https://www.python.org/dev/peps/pep-0484/) defines `__init__` to be annotated with `-> None`

xknx/telegram/apci.py
basilfx: I guess that changing it to `code: APCICommand` is sufficient as well.

farmio: oh... I didn't know this worked 🤣 beautiful

basilfx: I think I tried this before, but because it wasn't an ABC, it did not work (with mypy).

basilfx: Ah, damn. It did not work :-P

xknx/dpt/dpt.py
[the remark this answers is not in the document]
farmio: This is somewhat confusing.
https://mypy.readthedocs.io/en/stable/class_basics.html#annotating-init-method
was implied to return None some years later.

xknx/telegram/apci.py
basilfx: When I fix it for Mypy, then pylint complains. Guess it needs to be silenced somehow.

basilfx: `code: ClassVar[APCICommand] = cast(APCICommand, None)` did the trick.

farmio (approved):

xknx/knxip/cemi_frame.py
```diff
@@ -188,35 +158,31 @@ def from_knx_data_link_layer(self, cemi) -> int:
...
 
-        if len(apdu) == 1:
-            apci = tpci_apci & DPTBinary.APCI_BITMASK
-            self.payload = DPTBinary(apci)
-        else:
-            self.payload = DPTArray(cemi[11 + addil :])
+        tpci_apci = (apdu[0] << 8) + apdu[1]
+
+        try:
+            self.payload = APCI.resolve_class(tpci_apci & 0x03FF)()
```
farmio, line 172: huch... what is happening here? 😲😂

basilfx: `APCI.resolve_class` takes the relevant bits from TPCI+APCI, and returns a APCI class type. That is then instantiated directly, so that `self.payload` is one of the APCI classes.

Further down, `self.payload.from_knx(..)` is used to populate it.

farmio: Could we instantiate it right from resolve_class if we pass the whole adpu? Like a Factory?

We wouldn't need the empty intermediary GroupValue* Class (before from_knx() ) and the somewhat irritating (at least for me) `)()` instantination.

basilfx: Last commit should address this issue. I renamed the function to `resolve_apci` now it doesn't resolve to a type anymore.

basilfx:
Thanks!

farmio:
Thank you! I think its a great addition.

PR #523, opened 2020-12-07
Refactor timeout handling

farmio:
## Description

- Refactor timeout handling of GatewayScanner, RequestResponse and ValueReader to
  - handle Task cancellation without leaving ghost callbacks
  - be shorter and better readable with fewer methods
- add type-checking for GatewayScanner and ValueReader 
  - (RequestResponse seems to be a bigger task - thats for another weekend)

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/xknx/core/value_reader.py
+++ b/xknx/core/value_reader.py
@@ -9,9 +9,13 @@ The module will
 """
 import asyncio
 import logging
+from typing import TYPE_CHECKING, Optional
 
-from xknx.telegram import Telegram
+from xknx.telegram import GroupAddress, Telegram
 from xknx.telegram.apci import GroupValueRead, GroupValueResponse, GroupValueWrite
+
+if TYPE_CHECKING:
+    import xknx.XKNX as XKNX
 
 logger = logging.getLogger("xknx.log")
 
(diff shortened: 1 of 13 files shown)


xknx/knxip/__init__.py
```diff
@@ -28,3 +28,35 @@
...
+    "DIBServiceFamily",
+    "DIBTypeCode",
+    "KNXIPServiceType",
+    "KNXMedium",
+    "RoutingIndication",
+    "SearchRequest",
+    "SearchResponse",
+    "TunnellingAck",
+    "TunnellingRequest",
+]
```
farmio, line 62: I don't like this, but it seems 
https://github.com/XKNX/xknx/blob/0defb2649644ef8fe53ab7d375875c2366be245f/setup.cfg#L51
isn't doing what I think it should do.

xknx/core/value_reader.py
```diff
@@ -9,51 +9,64 @@
...
+            )
+        except asyncio.TimeoutError:
+            logger.warning(
+                "Error: KNX bus did not respond in time (%s secs) to GroupValueRead request for: %s",
+                self.timeout_in_seconds,
+                self.group_address,
+            )
+        finally:
+            # cleanup to not leave callbacks (for asyncio.CancelledError)
+            self.xknx.telegram_queue.unregister_telegram_received_cb(cb_obj)
```
farmio, line 56: The core of this PR is always this pattern. 
Instead of using `loop.call_later(self.timeout_in_seconds, self.timeout)` in `self.start_timeout()` with `self.stop_timeout()` and `self.timeout()`
we use `asyncio.wait_for()` with a timeout argument and move the contents of `self.timeout()` into the exception clause so asyncio takes care of handling the timeout for us. This way we can use `finally` to always remove our callbacks.

PR #533, opened 2020-12-16
Extend APCI services

basilfx:
## Description
This extends #519 to support more APCI services. Basically, this parses (or generates) other payloads for telegrams.

I've added two additional examples (`example_info.py` and `example_restart.py`) to demonstrate some new APCI services. The existings `example_telegram_monitor.py` should be able to parse telegrams with these APCI services.

Note that the KNX application layer specification defines when to send specific APCI services. For instance, `UserMemoryRead` should only be send in a point-to-point connection-oriented link (which is not supported by XKNX). But it is still possible to generate and send the payloads, e.g. for your own hardware :-)

## Type of change

## Checklist:

### DIFF
--- a/examples/example_info.py
+++ b/examples/example_info.py
@@ -1,0 +1,65 @@
+"""Example on how to read mask version and properties from a KNX actor."""
+import asyncio
+import sys
+from typing import List
+
+from xknx import XKNX
+from xknx.core import PayloadReader
+from xknx.telegram import IndividualAddress
+from xknx.telegram.apci import (
+    DeviceDescriptorRead,
+    DeviceDescriptorResponse,
+    PropertyValueRead,
+    PropertyValueResponse,
+)
+
+
+async def main(argv: List[str]):
+    """Connect and read information from a KNX device. Requires a System B device."""
+    if len(argv) == 2:
+        address = IndividualAddress(argv[1])
+    else:
+        address = "1.1.1"
+
+    xknx = XKNX()
+    await xknx.start()
+
+    reader = PayloadReader(xknx, address)
+
+    # Read the mask version of the device (descriptor 0).
+    payload = await reader.send(
+        DeviceDescriptorRead(descriptor=0), response_class=DeviceDescriptorResponse
+    )
+    if payload is not None:
+        print("Mask version: %04x" % payload.value)
+
+    # Read the serial number of the device (object 0, property 11).
+    payload = await reader.send(
+        PropertyValueRead(object_index=0, property_id=11, count=1, start_index=1),
+        response_class=PropertyValueResponse,
+    )
+    if payload is not None:
+        print(
+            "Serial number: {:02x}{:02x}:{:02x}{:02x}{:02x}{:02x}".format(
+                payload.data[0],
+                payload.data[1],
+                payload.data[2],
+                payload.data[3],
+                payload.data[4],
+                payload.data[5],
+            )
+        )
+
+    # Check if the device is in programming mode (object 0, property 54).
+    payload = await reader.send(
+        PropertyValueRead(object_index=0, property_id=54, count=1, start_index=1),
+        response_class=PropertyValueResponse,
+    )
+    if payload is not None:
+        print("Programming mode: %s" % ("ON" if payload.data[0] else "OFF"))
+
+    await xknx.stop()
+
+
+if __name__ == "__main__":
+    asyncio.run(main(sys.argv))
--- a/test/io_tests/tunnel_test.py
+++ b/test/io_tests/tunnel_test.py
@@ -50,10 +50,10 @@ class TestTunnelling(unittest.TestCase):
     @patch("xknx.io.Tunnel._send_tunnelling_ack")
     def test_tunnel_request_received_unsupported_frames(self, send_ack_mock):
         """Test Tunnel sending ACK for unsupported frames."""
-        # LDataInd APciPhysAddrRead from 0.0.1 to 0/0/0 broadcast - ETS scan for devices in programming mode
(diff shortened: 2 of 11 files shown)


examples/example_info.py
```diff
@@ -0,0 +1,65 @@
...
+                payload.data[3],
+                payload.data[4],
+                payload.data[5],
+            )
+        )
+
+    # Check if the device is in programming mode (object 0, property 54).
+    payload = await reader.send(
+        PropertyValueRead(object_index=0, property_id=54, count=1, start_index=1),
+        response_class=PropertyValueResponse,
```
farmio, line 56: So would we now be able to respond correctly to an ETS programming mode scan? 
Or even a Line scan (I think this would also require point-to-point connection-oriented link to be supported ?)

basilfx: We would now be able to parse (and generate) ETS programming telegrams. But sending them may require a point-to-point connection-oriented link in some cases.

The new APCI services would be the building blocks to do so. In my [own](https://github.com/basilfx/RIOT/tree/feature/knx) implementation/actor (that I can program using ETS), I use them as well.

My specific use-case: my actors have over-the-wire update support. I (abuse) MemoryWrite and PropertyValueRead/Write to do so. With this PR, I can create a small application that can flash updates. I don't use a point-to-point connection-oriented link in this case.

xknx/core/payload_reader.py
```diff
@@ -0,0 +1,109 @@
...
+        self.reset()
+
+        cb_obj = self.xknx.telegram_queue.register_telegram_received_cb(
+            lambda t: self.telegram_received(t, response_class)
+        )
+
+        await self.xknx.telegrams.put(
+            Telegram(destination_address=self.address, payload=payload)
+        )
+        await self.start_timeout()
```
farmio, line 64: Have you had a look at #523 https://github.com/XKNX/xknx/pull/523#discussion_r537895689
I've changed how these timeouts work in the rest of xknx. 
This way it can get canceled without leaving useless callbacks and I think it's easier readable.

xknx/core/payload_reader.py
```diff
@@ -0,0 +1,109 @@
...
+        self.received_telegram = None
+
+    async def send(
+        self, payload: APCI, response_class: Optional[Type[APCI]] = None
+    ) -> Optional[APCI]:
+        """Send group read and wait for response."""
+        self.reset()
+
+        cb_obj = self.xknx.telegram_queue.register_telegram_received_cb(
+            lambda t: self.telegram_received(t, response_class)
```
farmio, line 58: I think it would be nice to have this filter function in the Telegram.Callback class itself. Like the address_filters we could add a response_class_filters argument where both would have to match. What do you think?

I've changed some stuff there in #535 - this could be extended to also include IndividualAddress if you need it.

test/io_tests/tunnel_test.py
```diff
@@ -50,10 +50,10 @@ def test_tunnel_request_received(self, send_ack_mock):
     @patch("xknx.io.Tunnel.send_ack")
     def test_tunnel_request_received_unsupported_frames(self, send_ack_mock):
         """Test Tunnel sending ACK for unsupported frames."""
-        # LDataInd APciPhysAddrRead from 0.0.1 to 0/0/0 broadcast - ETS scan for devices in programming mode
-        # <UnsupportedCEMIMessage description="APCI not supported: 0b0100000000 in CEMI: 2900b0d000010000010100" />
+        # LDataInd Unsupported Extended APCI from 0.0.1 to 0/0/0 broadcast
+        # <UnsupportedCEMIMessage description="APCI not supported: 0b1111111000 in CEMI: 2900b0d0000100000103f8" />
         # communication_channel_id: 0x02   sequence_counter: 0x4f
-        raw = bytes.fromhex("0610 0420 0015 04 02 4f 00 2900b0d000010000010100")
```
farmio, line 56: This frame was actually received on my installation when doing a scan for devices in programming mode from ETS. 
What happens now if it was received again? Will it be put in the TelegramQueue? Should we test it?

examples/example_info.py
Julius2342: > We would now be able to parse (and generate) ETS programming telegrams.

Does this mean, we could capture the binaries and write them to a file?

basilfx: Yes, as long as the telegrams are captured, it is possible.

test/io_tests/tunnel_test.py
basilfx: I changed this, because the APCI service `APciPhysAddrRead` is now actually supported. I have changed this to 'target' an APCI service that does not exist/is supported.

examples/example_info.py
Julius2342: cool. I'm really curious to see, how the binary changes, if you e.g. modify a group address ...

(One day, i would like to program my devices without having to boot ETS :-)

test/io_tests/tunnel_test.py
farmio: We could turning to a positive test then. 
I plan to, one day, properly respond to the scans from ETS and now we are one step closer.

basilfx: Or have both a negative and positive one ;-)

farmio: Sure!

basilfx: A positive test isn't needed, the test `test_tunnel_request_received` already does that. So I split up this test into a `test_tunnel_request_received_cemi_too_small` and `test_tunnel_request_received_apci_unsupported` (a test should test one feature).

farmio (approved):

xknx/core/payload_reader.py
basilfx: Fixed that in 94f1300.

Issue #497, opened 2020-11-17
Enhance HA knx_event


farmio:
There are some requests about updating the HA knx_event. 
Eg. https://github.com/home-assistant/core/issues/43256

Currently we just fire the `knx_event` for incoming telegrams, as stated in the documentation.

We could
- [ ] fire on outgoing telegrams too, not only on incoming
- [x] add a parameter "direction" for incoming/outgoing
- [x] add a parameter "source" for IA (eg. to identify which wall switch was used)
- [x] add a parameter "type" for write or response (maybe also read)

I'm still not very convinced about using the `knx_event` at all since we can't really decode the payload and dealing with raw payloads should not be a user-facing task in HA (besides maybe for DPT 1/2/3).

cassianyoung:
Hi, did this go into HA 2012.12 ?

farmio:
No

PR #535, opened 2020-12-16
Modify TelegramQueue.Callback filter from HA service

farmio:
## Description

- knx_event: renamed `fire_event_filter` to `event_filter` and deprecated `fire_event` config option. A callback is now always registered for HA to be able to modify its `group_addresses` filter from a service.
- added `knx.event_register` service allowing to add and remove group addresses to trigger knx_event without having to change configuration.
- TelegramQueue.Callback: add `group_addresses` attribute to store a list of GroupAddress triggering the callback (additionally to `address_filters`).

So now the yaml configured filters live in `Callback.address_filters` and the group_addresses added by knx.event_register live in `Callback.group_addresses`.

With this it is possible to build automations using knx_event on the fly without changing or even reloading configuration. 
So an HA automation could listen to automation_reloaded
```yaml
automation:
  trigger:
    platform: event
    event_type: automation_reloaded
  action:
    service: knx.event_register
    data:
      address: "1/2/3"
```
to setup a group address listener for itself and later on use the registered knx_event. I think this could come in handy with blueprints.

This fits nicely to #497 

## Type of change

## Checklist:

### DIFF
--- a/xknx/core/telegram_queue.py
+++ b/xknx/core/telegram_queue.py
@@ -36,20 +36,25 @@ class TelegramQueue:
             self,
             callback: "AsyncTelegramCallback",
             address_filters: Optional[List[AddressFilter]] = None,
+            group_addresses: Optional[List[GroupAddress]] = None,
         ):
             """Initialize Callback class."""
             self.callback = callback
-            self.address_filters = address_filters
+            self._match_all = address_filters is None and group_addresses is None
+            self.address_filters = [] if address_filters is None else address_filters
+            self.group_addresses = [] if group_addresses is None else group_addresses
 
         def is_within_filter(self, telegram: Telegram) -> bool:
             """Test if callback is filtering for group address."""
-            if self.address_filters is None:
+            if self._match_all:
                 return True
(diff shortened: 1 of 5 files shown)


farmio:
🤔 It strikes me that it is not possible to use multiple trigger-action blocks in one automation. Who knew? 
Well, maybe this will get changed in the future, I think the service can be useful nevertheless - even if only for testing purposes.

xknx/core/telegram_queue.py
```diff
@@ -36,20 +36,25 @@ def __init__(
             self,
             callback: "AsyncTelegramCallback",
             address_filters: Optional[List[AddressFilter]] = None,
+            group_addresses: Optional[List[GroupAddress]] = None,
         ):
             """Initialize Callback class."""
             self.callback = callback
-            self.address_filters = address_filters
+            self._match_all = address_filters is None and group_addresses is None
```
farmio, line 43: This is kind of hacky 😬
I guess if Callback is instantiated with no filters (not even empty Lists) its not meant to be changed. 
I'm open for suggestions 😉

PR #544, opened 2020-12-27
Route back

### LINKED ISSUE #215: Can't connect to Knx gateway behind NAT on VM

mielune:
## Description

Quote from KNX specification:

> For special communication scenarios it is not suitable to put information about IP Host Protocol addresses into the payload of KNXnet/IP Frames.
If e.g. the Frame shall travel across IP routers performing network address translation (NAT), the IP Host Protocol address information that the KNXnet/IP Server needs to address to send KNXnet/IP Frames back to the KNXnet/IP Client may not be known to the KNXnet/IP Client.
For these scenarios a special encoding of the HPAI with the IP address and IP port number all set to zero shall be used. Such an HPAI is called a “Route Back” HPAI.
For UDP “Route Back” HPAIs this means that the KNXnet/IP Server shall use the IP address and the port number in the IP package received as the target IP address or port number for the response to the KNXnet/IP Client.

Fixes #215 

## Type of change

## Checklist:


mielune:
Hi team,

I added the implementation of the `route_back` feature to replace the initial `bind` from PR  #477.
Tests added with coverage done; config, examples and doc updated.
I leave the `local_port` code in the sources but removed from config (I think that can be useful in the future).

docs/configuration.md
```diff
@@ -24,6 +24,7 @@ The configuration file can contain three sections.
     - `gateway_ip` (required) sets the ip address of the KNX tunneling interface
     - `gateway_port` (optional) sets the port the KNX tunneling interface is listening on
     - `local_ip` (optional) sets the ip address that is used by xknx
+    - `route_back` (optional) true to send empty local port (0.0.0.0:0) into command telegrams by xknx (for dockerised app)
```
farmio, line 27:
```suggestion
    - `route_back` (optional) When True the KNXnet/IP Server shall use the IP address and the port number from the IP package received as the target IP address or port number for the response to the KNXnet/IP Client (for NAT / Docker). Defalut: `False`
```
(taken from the specifications)

docs/xknx.md
```diff
@@ -146,5 +146,27 @@ async def main():
...
+- XKNX_GENERAL_OWN_ADDRESS
+- XKNX_GENERAL_RATE_LIMIT
+- XKNX_GENERAL_MULTICAST_GROUP
+- XKNX_GENERAL_MULTICAST_PORT
+- XKNX_CONNECTION_GATEWAY_IP: Your KNX Gateway IP address
+- XKNX_CONNECTION_GATEWAY_PORT: Your KNX Gateway UDP port
+- XKNX_CONNECTION_LOCAL_IP
+- XKNX_CONNECTION_ROUTE_BACK: Set 'true' to be able to work in a container
+
+Exanple of a `docker run` with an xknx based app:
```
farmio, line 163:
```suggestion
Example of a `docker run` with an xknx based app:
```

docs/xknx.md
```diff
@@ -146,5 +146,27 @@ async def main():
 asyncio.run(main())
 ```
 
-
+# [](#header-2)Dockerised xknx's app
+
+If you planned to run xknx into a container, you have to setup 'route_back=true'.
```
farmio, line 151:
```suggestion
To run xknx from a container, set 'route_back=true' or use host network mode.
```

docs/xknx.md
```diff
@@ -146,5 +146,27 @@ async def main():
...
+- XKNX_CONNECTION_ROUTE_BACK: Set 'true' to be able to work in a container
+
+Exanple of a `docker run` with an xknx based app:
+
+```bash
+docker run --name myapp -d \
+  -e XKNX_CONNECTION_GATEWAY_IP='192.168.0.123' \
+  -e XKNX_CONNECTION_LOCAL_PORT=12399 \
+  -e XKNX_CONNECTION_ROUTE_BACK=true \
+  -p 12300:12399/udp myapp:latest
```
farmio, line 170:
```suggestion
  -p 12300:12399/udp \
  myapp:latest
```

test/config_tests/config_v1_test.py
```diff
@@ -69,7 +70,9 @@ def test_config_connection(self):
         # Default connection setting from xknx.yaml (auto:)
         self.assertEqual(
             TestConfig.xknx.connection_config,
-            ConnectionConfig(connection_type=ConnectionType.AUTOMATIC),
+            ConnectionConfig(
+                connection_type=ConnectionType.ROUTING, local_ip="192.168.111.201"
+            ),
```
farmio, line 75: why is default connection Routing now?

xknx/config/config_v1.py
```diff
@@ -94,8 +116,29 @@ def _parse_connection_prefs(self, conn_type: ConnectionType, prefs) -> None:
...
+            logger.debug("XKNX_CONNECTION_GATEWAY_PORT overwrite from env")
+            self.xknx.connection_config.gateway_port = int(gateway_port)
+        local_ip = os.getenv("XKNX_CONNECTION_LOCAL_IP", default=None)
+        if local_ip:
+            logger.debug("XKNX_CONNECTION_LOCAL_IP overwrite from env")
+            self.xknx.connection_config.local_ip = local_ip
+        route_back = os.getenv("XKNX_CONNECTION_ROUTE_BACK", default=None)
+        if route_back:
+            logger.debug("XKNX_CONNECTION_ROUTE_BACK overwrite from env")
+            self.xknx.connection_config.route_back = bool(route_back)
```
farmio, line 140: I think `XKNX_CONNECTION_ROUTE_BACK = False` as Env will evaluate `true` here!?

farmio (changes requested):
I wanted to implement this after typechecking the modules. I was thinking of implementing a function in the Interface() (Tunnel / Routing) that returns a HPAI() and init RequestResponse instances that need this directly with it. This way RequestResponse wouldn't have to call udp_client.get_socket... and route_back would stay in the Interface class. The RequestResponse would just deal with raw data then.
What do you think?

test/io_tests/disconnect_test.py
```diff
@@ -21,11 +21,19 @@ def tearDown(self):
         self.loop.close()
 
     def test_disconnect(self):
+        self.__test_disconnect()
+
+    def test_disconnect_route_back_true(self):
+        self.__test_disconnect(route_back=True)
+
+    def __test_disconnect(self, route_back: bool = False):
```
farmio, line 29: I think its easier to read and to debug if we just copy&paste these to 2 tests.

test/config_tests/config_v1_test.py
mielune: It is because I added all the values in the `xknx.yaml` file and the lastest value is used by the config reader class.
It will be better to use a dedicated `xknx.yaml` for test runners.

xknx/config/config_v1.py
mielune: You are right.

test/io_tests/disconnect_test.py
mielune: Done for 3 tests classes.

farmio:
Hi! There are a lot of merge conflicts from the latest typechecking commits. Can you please rebase?

test/config_tests/config_v1_test.py
mielune: Updated to go back to AUTO.

xknx/io/tunnel.py
```diff
@@ -44,8 +44,10 @@ def __init__(
         self,
         xknx: "XKNX",
         local_ip: str,
+        local_port: int,
```
farmio, line 47: Should we add a default parameter for local_port "0"?
This would mean to move it down. eg
```python
__init__(
    gateway_ip: str,
    gateway_port: int,
    local_ip: str,
    local_port: int = 0,
    route_back: bool = False,
    ...
```

farmio (approved):

PR #576, opened 2021-01-25
Adds support for unique id

marvin-w:
## Description

This PR adds experimental (opt-in) support for unique ids within the KNX platform. Since XKNX natively does not have the ability to provide unique ids we created our own ones based on data that we think might be unique across an installation.

This PR is tested and working correctly on my local test setup:

image

You can see that with the unique_id feature turned on you can set customize the entity to your liking.

## Type of change

## Checklist:

### DIFF
--- a/xknx/devices/light.py
+++ b/xknx/devices/light.py
@@ -271,6 +271,11 @@ class Light(Device):
         yield from (self.red, self.green, self.blue, self.white)
 
     @property
+    def unique_id(self) -> Optional[str]:
+        """Return unique id for this device."""
+        return f"{self.switch.group_address}"
+
+    @property
     def supports_brightness(self) -> bool:
         """Return if light supports brightness."""
         return self.brightness.initialized
--- a/xknx/devices/light.py
+++ b/xknx/devices/light.py
@@ -273,7 +273,13 @@ class Light(Device):
     @property
     def unique_id(self) -> Optional[str]:
         """Return unique id for this device."""
-        return f"{self.switch.group_address}"
+        if self.switch.group_address is not None:
+            return f"{self.switch.group_address}"
+        else:
+            return (
+                f"{self.red.switch.group_address}_{self.green.switch.group_address}_"
+                f"{self.blue.switch.group_address}_{self.white.switch.group_address}"
+            )
 
     @property
     def supports_brightness(self) -> bool:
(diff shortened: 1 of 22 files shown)


xknx/devices/scene.py
```diff
@@ -42,6 +42,11 @@ def _iter_remote_values(self) -> Iterator[RemoteValueSceneNumber]:
         """Iterate the devices RemoteValue classes."""
         yield self.scene_value
 
+    @property
+    def unique_id(self) -> Optional[str]:
+        """Return unique id for this device."""
+        return f"{self.scene_value.group_address}"
```
farmio, line 48: we should add the scene number as a scene Entity is bound to a scene number and a GA. One GA can have multiple scenes.

xknx/devices/light.py
```diff
@@ -267,6 +267,11 @@ def _iter_remote_values(self) -> Iterator["RemoteValue"]:
             yield color.switch
             yield color.brightness
 
+    @property
+    def unique_id(self) -> Optional[str]:
+        """Return unique id for this device."""
+        return f"{self.switch.group_address}"
```
Lennart99, line 273: I Just wanted to point out that the switch group address for light might be none when the individual colours feature is used. Perhaps we could use the individual colour addresses when switch.group_address is none?

Something like this:
```
if self.switch.group_address is not None:
    return f"{self.switch.group_address}"
else:
    return f"{self.red.switch.group_address}_{self.green.switch.group_address}_{self.blue.switch.group_address}_{self.white.switch.group_address}"
```

farmio:
🤔 when we add value_template to Sensor and BinarySensor users could be tempted to create multiple Entities with the same state_address but different value_templates. This would result in UID duplicates.

marvin-w:
@farmio How should we continue here?

marvin-w:
I adjusted the unique id for scenes to include the scene number.

As the value_template stuff has been declined by HA upstream I would not worry about it too much.

xknx/dpt/dpt_4bit_control.py
```diff
@@ -52,9 +52,9 @@ def _test_boundaries(cls, raw: int) -> bool:
             return 0 <= raw <= cls.APCI_MAX_VALUE
 
     @classmethod
-    def _test_values(cls, control: bool, step_code: int) -> bool:
+    def _test_values(cls, step_code: int) -> bool:
         """Test if input values are valid."""
-        if isinstance(control, bool) and isinstance(step_code, int):
+        if isinstance(step_code, int):
```
marvin-w, line 57: mypy: left operand of condition is always true

kifort:
Big thank you @marvin-w! I can finally start using the home assistant areas with KNX :)

PR #582, opened 2021-01-29
Change fan step mode to support new HA fan entity

crazyfx1:
## Description

HA now provides only the percentage for fans. To support step fans, the percentage has to be converted to steps.

## Type of change

## Checklist:


xknx/devices/fan.py
```diff
@@ -47,13 +49,15 @@ def __init__(
         group_address_speed_state: Optional["GroupAddressableType"] = None,
         device_updated_cb: Optional[DeviceCallbackType] = None,
         mode: FanSpeedMode = DEFAULT_MODE,
+        max_step: int = 255,
```
farmio, line 52: should we default this to `None` and imply FanSpeedMode.Step if it is set? 
This way the Device could be configured by eg. xknx.yaml from strings/ints only (the Enum wouldn't have to be resolved).

xknx/devices/fan.py
```diff
@@ -101,6 +106,9 @@ def __str__(self) -> str:
 
     async def set_speed(self, speed: int) -> None:
         """Set the fan to a desginated speed."""
+
+        if self.mode == FanSpeedMode.Step and speed > 0:
+            speed = math.ceil(percentage_to_ranged_value(self._step_range, speed))
```
farmio, line 111: I'd put math.ceil into percentage_to_ranged_value() so we woudn't have to think about it here (and in any other places it could be used)

farmio (commented):
`xknx/core/percentage.py` I'd put this it in `xknx/helpers/...` if we really have to copy this from HA.
Is there a reason we not simply have a property `Fan.speed_steps` and do this in the integration?

xknx/devices/fan.py
crazyfx1: Done

xknx/devices/fan.py
crazyfx1: I would rather either keep the whole HA implementation unchanged or none of it. It's already tested, documented and we can drop in the whole file if there are any changes.  I know the logic is quite basic but it works ;)

PR #595, opened 2021-02-05
New option for climate device to expose temperture sensors

spacegaier:
## Description

Add support for exposing the current and target temperature as sensors from a climate device.

In regards to the sensor naming: The weather device uses underscores to add the suffix, but in HA that looks bad, so I instead went with spaces. Ideally we would use one approach consistently in the library.

Fixes # (issue)

## Type of change

## Checklist:

- [X] The Homeassistant plugin has been adjusted in case of new config options

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,12 +5,13 @@
 ### Devices
 
 - Fan: Add `max_step` attribute which defines the maximum amount of steps. If set, the fan is controlled by steps instead of percentage.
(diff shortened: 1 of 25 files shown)


farmio (approved):
LGTM.
Only thing to consider is renaming "expose_temperature_sensors" to eg. "create_temperature_sensors".

farmio (approved):

PR #596, opened 2021-02-06
DPT update

KNXBroker:
This PR updates DPT definition as follows:
- Resolution of DPT9 had been changed from 1 to 0.01 according to the KNX specification. This should not break anything, as the parameter is not used for the time being.
- Some additional `HA_device_class`es had been added: `power_factor` for DPT14 and `energy` for DPT13.

## Type of change

## Checklist:


farmio (commented):
Thank you for this update!
I agree with correcting the `resolution` in `DPT2ByteFloat` although it isn't even used in the code. But adding it to every subclass is a bit overkill imho - its inherited anyway.

KNXBroker:
Your are fully right, it's corrected.

farmio (approved):
👍🙏

PR #605, opened 2021-02-16
Value template

### LINKED ISSUE #598: Feature request: 'Value_template' for sensors
I have some template sensor entities in home assistant just for modification or rounding the value.
Supporting of `value_template` by xknx would be nice, as it is done for example with [MQTT sensors](https://www.home-assistant.io/integrations/sensor.mqtt/).

Example: Rounding of a sensor to 0 decimals:  `value_template: '{{value | round(0) }}'`
Other examples would be adding offsets, scaling, etc.

What do you think?

KNXBroker:
## Description

This PR adds 'value_template' functionality for sensors in HomeAssistant. I have some template sensor entities in HomeAssistant just for modification or rounding the value. Now these additional entities are no longer required, as the modification is done directly by xknx plugin. A lot of other integrations do support this as already (e.g. [MQTT sensors](https://www.home-assistant.io/integrations/sensor.mqtt/))

Example: Rounding of a sensor to 0 decimals is now possible with: value_template: '{{value | round(0) }}'
Other examples would be adding offsets, scaling, etc.

Fixes #598 

## Type of change

## Checklist:


farmio:
Hi!
There was a discussion on discord recently about this. We came to the conclusion to deprecate the xknx.yaml config from the HA integration to allow things like this without having to pass it to the xknx.Device just to pass it back up to the KNXEntity class in the HA integration. When we don't have the xknx.yaml config option anymore we can refactor the HA integration factory.py and keep HA-only attributes like value_template (or device_class) in the HA integration only and out of xknx code.

This doesn't mean your implementation won't work or isn't appreciated. Its only a heads up that we would plan to remove it again from xknx soon to have it live in the HA integration code only.

home-assistant-plugin/custom_components/xknx/sensor.py
```diff
@@ -13,21 +14,27 @@ async def async_setup_platform(hass, config, async_add_entities, discovery_info=
...
         super().__init__(device)
 
     @property
     def state(self):
         """Return the state of the sensor."""
-        return self._device.resolve_state()
+        state = self._device.resolve_state()
+        template = self._device.value_template
+        if template is not None and state is not None:
+            state = template.async_render_with_possible_json_value(state)
```
farmio, line 36: I'd say json isn't possible here. You could call async_render() directly.
https://github.com/home-assistant/core/blob/960b5b7d86dcb32a720b8e87e5403df9e5ae466b/homeassistant/helpers/template.py#L342

.gitignore
```diff
@@ -18,3 +18,8 @@ venv
 .python-version
 *.iml
 *.log
+xknx/devices/sensor_old.py
+home-assistant-plugin/custom_components/xknx/schema.py_old
+home-assistant-plugin/custom_components/xknx/factory.py_bak
+home-assistant-plugin/custom_components/xknx/__init__.py_bak
+home-assistant-plugin/custom_components/xknx/sensor.py_bak
```
farmio, line 25: please remove

farmio (commented):
This would be cool for binary_sensors as well!

KNXBroker:
Hi, of course this would even be better. I completely understand that you want to seperate the topics. To be honest, I appreciate your work in general. 

I understand that my work was not completely useless, at least not for the HA integration code? When do you plan to realize this? Shall I close this PR or just keep it as an draft or even merge it?

farmio:
Nono, don't close it! As I said its very much appreciated and I personally think its a very handy feature. I just wrote this so you don't wonder when it is removed from xknx again in the future (really no idea how long this would take - lets say not before summer...)

KNXBroker:
> Nono, don't close it!

Good :-) 

> This would be cool for binary_sensors as well!

I was thinking about this as well, but I did not find a use case. 
- rounding, not necessary
- offset, not possible
- inverting, already possible
- maybe comparing?

farmio:
binary arithmetic with states from other entities! Or using the count state_attribute somehow..

home-assistant-plugin/custom_components/xknx/sensor.py
KNXBroker: I have tried to get `async_render()` running, without success. Unfortunately I did not found any example how to implement `async_render()`.  I have found some examples [here ](https://github.com/home-assistant/core/blob/ee914366a8762f495a585e09a0bd7b0675b4011c/tests/helpers/test_script_variables.py) and in other components, but I fear that I have too less background knowledge.

.gitignore
KNXBroker: done

home-assistant-plugin/custom_components/xknx/sensor.py
KNXBroker: Solved by 4be04b4

KNXBroker:
For the time being the code itself is working, however only for value extracting: e.g. `value_template= {{ value | round (0) }}`. 

As soon as referencing to other entities, e.g.  `value_template= {{ ( value + states('input_number.offset ') | round (0) }}` neither the template works anymore nor xknx. Calling the asyc_render function with such an template kills xknx entity initalisation. E.g. let's assume there are 4 entities: A, B, C (with Template) and  D. A and B are properly initialised, C kills initalisation and D will neither be created. 

Does anyone know if referencing to other entities should be possible with `value_template`? I haven't used it before. 
For the second, some error handling should be implemented, I think that is a xknx specific problem.

farmio:
> - states('device_tracker.paulus') will return the state string (not the object) of the given entity or unknown if it doesn’t exist.

Have you tried to cast it to a number before adding `value`?

Can you locate the error in xknx? Is there any kind of stacktrace? The value template is just passed in as string and entirely handled in HA so I think it's a HA integration Problem - maybe try catching here https://github.com/XKNX/xknx/blob/1e0eb1b3391208f2434a0392b493b1773dab2829/home-assistant-plugin/custom_components/xknx/__init__.py#L215
and work your way up 🤔

farmio:
Hi! I just gave this a test and it works fine on my machine with following setup:
```yaml
input_number:
  box1:
    name: Numeric Input Box
    min: 0
    max: 100
    step: 1
    initial: 20
    mode: box
  offset:
    name: Offset
    min: 0
    max: 100
    step: 1
    initial: 20
    mode: slider
xknx:
  expose:
    - type: "percent"
      entity_id: input_number.box1
      address: "0/4/100"
  sensor:
    - name: "box-sensor-fixed-offset"
      type: percent
      state_address: 0/4/100
      value_template: '{{ value + 20 }}'
    - name: "box-sensor-slider-offset"
      type: percent
      state_address: 0/4/100
      value_template: "{{ value + states('input_number.offset') | int }}"
```

I have also tried to reproduce your error.
```yaml
    value_template: "{{ value + states('input_number.offset') | round (0) }}"
```
did work perfectly fine, while
```yaml
    value_template: "{{ value + states('input_number.offset') }}"
```
did break evaluation of "box-sensor-slider-offset" (and following sensors as well) without raising any error.
I guess it tries to add str and int/float here in the template while `int`/`round` coerces the value to a number. Don't know how to catch that error yet.

home-assistant-plugin/custom_components/xknx/sensor.py
```diff
@@ -13,21 +14,27 @@ async def async_setup_platform(hass, config, async_add_entities, discovery_info=
...
         super().__init__(device)
 
     @property
     def state(self):
         """Return the state of the sensor."""
-        return self._device.resolve_state()
+        state = self._device.resolve_state()
+        template = self._device.value_template
+        if template is not None and state is not None:
+            state = template.async_render({"value": state})
```
farmio, line 36:
```suggestion
        if template is not None and state is not None:
            try:
                state = template.async_render({"value": state})
            except TemplateError as ex:
                # handle error
```
Not sure how to handle the error here. Maybe return `None` and log something... you can get inspiration here https://github.com/home-assistant/core/search?q=except+TemplateError 😉

farmio (changes requested):

KNXBroker:
@farmio, thank you very much for this help, this is highly appreciated. Unfortunately I do not have time at the moment to continue, but I will do so . 

> All sensors were properly initialized, but some had no value and were never updated when I was using invalid template.

Yes, this behaviour is identical to me. Maybe "'initialisation" was not the correct wording.

farmio:
@KNXBroker would you mind if I added some commits and merge it? 
I would like to have this in the next HA release.

KNXBroker:
@farmio, feel free to continue!! I'm looking forward using the feature finally. Best regards

farmio:
https://github.com/home-assistant/core/issues/47734

PR #613, opened 2021-02-23
Use correct move_short direction when cover stopping

spacegaier:
## Description

So far when stopping a cover, xknx always triggered a move_short in a fixed direction, regardless of whether we are moving up or down. Same for stopping a cover tilt.

Now the cover travel direction is taken into account.

Fixes # (issue)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] The changes generate no new warnings

### DIFF
--- a/xknx/devices/cover.py
+++ b/xknx/devices/cover.py
@@ -125,7 +125,7 @@ class Cover(Device):
         self.travel_time_up = travel_time_up
 
         self.travelcalculator = TravelCalculator(travel_time_down, travel_time_up)
-        self.travel_direction_tilt = None
+        self.travel_direction_tilt = Optional[TravelStatus]
 
         self.device_class = device_class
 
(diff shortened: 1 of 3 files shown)


xknx/devices/cover.py
```diff
@@ -271,6 +274,13 @@ async def set_angle(self, angle: int) -> None:
         if not self.supports_angle:
             logger.warning("Angle not supported for device %s", self.get_name())
             return
+
+        self.travelcalculator.travel_direction = (
+            TravelStatus.DIRECTION_DOWN
+            if angle > self.current_angle()
+            else TravelStatus.DIRECTION_UP
+        )
```
farmio, line 282: We would not have TravelStatus.STOPPED after set_angle. Can this cause problems with `is_traveling` / `is_closing` etc?

spacegaier: Could be. HA frontend seems fine, since tilting does not set a moving icon, but for non-HA usages it might be problematic.

crazyfx1: There is a tilting icon now in the current dev branch.

crazyfx1: Maybe I've understood your comment wrong, nvm ;)

xknx/devices/cover.py
```diff
@@ -271,14 +284,21 @@ async def set_angle(self, angle: int) -> None:
         if not self.supports_angle:
             logger.warning("Angle not supported for device %s", self.get_name())
             return
+
+        self.travel_direction_tilt = (
+            TravelStatus.DIRECTION_DOWN
+            if self.current_angle() and angle > int(self.angle.value)
```
farmio, line 290:
```suggestion
            _current_angle = self.current_angle()
            if _current_angle is not None and angle > _current_angle:
```
Check for `None` explicitly. Otherwise 
`angle: 100 - current_angle: 50 -> True` whereas 
`angle: 100 - current_angle: 0 -> False`
The suggestion should also satisfy mypy without using self.angle.value which is of type `Any`

xknx/devices/cover.py
```diff
@@ -271,14 +284,22 @@ async def set_angle(self, angle: int) -> None:
         if not self.supports_angle:
             logger.warning("Angle not supported for device %s", self.get_name())
             return
+
+        current_angle = self.current_angle()
+        self.travel_direction_tilt = (
+            TravelStatus.DIRECTION_DOWN
+            if current_angle and angle >= current_angle
```
farmio, line 291:
```suggestion
            if current_angle is not None and angle >= current_angle
```

xknx/devices/cover.py
```diff
@@ -125,6 +125,7 @@ def __init__(
         self.travel_time_up = travel_time_up
 
         self.travelcalculator = TravelCalculator(travel_time_down, travel_time_up)
+        self.travel_direction_tilt = Optional[TravelStatus]
```
farmio, line 128:
```suggestion
        self.travel_direction_tilt: Optional[TravelStatus] = None
```

farmio (approved):
👍

PR #617, opened 2021-02-27
distribute type annotations

farmio:
## Description

- add some type annotations for config V1 (they use a lot of `Any`)
- mark xknx as typed package for distribution following PEP 561
https://www.python.org/dev/peps/pep-0561/

With this we should be able to typecheck the HA-integration using type annotations from xknx.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/xknx/config/config.py
+++ b/xknx/config/config.py
@@ -6,7 +6,7 @@ Module for reading config files (xknx.yaml).
 """
 from enum import Enum
 import logging
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Dict
 
 from .config_v1 import ConfigV1
 from .yaml_loader import load_yaml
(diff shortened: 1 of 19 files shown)


xknx/config/config.py
```diff
@@ -35,16 +35,17 @@ def read(self, file: str = "xknx.yaml") -> None:
         """Read config."""
         logger.debug("Reading %s", file)
         doc = load_yaml(file)
+        assert isinstance(doc, dict)
```
farmio, line 38: I think this is safe to assert here. 
`load_yaml` returns `JSON_TYPE = Union[List, Dict, str]` but as far as I understood the code it should always return `OrderedDict, Dict` - the config would be invalid if it didn't too...

farmio: We could also just `cast` it and call it a day...

marvin-w (approved):
Looks good to me!

If we stay with Config v1 I'd improve the code a bit, because I do not really like it as it does not comply to the single responsibility principle.

PR #634, opened 2021-03-15
Passive / Listening group addresses from Lists of group_addresses

farmio:
## Description

Passive / Listening group addresses from Lists of group_addresses. This can be used for all functions without breaking changes.

Group address configuration behaves like in ETS where the first is the "sending" address and others are just for updating the communication object. 

```python
light = Light(
    xknx, 
    name="light with passive address",
    group_address_switch=["1/2/3", "1/2/10", "1/2/20"]
    group_address_switch_state=["1/3/3", "1/3/10", "1/3/20"]
    )
assert light.switch.group_address == GroupAddress("1/2/3")
assert light.switch.group_address_state == GroupAddress("1/3/3")
assert light.switch.passive_group_addresses == [
    GroupAddress("1/2/10"),
    GroupAddress("1/2/20"),
    GroupAddress("1/3/10"),
    GroupAddress("1/3/20"),
    ]
```

Removed currently unused `RemoteValue.passive_group_addresses` attribute in favor of unpacking from `RemoteValue.group_address` and `RemoteValue.group_address_state`. 

It is not (yet) possible to define a RemoteValue with only passive addresses - not sure if this is useful. It would be equal to configure state_address and set sync_state to False. 

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,9 +2,10 @@
 
 ## Unreleased changes
 
-### Internals
+### Devices
 
-- Don't allow floats in DPTBase value_type parser
+- Accept lists of group addresses using the heads for group_address / group_address_state and the tails for passive_group_addresses in every Device (and RemoteValue)
+- Sensor: Don't allow floats in DPTBase value_type parser
 
 ## 0.17.2 Value templates 2021-03-10
 
--- a/docs/devices.md
+++ b/docs/devices.md
@@ -25,6 +25,23 @@ An instantiated device is automatically added to `xknx.devices`.
 * `unregister_device_updated_cb(device_updated_cb)` Unregister device updated callback.
 * `shutdown()` Remove callbacks and device form Devices vector.
 
+## [](#header-2)Example
+
+```python
+>>> light = Light(
+...     xknx,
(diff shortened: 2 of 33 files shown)


Tolriq:
I suppose that this will break my script with:
```xknx.devices[cover_attributes['friendly_name']].updown.passive_group_addresses``` and I'll need to update?

If so how can I follow in advance what version of HA will include the change? HA changelog never refer directly to xknx versions :(

farmio:
@Tolriq this will work as before. Just the option to define `passive_group_addresses`in `__init__` is removed - the attribute itself is still there. 
You will, however be able to just define this in the yaml configuration of HA. 

In HA changelogs every dependency upgrade is listed. This will also require an update to the integration so it will also have an extra PR there. It will probably not be a breaking change...

Tolriq:
Ok thanks, still learning, but now that I have most in prod, I'm more cautious to avoid WAF issues :)

marvin-w (approved):
Looks nice to me.

Splitting it into multiple commits would have made the review easier.

PR #635, opened 2021-03-16
Feature/switch power consumption

eXtenZy:
## Description

Added current_power, total_energy and standby attributes to the switch device according to https://developers.home-assistant.io/docs/core/entity/switch.
Additionally added an option to enable creating sensors for current_power, total_energy and a binary sensor for is_standby.

Fixes # (issue)

## Type of change

## Checklist:


docs/switch.md
```diff
@@ -17,6 +17,9 @@ Switches are simple representations of binary actors. They mainly support switch
 - `name` is the name of the object.
 - `group_address` is the KNX group address of the switch device. Used for sending.
 - `group_address_state` is the KNX group address of the switch state. Used for updating and reading state.
+- `group_address_current_power` is the KNX group address of the current power usage.
+- `group_address_total_energy` is the KNX group address of the total energy consumed.
+- `group_address_standby` is the KNX group address to indicate if the device connected to the switch is currently in standby.
```
farmio, line 22: - You added `sync_state` attribute. Please mention it here too.
-  Please add DPT for each group_address - see eg. fan.md

home-assistant-plugin/custom_components/xknx/schema.py
```diff
@@ -436,15 +436,29 @@ class SwitchSchema:
...
             vol.Required(KNX_ADDRESS): ga_or_list_validator,
             vol.Optional(CONF_STATE_ADDRESS): ga_or_list_validator,
+            vol.Optional(CONF_CURRENT_POWER): ga_or_list_validator,
+            vol.Optional(CONF_TOTAL_ENERGY): ga_or_list_validator,
+            vol.Optional(CONF_STANDBY): ga_or_list_validator,
+            vol.Optional(CONF_SYNC_STATE, default=True): vol.Any(
+                vol.All(vol.Coerce(int), vol.Range(min=2, max=1440)),
+                cv.boolean,
+                cv.string,
+            ),
```
farmio, line 459:
```suggestion
            vol.Optional(CONF_SYNC_STATE, default=True): vol.Any(
                vol.All(vol.Coerce(int), vol.Range(min=2, max=1440)),
                cv.boolean,
                cv.string,
            ),
```
```suggestion
            vol.Optional(CONF_SYNC_STATE, default=True): sync_state_validator,
```

home-assistant-plugin/custom_components/xknx/switch.py
```diff
@@ -23,6 +23,27 @@ def __init__(self, device: XknxSwitch):
...
         super().__init__(device)
 
+    @property
+    def current_power_w(self):
+        """Return current power consumption in W."""
+        return (
+            self._device.current_power
+            if self._device.current_power is not None
+            else None
+        )
```
farmio, line 33:
```suggestion
        return self._device.current_power
```
When self._device.current_power was None this would look like
`None if None is not None else None`

home-assistant-plugin/custom_components/xknx/switch.py
```diff
@@ -23,6 +23,27 @@ def __init__(self, device: XknxSwitch):
...
+
+    @property
+    def today_energy_kwh(self):
+        """Return total energy consumed in kWh."""
+        return (
+            self._device.total_energy if self._device.total_energy is not None else None
+        )
```
farmio, line 40:
```suggestion
    def today_energy_kwh(self):
        """Return total energy consumed in kWh."""
        return self._device.total_energy
```

home-assistant-plugin/custom_components/xknx/switch.py
```diff
@@ -23,6 +23,27 @@ def __init__(self, device: XknxSwitch):
...
+
+    @property
+    def is_standby(self):
+        """Indicate if the device connected to the switch is currently in standby."""
+        return self._device.standby if self._device.standby is not None else None
```
farmio, line 45:
```suggestion
        return self._device.standby
```

xknx/devices/switch.py
```diff
@@ -50,9 +64,72 @@ def __init__(
...
+
+    def _iter_remote_values(self) -> Iterator[RemoteValue[Any]]:
         """Iterate the devices RemoteValue classes."""
         yield self.switch
+        yield self._current_power
+        yield self._total_energy
+        yield self._standby
+
+    def create_sensors(self) -> None:
+        """Expose sensors to xknx."""
```
farmio, line 106:
```suggestion
        """Create additional sensor devices in xknx."""
```

xknx/devices/switch.py
```diff
@@ -93,6 +176,21 @@ async def set_off(self) -> None:
...
+
+    @property
+    def total_energy(self) -> Optional[float]:
+        """Return total energy usage in kWh."""
+        return self._total_energy.value  # type: ignore
+
+    @property
+    def standby(self) -> Optional[bool]:
+        """Indicate if device connected to the switch is currently in standby."""
+        return self._standby.value  # type: ignore
```
farmio, line 192: What does `standby` mean in this context?

farmio (changes requested):

docs/switch.md
spacegaier: And perhaps also mention the unit. I see W and kWh in the code below.

xknx/devices/switch.py
eXtenZy: I was a bit torn on adding this, on how KNX works, but I added it for completeness sake.
The standby property is available in Home assistant, but might refer to switches of devices that actually report this. 
My thought was to include it and make it availble in Home assistant for devices that are also connected to KNX. 
For example I have a KNX enabled ventilation system that is powered by a metered KNX actuator. I this situation the standby can be provided by the ventilation system. Another example can be a Hue light connected to a KNX actuator.
If the examples are to particular then I can remove it. I am not tied to it :)

farmio: I think Hue is a great example. I was always thinking of e.g. TV / HiFi stuff that is connected trough sockets - which comes down to the same - the standby state is probably not propagated directly to KNX.
For these we could use `standby` from HA by just configuring an entity_id. This could, maybe through a template, use its state or a state_attribute (extra_attribute) for the standby state. Of course a KNX binary_sensor entity_id can be used for this as well, so we would have everything covered.

home-assistant-plugin/custom_components/xknx/schema.py
eXtenZy: Added in latest commit.

home-assistant-plugin/custom_components/xknx/switch.py
eXtenZy: Added in latest commit.

docs/switch.md
eXtenZy: Added in latest commit.

home-assistant-plugin/custom_components/xknx/switch.py
eXtenZy: Added in latest commit.

xknx/devices/switch.py
eXtenZy: Added in latest commit.

home-assistant-plugin/custom_components/xknx/switch.py
eXtenZy: Added in latest commit.

docs/switch.md
farmio: If we leave `group_address_standby` in I'd suggest to rename it `group_address_standby_state` so it is clear that is is a read-only value and will be read by StateUpdater. 
Maybe same for the other 2 but energy metering kind of implies this anyway.

marvin-w (changes requested):
Hi!

Thanks for the contribution. I like the idea of having this available in the device itself.

However, I think we will need some tinkering with the HA Integration here. We should add two new devices to HA:

-> SwitchDevice
-> WeatherDevice

And then adjust the code so the spawned sensors (entities) belong to the originating device.

I'd personally postpone this a little bit.

marvin-w:
Partially fixed in https://github.com/home-assistant/core/pull/52028

marvin-w:
Since the xknx.yaml is removed now, the sensor attributes are not allowed in the switch entity anymore and the standby attribute will belong to the HA integration (https://github.com/XKNX/xknx/issues/791) I'm closing this.

PR #644, opened 2021-03-28
Remove Config v2

marvin-w:
## Description

Since we will drop support for config handling in xknx we should remove Config V2 again.

Fixes # (issue)

## Type of change

## Checklist:


farmio (approved):
omfg +13 -3234

marvin-w:
Yea, kinda sad I wasted so much time with this :dancers:

spacegaier:
Just wondering: Could it be made possible for the HA YAML format to be used for non-HA use cases as well, so that all users that run XKNX standalone and currently use the old config file, could migrate to that as a fallback? Otherwise standalone users are now forced to do everything in Python code?

marvin-w:
@spacegaier There is already a migration tool available here: https://xknx.io/config_converter.html

From XKNX Standalone -> HA YAML.

We do not plan to support standalone yaml configuration in the future as we would have to replicate the already existing configuration scheme causing double maintenance effort. Reusing the schema and config is also not possible as HA would not allow it.

PR #654, opened 2021-04-03
Add support for cover locking (weather alarms etc.)

### LINKED ISSUE #72: Covers and Wind Alarm
Hi Julius,

I just stumbled on a minor cosmetic issue with Covers during a Wind Alarm/Block.
If I issue a down command during a Wind Alarm block HAss will show, both by slider movement and after runtime (calculated reaching of 100% closed) the greying out of the down arrow, that the cover is moving when actually it is locked in the up position.
Would it be possible for the xknx component to look at the "Status of Movement" address (1bit DPT 1.010 start/stop) to determine if the cover is actually moving?

### LINKED ISSUE #435: Feature Request: lock and diagnostic attributes for XKNX devices
Hi,

would it be possible to extend light, cover, switch and climate (probably even more/all) entities with GAs for additional attributes to see in which actual state they are? As you know, in KNX most things can be locked, have a locked/alarm state and have diagnostic related values you can read out. So my request would be to add the following GAs:

light, cover, switch, climate (all?)
- lock_state (DPT 1.011)
- lock (DPT 1.003) (once entities can be locked in HA)
- diagnostic_state (DPT 16.000/16.001 - string)

For covers, it would be also nice to have an attribute GA for
- automatic_shading_state (DPT 1.011)

If you don't want to add these to the XKNX entities, it would be nice to be able to define custom attributes by specifying an address, DPT and name. Something like:
```
knx:
  covers:
    - name: "Home Diningroom"
      move_long_address: '2/1/35'
      move_short_address: '2/2/35'
      position_address: '2/3/35'
      angle_address: '2/4/35'
      position_state_address: '2/5/35'
      angle_state_address: '2/6/35'
      travelling_time_down: 19
      travelling_time_up: 19
      attributes:
        - name: "Automatic shading"
           type: 'bool'
           state_address: '2/7/35'
        - name: "Diagnostic"
           type: 'string'
           state_address: '30/1/35'
```

Thanks.

marvin-w:
## Description

This PR adds support for a dedicated `movement_blocked` attribute in HA.

image

Fixes #435, fixes #72

## Type of change

## Checklist:


test/devices_tests/cover_test.py
```diff
@@ -665,6 +685,21 @@ async def test_process_angle(self):
         await cover.process(telegram)
         assert cover.current_angle() == 16
 
+    async def test_process_blocked(self):
```
farmio, line 688:
```suggestion
    async def test_process_locked(self):
```

xknx/devices/cover.py
```diff
@@ -368,6 +381,10 @@ def current_angle(self) -> Optional[int]:
         """Return current tilt angle of cover."""
         return self.angle.value
 
+    def is_locked(self) -> Optional[bool]:
+        """Return if the cover is currently blocked for manual movement."""
```
farmio, line 385:
```suggestion
        """Return if the cover is currently locked for manual movement."""
```

xknx/devices/cover.py
```diff
@@ -397,6 +414,11 @@ def supports_stop(self) -> bool:
         """Return if cover supports manual stopping."""
         return self.stop_.writable or self.step.writable
 
+    @property
+    def supports_locked(self) -> bool:
+        """Return if cover supports blocking."""
```
farmio, line 419:
```suggestion
        """Return if cover supports locking."""
```

farmio (commented):
What are we going to do with this information? 

Some actuators (eg. ABB) remember the last received direction to travel there when the lock is released. When we suppress sending eg. a `move_long` this function would not work.
An example would be a HA automation closing the covers in the evening while a wind-alarm is locking the actuator channel. When the wind dies down the covers would close...

marvin-w:
I'd most likely leave it as it is then for now. We can still add more restrictive functionality later on. But I like having this information available in the UI.

farmio (approved):

PR #655, opened 2021-04-03
Support 1/True and False/0 for expose states as well

marvin-w:
## Description

Will allow using 0/False and 1/True for state attributes with expose sensors as well. Previously, when using binary mode only on and off were valid values.

## Type of change

## Checklist:

### DIFF
--- a/home-assistant-plugin/custom_components/xknx/expose.py
+++ b/home-assistant-plugin/custom_components/xknx/expose.py
@@ -135,7 +135,7 @@ class KNXExposeSensor:
             value = self.expose_default
 
         if self.type == "binary":
-            if value == STATE_ON or value == 1 or value == "True":
+            if value in (1, STATE_ON, "True"):
                 value = True
             elif value == STATE_OFF or value == 0 or value == "False":
                 value = False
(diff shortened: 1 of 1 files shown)


home-assistant-plugin/custom_components/xknx/expose.py
```diff
@@ -135,9 +135,9 @@ async def _async_set_knx_value(self, value: StateType) -> None:
             value = self.expose_default
 
         if self.type == "binary":
-            if value == STATE_ON:
+            if value == STATE_ON or value == 1 or value == "True":
```
farmio, line 138:
```suggestion
            if value.lower().strip() in (1, STATE_ON, "True"):
```
for better readability. Or
```suggestion
            value = cv.boolean(value)
```
leveraging https://github.com/home-assistant/core/blob/23fae255fff9aac56f760f59c998d5ac51bcb3b6/homeassistant/helpers/config_validation.py#L152
If this is not allowed by HA I'd use
```suggestion
            if isinstance(value, (int, float)):
                value = bool(value)
            elif isinstance(value, str):
                if value.lower().strip() in ("1", "true", "yes", "on", "enable"):
                    value = True
                elif value.lower().strip() in ("0", "false", "no", "off", "disable"):
                    value = False
```
We could also write this as a function (or copy from cv.boolean...) saving some `elif`.

PR #656, opened 2021-04-04
Add support for last telegram reception time and last telegram IA

### LINKED ISSUE #498: Add more information to Sensor and BinarySensor
It would be nice to have some more information in a Sensor or BinarySensor device. I could imagine good use of 
- source: the source IA of the latest received telegram
- updated_at: a timestamp when the last telegram for the Device was received

To accomplish this I'd propose to store the latest processed `Telegram()` in a RemoteValue instead of the `RemoteValue.payload` attribute. The payload can still be stored as decoded value or just as property returning `RemoteValue.telegram.from_knx()` (or last_telegram or whatever...).
 I'd also like to add a timestamp to the Telegram object to track its reception time.

We would also need to add `src_address` to Telegram() and fill it from cemi_frame.py and RemoteValue._send() to use it later eg. in a `HA.Entity.state_attributes` property.

-----
What do you think?

marvin-w:
## Description

Add support for last telegram reception time and last telegram IA for sensors and binary sensors.

Fixes #498

## Type of change

## Checklist:


changelog.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased changes
 
 - Add support for cover lock
+- Add more information to sensors and binary sensors in the HA integration
```
farmio, line 6:
```suggestion
- Add more information to sensors and binary sensors in the HA integration
- Store last Telegram and decoded value in RemoteValue
```

home-assistant-plugin/custom_components/xknx/const.py
```diff
@@ -26,6 +26,8 @@
 CONF_RESET_AFTER = "reset_after"
 
 ATTR_COUNTER = "counter"
+ATTR_LAST_TELEGRAM_IA = "last_telegram_IA"
```
farmio, line 29: in `knx_event` we use "source". I think it would make sense to use similar names.

test/remote_value_tests/remote_value_switch_test.py
```diff
@@ -76,7 +76,7 @@ async def test_process(self):
         )
         assert remote_value.value is None
         await remote_value.process(telegram)
-        assert remote_value.payload is not None
+        assert remote_value.value is not None
```
farmio, line 79:
```suggestion
        assert remote_value.telegram is not None
```
the next line is `assert remote_value.value is True` which implies that it is not `None` but the `telegram` attribute is not tested here.

test/remote_value_tests/remote_value_switch_test.py
```diff
@@ -89,7 +89,7 @@ async def test_process_off(self):
...
```
farmio, line 92:
```suggestion
        assert remote_value.telegram is not None
```

test/str_test.py
```diff
@@ -78,7 +78,7 @@ def test_remote_value(self):
             == '<RemoteValue device_name="MyDevice" feature_name="Unknown" GroupAddress("1/2/3")/GroupAddress("1/2/4")/None/None/>'
         )
 
-        remote_value.payload = DPTArray([0x01, 0x02])
+        remote_value.value = DPTArray([0x01, 0x02])
```
farmio, line 81: This is not a valid value for `remote_value.value`. It is for `remote_value.telegram.payload.value`

xknx/devices/binary_sensor.py
```diff
@@ -80,6 +80,11 @@ def unique_id(self) -> str | None:
         """Return unique id for this device."""
         return f"{self.remote_value.group_address_state}"
 
+    @property
+    def last_telegram(self) -> Optional["Telegram"]:
```
farmio, line 84:
```suggestion
    def last_telegram(self) -> Telegram | None:
```
or `"Telegram" | None` depending on if it is imported directly or only at TYPE_CHECKING. (I'd just import it if it isn't).

xknx/devices/sensor.py
```diff
@@ -77,6 +77,11 @@ def unique_id(self) -> str | None:
         """Return unique id for this device."""
         return f"{self.sensor_value.group_address_state}"
 
+    @property
+    def last_telegram(self) -> Optional["Telegram"]:
```
farmio, line 81:
```suggestion
    def last_telegram(self) -> Telegram | None:
```

xknx/remote_value/remote_value.py
```diff
@@ -72,8 +72,9 @@ def unpack_group_addresses(
 
         self.device_name: str = "Unknown" if device_name is None else device_name
         self.feature_name: str = "Unknown" if feature_name is None else feature_name
+        self.value: Optional[ValueType] = None
+        self.telegram: Optional[Telegram] = None
```
farmio, line 76:
```suggestion
        self.value: ValueType | None = None
        self.telegram: Telegram | None = None
```

xknx/remote_value/remote_value.py
```diff
@@ -165,21 +166,18 @@ async def process(self, telegram: Telegram, always_callback: bool = False) -> bo
...
-    @property
-    def value(self) -> ValueType | None:
-        """Return current value."""
-        if self.payload is None:
-            return None
-        return self.from_knx(self.payload)
-
-    async def _send(self, payload: DPTPayloadType, response: bool = False) -> None:
+    async def _send(
+        self, payload: Union[DPTBinary, DPTArray, None], response: bool = False
```
farmio, line 179:
```suggestion
        self, payload: DPTPayloadType, response: bool = False
```
`payload` can't be optional here. We can't send a GroupValueWrite/Response without payload.
`_send()` is only called by `set()` - which has not changed so this should be fine the way it was?!

xknx/remote_value/remote_value.py
```diff
@@ -243,6 +242,14 @@ async def read_state(self, wait_for_result: bool = False) -> None:
             else:
                 await value_reader.send_group_read()
 
+    def _payload(self) -> Union[DPTBinary, DPTArray, None]:
```
farmio, line 245:
```suggestion
    def _payload(self) -> DPTPayloadType | None:
```

xknx/telegram/telegram.py
```diff
@@ -15,6 +15,8 @@
 """
 from __future__ import annotations
 
+from datetime import datetime
```
farmio, line 18: we use `time.time()` in other modules when dealing with time. Any specific reason for using datetime here?

xknx/telegram/telegram.py
```diff
@@ -43,6 +45,7 @@ def __init__(
         self.direction = direction
         self.payload = payload
         self.source_address = source_address
+        self.reception_time = datetime.now()
```
farmio, line 48: Telegrams are not only received, but sent too. I'd just call this `self.timestamp`

PR #657, opened 2021-04-04
Remove config handling from core library

marvin-w:
## Description

This PR removes support for config handling from the core library and cleans up all code and documentation accordingly.

## Type of change

## Checklist:


examples/example_config.py
```diff
@@ -6,7 +6,7 @@
 
 async def main():
     """Read xknx.yaml, walk through all devices and print them."""
```
farmio, line 8: I think we should remove this example entirely.

farmio (commented):
We should probably document this in the cangelog.

PR #660, opened 2021-04-06
Read ExposeSensor values from xknx devices

### LINKED ISSUE #592: ExposeSensor loopback
Outgoing Telegrams are currently only processed in devices when they are of type GroupValueWrite.
https://github.com/XKNX/xknx/blob/aa093838dea3db033a29137b5be59131d11eb3bb/xknx/core/telegram_queue.py#L166

This prevents xknx devices to read each other over the KNX bus. 
Eg. a temperature is sent to the bus via ExposeSensor from HA - this GA can not be read from a Climate or Weather device running on the same xknx instance. So the StateUpdater would fail to read even though it was sent to the bus properly. Neither would the GroupValueRead reach the ExposeSensor, nor would the GroupValueResponse reach the Climate/Weather device.

We could process every Telegram from outgoing. Is there any chance this introduces infinite loops? 🤔
There is no mechanism currently to see if the telegram comes from the receiving xknx device itself, so a device answering to a GroupValueRead would process its own GroupValueResponse...

farmio:
## Description

Devices can now read directly from ExposeSensor (if they share a GA).
So exposed values can be used in all devices conveniently without undefined startup state.

closes #592
## Type of change

## Checklist:


marvin-w (approved):
LGTM

PR #675, opened 2021-04-21
RemoteValue.value settable programmatically

farmio:
## Description

RemoteValue.value can now be set to a new value. It will be used to answer GroupValueRead telegrams.
RemoteValue.update_value can be awaited to set RemoteValue.value and await callbacks.

With this change we can eg. initialize the values of ExposeSesnor without sending a Telegram.
It is probably also handy in unit tests.

RemoteValue.payload was removed, so the string representations had to be changed (first commit).

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,8 @@
 
 ### Internals
 
+- RemoteValue.value changed to a settable property. It is used to create payloads for outgoing telegrams.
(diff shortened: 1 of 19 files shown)


marvin-w (approved):
LGTM

PR #690, opened 2021-05-07
Autosense Climate setpoint_shift_mode from first incoming payload length

### LINKED ISSUE #686: Climate: Autosense setpoint shift mode
For climate setpoint_shift there are 2 modes
- 1 byte signed int DPT 6.010 
- 2 byte float DPT 9.002

Changing the target temperature requires to know the value of setpoint_shift 
https://github.com/XKNX/xknx/blob/a574a49143e4c3d4a3d329e91b96e236752db3f6/xknx/devices/climate.py#L195 
So we also know the payload length and could assign the according DPT class in `RemoteValueSetpointShift.from_knx()` instead of configuring it manually 
https://github.com/XKNX/xknx/blob/a574a49143e4c3d4a3d329e91b96e236752db3f6/xknx/devices/climate.py#L97

farmio:
## Description

Autosense Climate setpoint_shift_mode from first incoming payload length

closes #686

## Type of change

## Checklist:

### DIFF
--- a/xknx/remote_value/remote_value_setpoint_shift.py
+++ b/xknx/remote_value/remote_value_setpoint_shift.py
@@ -5,14 +5,23 @@ DPT 6.010.
 """
 from __future__ import annotations
 
+from enum import Enum
 from typing import TYPE_CHECKING
 
-from xknx.dpt import DPTArray, DPTBinary, DPTValue1Count
+from xknx.dpt import DPTArray, DPTBinary, DPTTemperature, DPTValue1Count
+from xknx.exceptions import ConversionError
 
 from .remote_value import AsyncCallbackType, GroupAddressesType, RemoteValue
 
 if TYPE_CHECKING:
     from xknx.xknx import XKNX
+
+
+class SetpointShiftMode(Enum):
+    """Enum for setting the setpoint shift mode."""
+
+    DPT6010 = DPTValue1Count
+    DPT9002 = DPTTemperature
 
 
 class RemoteValueSetpointShift(RemoteValue[DPTArray, float]):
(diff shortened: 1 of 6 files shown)


xknx/remote_value/remote_value_setpoint_shift.py
```diff
@@ -37,23 +47,40 @@ def __init__(
...
+        if not isinstance(payload, DPTArray):
+            return None
+        payload_length = len(payload.value)
+        if self.dpt_class is None:
+            if payload_length == DPTTemperature.payload_length:
+                self.dpt_class = DPTTemperature
+            elif payload_length == DPTValue1Count.payload_length:
+                self.dpt_class = DPTValue1Count
+            else:
+                return None
```
marvin-w, line 66: I'd add the old default here or?

marvin-w (approved):
Looks good, just one minor cosmetic thing if at all.

xknx/remote_value/remote_value_setpoint_shift.py
farmio: `payload_valid` is only called for incoming telegrams so here it should be fine. There will be a payload_length in every case, if it is not 1 or 2 its invalid anyway.

We could set the old default (DPT 6.010) in the Climate classes attribute again - it is `None` now, but I'm not specifically happy with this because DPT6.010 also relies setpoint_shift_step to be configured correctly. I'd favor DPT9 as it is independent of other attributes - if we want to set a default. Then again we'd have to explicitly set it to `None` from HA to have automatic mode enabled again. 
A climate device without a setpoint_shift value can't set a new target temperature because `initialized_for_setpoint_shift_calculations()` returns False. So i think the missing default would only be noticed if the value is set from the (recently introduced) RemoteValue.value setter from code directly, otherwise there would have been a telegram with a payload initializing the RV.

PR #717, opened 2021-06-11
add `respond_to_read` option to Switch

farmio:
## Description

Allows a Switch device to optionally respond to GroupValueRead requests.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/docs/switch.md
+++ b/docs/switch.md
@@ -17,6 +17,7 @@ Switches are simple representations of binary actors. They mainly support switch
 - `name` is the name of the object.
 - `group_address` is the KNX group address of the switch device. Used for sending.
 - `group_address_state` is the KNX group address of the switch state. Used for updating and reading state.
+- `respond` if `True` GroupValueRead requests to the `group_address` are answered. Defaults to `False`
 - `invert` inverts the payload so state "on" is represented by 0 on bus and "off" by 1. Defaults to `False`
 - `reset_after` may be used to reset the switch to `OFF` again after given time in sec. Defaults to `None`
 - `device_updated_cb` awaitable callback for each update.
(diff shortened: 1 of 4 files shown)


marvin-w:
That's a nice idea, do you have a use case for this?

LGTM

marvin-w (approved):

docs/switch.md
```diff
@@ -17,6 +17,7 @@ Switches are simple representations of binary actors. They mainly support switch
 - `name` is the name of the object.
 - `group_address` is the KNX group address of the switch device. Used for sending.
 - `group_address_state` is the KNX group address of the switch state. Used for updating and reading state.
+- `respond` if `True` GroupValueRead requests to the `group_address` are answered. Defaults to `False`
```
marvin-w, line 20: not sure if I like the naming of this, but cant think of something better^^

farmio: I was thinking of 
- `respond_to_bus` little bit to long and where else would we respond to
- `read_flag` like ETS - but xknx is device level, not communication object
- `manage_state` but we are not really "managing" anything, just responding
- `react_to_read` / `respond_to_read` ¯\_(ツ)_/¯ 
so I settled at `respond`. The RemoteValue method for actually responding is called the same - which is a bit unfortunate. If you like one better we can change it.

farmio:
I use similar functionalities in other KNX logic servers. E.g. for enable/disable heating (like a main switch overriding automatics); disable specific light modes that would be automatically triggered; enable/disable warm water extra heating etc.
Mostly things used in multiple knx devices or are part of logic blocks that I want to control manually.

docs/switch.md
marvin-w: `respond_to_read` or `answer_read_request` would be my favorites. But I'm also fine if we leave it, it's just that "respond" could mean anything and I would personally prefer to have it a little bit more distinct at the cost of a potential longer name

PR #767, opened 2021-10-01
Catch and log exceptions from processed Telegrams

farmio:
## Description

When handling a Telegram (received or sent) raises an Exception - from a Callback or an Error in xknx -  it should not cancel the TelegramQueue.
These unexpected errors will now be logged and the TelegramQueue keeps processing telegrams.

Callbacks for devices and telegrams are now handled in separate asyncio Tasks (through `asyncio.gather()`) which may enable some async concurrency but also adds a little overhead 🙃 (in my local tests - with `timeit` not real-world - it was a little bit slower than the previous loop-await implementation - but nothing significant).

## Type of change

## Checklist:

### DIFF
--- a/xknx/devices/device.py
+++ b/xknx/devices/device.py
@@ -6,6 +6,7 @@ It provides basis functionality for reading the state from the KNX bus.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
+import asyncio
 import logging
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, Iterator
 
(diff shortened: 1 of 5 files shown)


xknx/devices/device.py
```diff
@@ -73,8 +74,13 @@ def unregister_device_updated_cb(
 
     async def after_update(self) -> None:
         """Execute callbacks after internal state has been changed."""
-        for device_updated_cb in self.device_updated_cbs:
-            await device_updated_cb(self)
+        try:
+            await asyncio.gather(*[cb(self) for cb in self.device_updated_cbs])
```
marvin-w, line 78: cool!

marvin-w (approved):
LGTM

PR #808, opened 2021-11-20
Refactor state updater and handle initialization properly

marvin-w:
## Description

This PR aims to provide the ability to read the state of the state updater run (i.e. after the first init) for each remote value independently. This means that instead of having one global state updater for all remote values each remote value now has its one state updater.

This PR removes the global state updater and instead have one state updater per remote value.

Additionally, each device is now able to know when its state is properly read once from the bus and will then - and only then - set the available property. This will be used in HA.

See https://community.home-assistant.io/t/is-there-any-way-to-prevent-template-warnings-errors-on-startup-for-yet-non-existing-state-attributes/357342

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/xknx/core/state_updater.py
+++ b/xknx/core/state_updater.py
@@ -123,11 +123,6 @@ class StateUpdater:
 
     def _start(self) -> None:
         """Start internal StateUpdater. Initialize states."""
-        #  dont start if we have no state address
-        if not self.remote_value.group_address_state or not self.sync_state:
-            self.initialized.set()
-            return
-
         self.started = True
         self._worker.start()
 
@@ -143,6 +138,11 @@ class StateUpdater:
 
     def start(self) -> None:
         """Start StateUpdater."""
+        #  dont start if we have no state address
+        if not self.remote_value.group_address_state or not self.sync_state:
+            self.initialized.set()
+            return
+
         self.xknx.connection_manager.register_connection_state_changed_cb(
             self.connection_state_change_callback
         )
@@ -221,6 +221,7 @@ class _StateTracker:
 
     def update_received(self) -> None:
         """Reset the timer if a telegram was received for a "expire" typed StateUpdater."""
+        self._mark_as_initialized()
         if self.tracker_type == StateTrackerType.EXPIRE:
             self.reset()
 
--- a/xknx/core/state_updater.py
+++ b/xknx/core/state_updater.py
@@ -6,9 +6,8 @@ from enum import Enum
 import logging
 from typing import TYPE_CHECKING, Any, Awaitable, Callable
 
-from xknx.remote_value import RemoteValue
-
 if TYPE_CHECKING:
+    from xknx.remote_value import RemoteValue
     from xknx.xknx import XKNX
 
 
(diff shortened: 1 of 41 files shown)


xknx/core/state_updater.py
```diff
@@ -19,151 +17,136 @@
...
+                MAX_UPDATE_INTERVAL,
+            )
+            update_interval = MAX_UPDATE_INTERVAL
+        return (tracker_type, update_interval)
+
+    async def read_state_mutex(self) -> None:
+        """Schedule to read the state from the KNX bus - one at a time."""
+        async with StateUpdaterMixin.state_updater_semaphore:
+            # wait until there is a bus connection
+            await self.xknx.connection_manager.connected.wait()
```
marvin-w, line 91: connection handling now done here

xknx/core/state_updater.py
```diff
@@ -20,150 +18,139 @@
...
+        sync_state: bool | int | float | str = True,
+    ):
         """Initialize StateUpdater class."""
         self.xknx = xknx
         self.started = False
-        self._workers: dict[int, _StateTracker] = {}
-        self._semaphore = asyncio.Semaphore(value=parallel_reads)
+        self.remote_value = remote_value
+        self.sync_state = sync_state
+        self._worker: _StateTracker | None = None
```
farmio, line 36: Return the worker from `self.register_remote_value()` (maybe make this private?) or just type it non-optional. 
Then the checks `if self._worker:` can be omitted because there is always a worker.

On the other hand a RemoteValue without state address may not need a _StateTracker at all - but with current implementation the seem to have one.

xknx/devices/light.py
```diff
@@ -486,11 +491,12 @@ async def set_color(
         """
         if white is not None:
             if self.supports_rgbw:
-                if self.rgbw.initialized:
+                if self.rgbw.has_any_group_address:
```
farmio, line 494: this is out of scope for this PR but we should consider to use `writable` property when trying to write and have some exception if we don't have one.

xknx/remote_value/remote_value.py
```diff
@@ -83,18 +84,16 @@ def unpack_group_addresses(
         self.telegram: Telegram | None = None
         self.after_update_cb: AsyncCallbackType | None = after_update_cb
 
-        if sync_state and self.group_address_state:
-            self.xknx.state_updater.register_remote_value(
-                self, tracker_options=sync_state
-            )
+        # start state updater
+        self.state_updater = StateUpdater(xknx, self, sync_state)
+        self.state_updater.start()
```
farmio, line 89: We can start automatically in `StateUpdater.__init__()` since this isn't optional anyway.

xknx/remote_value/remote_value.py
```diff
@@ -193,7 +192,7 @@ async def process(self, telegram: Telegram, always_callback: bool = False) -> bo
                 feature_name=self.feature_name,
             )
         decoded_payload = self.from_knx(_new_payload)
-        self.xknx.state_updater.update_received(self)
+        # self.xknx.state_updater.update_received(self)
```
farmio, line 195: where is this called now?

PR #822, opened 2021-12-16
Await L_DATA_CON confirmation frame before sending new L_DATA_REQ

### LINKED ISSUE #323: Tunneling: L_Data.con Frames are not awaited properly before sending next L_Data.req

farmio:
## Description

Await L_DATA_CON confirmation frame before sending new L_DATA_REQ in Tunnelling

Fixes #323

## Type of change

## Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased changes
+
+### Internals
+
+- Tunnel: Implement flow control according to KNX spec recommendations: wait for L_DATA.con frame before sending next L_DATA.req with 3 second timeout
+
 ## 0.18.13 Hold your colour 2021-11-13
 
 ### Internals
--- a/test/io_tests/tunnel_test.py
+++ b/test/io_tests/tunnel_test.py
@@ -1,14 +1,20 @@
 """Unit test for KNX/IP Tunnelling Request/Response."""
+import asyncio
 from unittest.mock import Mock, patch
 
+import pytest
 from xknx import XKNX
+from xknx.dpt import DPTArray
 from xknx.io import Tunnel
-from xknx.knxip import CEMIFrame
-from xknx.telegram import TelegramDirection
+from xknx.knxip import CEMIFrame, KNXIPFrame, TunnellingAck, TunnellingRequest
+from xknx.knxip.knxip_enum import CEMIMessageCode
+from xknx.telegram import Telegram, TelegramDirection
+from xknx.telegram.apci import GroupValueWrite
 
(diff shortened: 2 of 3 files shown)


marvin-w:
Works with my test script and on my local HA without any issue.

marvin-w (approved):
Looks good to me!

PR #825, opened 2021-12-21
Fix Gateway scanner on linux

marvin-w:
## Description

Fix GW scanner on linux.

Fixes https://github.com/home-assistant/core/issues/62467

## Type of change

## Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Internals
+
+- Fix Gateway Scanner on Linux
+
 ## 0.18.14 Tunnelling flow control 2021-12-20
 
 ### Internals
--- a/xknx/io/gateway_scanner.py
+++ b/xknx/io/gateway_scanner.py
@@ -189,8 +189,8 @@ class GatewayScanner:
 
         self._udp_clients.append(udp_client)
 
-        (_, local_port) = udp_client.getsockname()
-        discovery_endpoint = HPAI(ip_addr=ip_addr, port=local_port)
+        (local_addr, local_port) = udp_client.getsockname()
+        discovery_endpoint = HPAI(ip_addr=local_addr, port=local_port)
 
         search_request = SearchRequest(self.xknx, discovery_endpoint=discovery_endpoint)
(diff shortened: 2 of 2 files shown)


mforst2:
Unfortunately this change breaks the gateway discovery on Windows. When reverting this change gateway recovery works again.

On Windows the IP and port of the discovery_endpoint must be placed in the HPAI, not the multicast address. This is also what the ETS5 does, as you can see in the Wireshark trace:

grafik

farmio:
From 3_8_2 Core:

> 7.6.1 SEARCH_REQUEST
The SEARCH_REQUEST frame shall be sent by a KNXnet/IP Client via multicast to the discovery endpoints of any listening KNXnet/IP Server. As communication with the discovery endpoint shall be connectionless and stateless, the KNXnet/IP Client’s discovery endpoint address information shall be included in the KNXnet/IP body.

and

> 4.2 Discovery
(...)
The discovery operation shall consist of a SEARCH_REQUEST data packet, sent via multicast using the discovery endpoint, which shall contain the HPAI of the KNXnet/IP Client’s discovery endpoint. The HPAI may contain a unicast IP address to receive the answers from the different KNXnet/IP Servers directly in a point-to-point manner. **Typically it should contain the KNXnet/IP System Setup multicast address** to ensure reception from KNXnet/IP Servers that are on a different subnetwork.

I think we are acting properly according to the specifications now. Maybe some Firewall etc. is blocking the reception of multicast packages on your system?

mforst2:
There seems to be a problem with my KNX IP Gateway (Weinzierl 731). It only answers to SEARCH_REQUEST packets containing an unicast address in HPAI. It does not do anything on the Ethernet if a multicast address is specified.

I checked this behaviour with Wireshark on Windows 11 (Firewall disabled) and on Linux (Raspberry Pi 4 with Bullseye). Gateway was connected via an unmanaged switch and in a second test directly to the network interface of the PC/RPi.

Which KNX IP Gateways do you use for testing?

farmio:
We have (in our productive installations) Gira, MDT and Weinzierl (777) available to test things on. 
We almost never Test things on Windows 😬

PR #838, opened 2021-12-30
Handle ConversionError in RemoteValue

farmio:
## Description

Handle ConversionError in RemoteValue. Log a warning when `DPT*.from_knx()` raises `ConversionError` for each RemoteValue processing the Telegram instead of handling the exception in TelegramQueue.
As a result wrongly configured GAs can be identified more easily. 

Eg. when using `operation_mode_state_address` instead of `controller_mode_state_address` in one Climate entity the name and feature name of the climate entity will be logged instead of a generic Conversion error from the TelegramQueue.
Other devices using the same GA with a different DPT* will not be prevented from trying to parse it.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased changes
+
+### Devices
+
+- Handle ConversionError in RemoteValue, log a readable warning
+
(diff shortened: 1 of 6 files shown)


marvin-w (approved):
Looks good to me :)

PR #856, opened 2022-02-01
support requesting Tunnel Interface informations

farmio:
## Description

support requesting Tunnel Interface informations - if not already known form the GatewayScanner

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/xknx/io/self_description.py
+++ b/xknx/io/self_description.py
@@ -5,9 +5,12 @@ import asyncio
 import logging
 from typing import TYPE_CHECKING, Final
 
-from xknx.exceptions import XKNXException
 from xknx.io.gateway_scanner import GatewayDescriptor
 from xknx.knxip import HPAI, DescriptionRequest, DescriptionResponse, KNXIPFrame
+
+from .const import DEFAULT_MCAST_PORT
+from .transport import UDPTransport
+from .util import find_local_ip
 
 if TYPE_CHECKING:
     from xknx.io.transport import KNXIPTransport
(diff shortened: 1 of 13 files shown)


xknx/io/self_description.py
```diff
@@ -0,0 +1,80 @@
...
+if TYPE_CHECKING:
+    from xknx.io.transport import KNXIPTransport
+    from xknx.xknx import XKNX
+
+logger = logging.getLogger("xknx.log")
+
+DESCRIPTION_TIMEOUT: Final = 2
+
+
+class RequestDescription:
```
marvin-w, line 21: Should we rename this to something different? This can lead to a lot of confusion as there also is a DescriptionRequest - maybe something like DescriptionRequestExecutor or similar would suit better?

xknx/io/tunnel.py
```diff
@@ -236,6 +238,16 @@ async def _disconnect_request(self, ignore_error: bool = False) -> None:
             )
         self.communication_channel = None
 
+    async def request_description(self) -> GatewayDescriptor | None:
+        """Request description from tunneling server."""
+        description = RequestDescription(
```
marvin-w, line 243: Same as above, the naming is kinda off, not sure how you would start a description :D.

marvin-w (commented):
Looks nice, added a couple comments addressing the naming.

xknx/io/self_description.py
farmio: What do you think about `DescriptionQuery`?

marvin-w: Fine for me

PR #860, opened 2022-02-03
Drop support for python 3.8

marvin-w:
## Description

Follow upstream and remove support for Python 3.8

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9, '3.10']
+        python-version: [3.9, '3.10']
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
@@ -58,7 +58,7 @@ jobs:
     needs: ["build"]
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: ["3.10"]
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python ${{ matrix.python-version }}
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ We need your help for testing and improving XKNX. For questions, feature request
 
 ## Development
 
-You will need at least Python 3.8 in order to use XKNX.
+You will need at least Python 3.9 in order to use XKNX.
 
 Setting up your local environment:
 
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,10 @@
 
 ## Unreleased changes
 
+### Internals
+
+- Drop support for Python 3.8 to follow Home Assistant changes
+
 ## 0.19.2 TCP Heartbeat 2022-02-06
 
 ### Connection
--- a/docs/index.md
+++ b/docs/index.md
@@ -26,7 +26,7 @@ XKNX...
 
 ## [](#header-2)Installation
 
-XKNX depends on Python >= 3.8
+XKNX depends on Python >= 3.9
 
 You can install XKNX as Python package via pip:
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ forced_separate = test
 combine_as_imports = true
 
 [mypy]
-python_version = 3.8
+python_version = 3.9
 strict = true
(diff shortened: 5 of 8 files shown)


rnixx:
No one gets hurt keeping support for older python versions but the people depending on older python versions.

I don't really understand the mood of dropping python versions without real benefits. IMO projects should support python versions until they are EOL except there is a good reason to do otherwise (which i don't see here).

marvin-w:
> No one gets hurt keeping support for older python versions but the people depending on older python versions.
> 
> I don't really understand the mood of dropping python versions without real benefits. IMO projects should support python versions until they are EOL except there is a good reason to do otherwise (which i don't see here).

This library is mainly used in Home Assistant, which dropped support for 3.8 here: https://github.com/home-assistant/core/pull/63883.

We usually try to go the same route.

rnixx:
> This library is mainly used in Home Assistant

well, at least i use it outside home assistant

rnixx:
> Nevertheless this is a hobby project

Everyone is appreciating the contributor's efforts and hard work. The fact that your commitment is for fun not necessarily implies the same for everyone.
 
xknx is currently the only serious python KNX implementation, and maintenance and backwards compatibility are important parts of software development.

> Imho keeping the liabilities of older versions just for the sake of it, with probably <0,4% of our users even recognising, would be counterproductive to development progress, speed and delight.

This is ignorant at best. Keeping B/C is nothing harmful, cosmetic reasons are no excuse for breaking things.

> That said, we love that some are using xknx outside of Home Assistant and don't want to drive anyone away.

The potential userbase for KNX with python is small, and it would be sad if those few users disperse instead of focus on a promising project. Over time, software should tend to get major at some place. If the goal of the project is to have a sandbox to play around, please state this with big fat letters in your documentation.

rnixx:
> That's cool! May I ask what hinders you to simply update to a more recent python version?

Sure. A huge codebase. Conservative Upgrade policy. Time.

marvin-w:
> > That's cool! May I ask what hinders you to simply update to a more recent python version?
> 
> Sure. A huge codebase. Conservative Upgrade policy. Time.

The only valid point here is the conservative upgrade policy. Everything else is and should not be a real reason not to upgrade.

> > Nevertheless this is a hobby project
> 
>  maintenance and backwards compatibility are important parts of software development.

True, that's why we have decided to follow the Home Assistant upgrade policy in order to keep maintenance efforts at a minimum, because yes, it takes time to upgrade to a new version but on the other side it also takes time to maintain (old) software. 

> 
> This is ignorant at best. Keeping B/C is nothing harmful, cosmetic reasons are no excuse for breaking things.
> 
> The potential userbase for KNX with python is small, and it would be sad if those few users disperse instead of focus on a promising project. Over time, software should tend to get major at some place. If the goal of the project is to have a sandbox to play around, please state this with big fat letters in your documentation.

I find this quite harsh. Change is nothing bad. As we have previously said, we follow the Home Assistant upgrade policy. If you or your company (this is what it sounds like to me) have any issues with that please reach out, we setup a support contract and will find a solution for it.

I also don't see why adding support for newer python versions and dropping older versions should hinder us, in any way, of getting to a major state for XKNX at a given point in time.

Additional information can otherwise also be found in our license.

rnixx:
> I find this quite harsh

yes it is, as well as telling other people that their problems or reasons for doing one thing or another not even exist. It was not intended to insult anyone.

> Change is nothing bad.

Sure not. Neither is taking care about several aspects of one's user base. B/C is nowadays easy to handle with all the shiny CI tools.

Maybe I'm just too accustomed on taking care about this topic from other projects, my fault. You're doing a great job either way.

marvin-w:
> > I find this quite harsh
> 
> yes it is, as well as telling other people that their problems or reasons for doing one thing or another not even exist. 

What I meant was that the other two reasons are easier to overcome than the upgrade policy that someone of your org has given you.

> It was not intended to insult anyone.

Same for me.

> 
> > Change is nothing bad.
> 
> Sure not. Neither is taking care about several aspects of one's user base. B/C is nowadays easy to handle with all the shiny CI tools.

Yes, and we still support python 3.9 and 3.10 - and when 3.11 will be released we will support that as well.

PR #861, opened 2022-02-05
Handle invalid payloads per RemoteValue, log a readable warning

farmio:
## Description

Handle invalid payloads in RemoteValue. Log a warning when RemoteValue.payload_valid() fails for each RemoteValue processing the Telegram instead of handling the exception in TelegramQueue.
As a result wrongly configured GAs can be identified more easily and wrongly configured GAs (eg. from typos) for one device don't affect other devices.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/xknx/remote_value/remote_value.py
+++ b/xknx/remote_value/remote_value.py
@@ -186,7 +186,7 @@ class RemoteValue(ABC, Generic[DPTPayloadType, ValueType]):
             decoded_payload = self.from_knx(_new_payload)
         except (ConversionError, CouldNotParseTelegram) as err:
             logger.warning(
-                "Can not process %s \n for %s - %s: %s",
+                "Can not process %s for %s - %s: %s",
                 telegram,
                 self.device_name,
                 self.feature_name,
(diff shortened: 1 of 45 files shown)


xknx/remote_value/remote_value.py
```diff
@@ -182,21 +180,13 @@ async def process(self, telegram: Telegram, always_callback: bool = False) -> bo
...
-            )
+
         try:
+            _new_payload = self.payload_valid(telegram.payload.value)
             decoded_payload = self.from_knx(_new_payload)
-        except ConversionError as err:
+        except (ConversionError, CouldNotParseTelegram) as err:
             logger.warning(
-                "Can not process %s for %s - %s: %s",
+                "Can not process %s \n for %s - %s: %s",
```
marvin-w, line 189: Do we need the newline here? I'd generally try to avoid newlines in log outputs.

marvin-w (commented):
Good idea, just some minor comment for now.

xknx/remote_value/remote_value.py
farmio: Oh... I was thinking about adding even more (not here but elsewhere. With proper indentation do our logs become better readable. 

Here the logged Telegram repr is probably wrapping a line.

farmio: I've reverted this in c41c2a8

farmio:
Issue reported in https://github.com/home-assistant/core/issues/59170#issuecomment-1029252173

PR #934, opened 2022-04-28
Unicast discovery endpoint

farmio:
## Description

- Use unicast discovery endpoint to receive SearchRespones frames
- Only send SearchRequests from one interface for each `scan()` call
- Use `ifaddr` instead of `netifaces`

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/xknx/io/gateway_scanner.py
+++ b/xknx/io/gateway_scanner.py
@@ -11,8 +11,8 @@ from functools import partial
 import logging
 from typing import TYPE_CHECKING
 
-import netifaces
-
+from xknx.exceptions import XKNXException
+from xknx.io import util
 from xknx.knxip import (
     HPAI,
     SRP,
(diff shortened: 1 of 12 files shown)


xknx/io/gateway_scanner.py
```diff
@@ -141,7 +141,7 @@ def __init__(
         tunnelling: bool | None = None,
         tunnelling_tcp: bool | None = None,
         routing: bool | None = None,
-        secure: bool | None = None,
+        secure: bool | None = False,
```
marvin-w, line 144: Can we remove None here?

farmio: The idea was if we had secure credentials, but there is a non-secure interface installed as well - we could use both (eg. the secure one is saturated).
"Automatic" doesn't support secure for now - this will be a follow-up

marvin-w (approved):

PR #935, opened 2022-04-28
Try multiple gateways in Automatic mode

farmio:
## Description

- Connect to next found interface in case of unsuccessful initial connection when using "automatic" mode
- Connect to first found interface immediately after it was discovered - while scanning for others in background until connection is successful

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/xknx/io/gateway_scanner.py
+++ b/xknx/io/gateway_scanner.py
@@ -191,7 +191,7 @@ class GatewayScanner:
         self.xknx = xknx
         self.local_ip = local_ip
         self.timeout_in_seconds = timeout_in_seconds
-        self.stop_on_found = stop_on_found or 0
+        self.stop_on_found = stop_on_found
         self.scan_filter = scan_filter
         self.found_gateways: dict[HPAI, GatewayDescriptor] = {}
         self._response_received_event = asyncio.Event()
@@ -212,6 +212,7 @@ class GatewayScanner:
                     return
                 yield gateway
         finally:
+            # cleanup after GeneratorExit or XKNXExceptions
             if not scan_task.done():
                 scan_task.cancel()
             await scan_task  # to bubble up exceptions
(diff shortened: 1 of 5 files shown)


farmio:
Dropped coverage in xknx.io.transport package comes from a fixed test that did open a socket - we should not do that unless we have a mock-server 😬

xknx/io/knxip_interface.py
```diff
@@ -139,24 +139,34 @@ async def _start(self) -> None:
...
-                gateway_ip=gateway.ip_addr,
-                gateway_port=gateway.port,
-            )
-        elif gateway.supports_routing:
-            await self._start_routing(local_ip=self.connection_config.local_ip)
+            scan_filter=self.connection_config.scan_filter,  # secure disabled by default
+        ).async_scan():
+            self._gateway_info = gateway
+            try:
+                if gateway.supports_tunnelling_tcp:
```
marvin-w, line 149: We should use `self._gateway_info` here since its set one line above or remove it if we dont need it (guess we have some logs where its used but havent looked)

xknx/io/gateway_scanner.py
```diff
@@ -182,27 +182,51 @@ class GatewayScanner:
...
-        local_ip: str | None = None,
-    ) -> list[GatewayDescriptor]:
+    async def scan(self) -> list[GatewayDescriptor]:
         """Scan and return a list of GatewayDescriptors on success."""
-        local_ip = local_ip or await util.get_default_local_ip(
+        await self._scan()
+        return list(self.found_gateways.values())
+
+    async def async_scan(self) -> AsyncGenerator[GatewayDescriptor, None]:
+        """Search and yield found gareways."""
```
marvin-w, line 205:
```suggestion
        """Search and yield found gateways."""
```

xknx/io/gateway_scanner.py
```diff
@@ -182,27 +182,51 @@ class GatewayScanner:
     def __init__(
         self,
         xknx: XKNX,
-        timeout_in_seconds: float = 4.0,
-        stop_on_found: int | None = 1,
+        local_ip: str | None = None,
+        timeout_in_seconds: float = 3.0,
+        stop_on_found: int | None = None,
```
marvin-w, line 187: Nice!

marvin-w (approved):
Cool! Works for me :)

xknx/io/knxip_interface.py
farmio: It's only used in HA diagnostics. We could move it to the break statement.

PR #943, opened 2022-05-03
Fix MAC calculation for 22-byte long secure payloads

farmio:
## Description

The calculation of b0 blocks was wrong if the data length was already a multiple of 16. This only shows on KNXIPFrames with a length of 22 bytes - which is eg. a TunnelingRequest with a 1 byte payload.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,10 @@
 
 ## Unreleased changes
 
+### Bugfixes
+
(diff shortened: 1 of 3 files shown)


marvin-w (approved):
Works fine with ETS 7.7.7 and a local 1 byte example. Thanks!

PR #951, opened 2022-05-13
Call cover callbacks periodically when moving

### LINKED ISSUE #483: Cover: Auto-stop Task
For setting a specific position for a cover that has no position address 
https://github.com/XKNX/xknx/blob/785d73b850a97a688a3b1a5e8baf34e82685d1b6/xknx/devices/cover.py#L202
we currently use HA to create an updater hook to call `Cover.auto_stop_if_necessary()` periodically to stop the cover again.
https://github.com/XKNX/xknx/blob/785d73b850a97a688a3b1a5e8baf34e82685d1b6/home-assistant-plugin/custom_components/xknx/cover.py#L144

We should implement the auto-stop functionality as a Task in xknx instead of relying on HA to check for updates and call stop then. We get no benefit from doing this over HA now and it would be easier to test.

farmio:
## Description

Cover: call `device_updated_cb` periodically when cover is moving

Closes #483

## Type of change

## Checklist:

### DIFF
--- a/xknx/devices/cover.py
+++ b/xknx/devices/cover.py
@@ -9,9 +9,11 @@ It provides functionality for
 """
 from __future__ import annotations
 
+import asyncio
 import logging
 from typing import TYPE_CHECKING, Any, Iterator
 
+from xknx.core import Task
 from xknx.remote_value import (
     GroupAddressesType,
     RemoteValue,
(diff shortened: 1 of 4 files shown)


marvin-w (commented):
Looks good!

marvin-w (approved):

xknx/devices/cover.py
```diff
@@ -33,12 +33,14 @@
 logger = logging.getLogger("xknx.log")
 
 
+DEFAULT_TRAVEL_TIME_DOWN: Final = 25
+DEFAULT_TRAVEL_TIME_UP: Final = 25
```
marvin-w, line 37: where is this used?

marvin-w: We should move this to the integration if it's not used in the lib

farmio: Only in init. (I guess for a lot of tests).

marvin-w: Ah! Yeah thats fine then :)

PR #964, opened 2022-06-03
Set unit_of_measurement for DPTString to None

CanisLupus-42:
## Description

The representation of a sensor in Home Assistant depends on the unit of measurement. If a sensor has an attribute 'unit_of_measurement' it will be treated as a numerical value and therefore be shown in a graph in the history. This does not work for strings. By removing the attribute the sensor will be shown in the logbook instead / the history depiction is changed from a graph to a "state list".

Wrong:
grafik

Fixed:
grafik

It was discussed in the HA forum:
https://community.home-assistant.io/t/knx-string-sensor/427315

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/test/dpt_tests/dpt_string_test.py
+++ b/test/dpt_tests/dpt_string_test.py
@@ -90,7 +90,6 @@ class TestDPTString:
         with pytest.raises(ConversionError):
             DPTString.from_knx(raw)
 
-    
     def test_no_unit_of_measurement(self):
         """Test for no unit set for DPT 16"""
         assert DPTString.unit is None
--- a/test/dpt_tests/dpt_string_test.py
+++ b/test/dpt_tests/dpt_string_test.py
@@ -93,7 +93,4 @@ class TestDPTString:
     
     def test_no_unit_of_measurement(self):
         """Test for no unit set for DPT 16"""
-        assert DPTString.unit == None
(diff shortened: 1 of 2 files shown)


test/dpt_tests/dpt_string_test.py
```diff
@@ -89,3 +89,11 @@ def test_from_knx_wrong_parameter_length(self, raw):
         """Test parsing of KNX string with wrong elements length."""
         with pytest.raises(ConversionError):
             DPTString.from_knx(raw)
+
+    
+    def test_no_unit_of_measurement(self):
+        """Test for no unit set for DPT 16"""
+        assert DPTString.unit == None
```
farmio, line 96:
```suggestion
        assert DPTString.unit is None
```
Use identity comparison instead of equality. You can remove the not-equality checks for `""` then.

farmio (changes requested):

test/dpt_tests/dpt_string_test.py
CanisLupus-42: Change is done and comitted.

farmio:
Great! Please run `black` and do commit its changes to make CI pass.

farmio (approved):
Thank you very much 👍

PR #966, opened 2022-06-04
GroupValue_Response triggers state update

### LINKED ISSUE #965: always_callback does not trigger a state update within homeassistant
**Description of problem:**

A KNX sensor is defined with `always_callback`, but does not trigger a state update within homeassistant based on any `GroupValue_Response`. Even if it is a direct response to a `GroupValue_Read` triggered bei the `sync_state` of the sensor itself. 

grafik

The green `GroupValue_Write` are found in the `state` db in homeassistant (screenshot form the SQL db, time mismatch by 3h due to timezone), and do show up in the developer-tools/event tab, when listening to `state_changed`. Even, when the same value is sent twice, as it is expected for `always_callback`.

grafik

 `GroupValue_Response` don't show up anywhere, therefore leading to the paradoxical, that a sensor is neither changed, nor updated for many minutes, even it is set to `sync_state: expire 1` and `always_callback: true`

grafik

- [ ] using xknx standalone

**Version information:**
- xknx / Home-Assistant release with the issue: 0.21.3
- last working xknx / Home-Assistant release (if known): - 

**KNX installation:**

It does not seem hardware related, can be reproduced within HA, nevertheless also occures also with real hardware KO connected to GAs.

**Problem-relevant `configuration.yaml` entries (fill out even if it seems unimportant):**

```
knx:
  sensor:
    - name: knx_test_input
      state_address: "0/0/255"
      type: percent
      sync_state: expire 1
      always_callback: true

  expose:
    - entity_id: input_number.knx_test_output
      type: percent
      address: "0/0/255"

input_number:
  knx_test_output:
    name: knx_test_output
    initial: 50
    min: 0
    max: 100
    step: 1

```

**Diagnostic data of the config entry (only when Home Assistant is used)**

CanisLupus-42:
## Description

A sensor with the option `always_callback: true` set, triggers a state_update in Homeassistant on a `GroupValue_Response` and therefore triggers state_updates in conjunction with `sync_state`.

Fixes #965 

## Type of change

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -13,6 +13,7 @@ nav_order: 2
 - Lock sending telegrams via a Tunnel until a confirmation is received
 - Use device subclass for `device_updated_cb` callback argument type hint
 - Fix CEMI Frame Ack-request flag set wrongly
+- `GroupValue_Response` triggers state update, when `always_callback: true`
 
 ## 0.21.3 Cover updates 2022-05-17
 
--- a/test/devices_tests/sensor_test.py
+++ b/test/devices_tests/sensor_test.py
@@ -725,9 +725,10 @@ class TestSensor:
         after_update_callback.assert_called_once()
         after_update_callback.reset_mock()
 
-        # verify not called when processing read responses
+        # verify called when processing read responses
         await sensor.process(response_telegram)
-        after_update_callback.assert_not_called()
+        after_update_callback.assert_called_once()
+        after_update_callback.reset_mock()
 
     #
     # SYNC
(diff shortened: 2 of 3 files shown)


CanisLupus-42:
Ok. The failed CI pushed me towards it: This behaviour is intentional, not a bug. 
@marvin-w : Why was the decision made to explicitily not process to response telegrams?

CanisLupus-42:
As discussed in the issue #965, this would break automations based on the original intention for the `always_callback` option.

PR #967, opened 2022-06-05
Add support for separate fan speed and on/off GAs

spacegaier:
## Description

Added two new optional variables to provide GAs for fan switch. This closes the gap for fans where the steps/percentage have no influence on the on/off state. If those new GAs are not provided, then the existing behavior of controlling on/off via the fan speed remains.

The only breaking change part is that now if you turn on a fan without a switch GA, a default speed of 50% will automatically be set (unless the caller provides an explicit value of course). This default was previously set by the HA integration. See comment below reg. the reasoning behind it: https://github.com/XKNX/xknx/pull/967#discussion_r891028912

Corresponding HA integration PR: https://github.com/home-assistant/core/pull/73131

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/test/devices_tests/fan_test.py
+++ b/test/devices_tests/fan_test.py
@@ -406,9 +406,13 @@ class TestFan:
             group_address_speed_state="1/7/2",
             group_address_oscillation="1/6/1",
             group_address_oscillation_state="1/6/2",
+            group_address_switch="1/5/1",
+            group_address_switch_state="1/5/2",
         )
         assert fan.has_group_address(GroupAddress("1/7/1"))
         assert fan.has_group_address(GroupAddress("1/7/2"))
         assert not fan.has_group_address(GroupAddress("1/7/3"))
         assert fan.has_group_address(GroupAddress("1/6/1"))
         assert fan.has_group_address(GroupAddress("1/6/2"))
+        assert fan.has_group_address(GroupAddress("1/5/1"))
+        assert fan.has_group_address(GroupAddress("1/5/2"))
--- a/test/devices_tests/fan_test.py
+++ b/test/devices_tests/fan_test.py
@@ -60,7 +60,6 @@ class TestFan:
         )
 
     #
-    #
     # TEST SET SPEED
     #
     async def test_set_speed(self):
@@ -76,7 +75,6 @@ class TestFan:
             payload=GroupValueWrite(DPTArray(140)),
         )
 
-    #
     #
     # TEST SET SPEED STEP
     #
@@ -99,7 +97,6 @@ class TestFan:
         )
 
     #
-    #
     # TEST SET OSCILLATION
     #
     async def test_set_oscillation(self):
@@ -120,12 +117,13 @@ class TestFan:
         )
 
     #
-    # TEST PROCESS
+    # TEST PROCESS SPEED
     #
     async def test_process_speed(self):
         """Test process / reading telegrams from telegram queue. Test if speed is processed."""
         xknx = XKNX()
         fan = Fan(xknx, name="TestFan", group_address_speed="1/2/3")
+        assert fan.is_on is False
         assert fan.current_speed is None
 
         # 140 is 55% as byte (0...255)
@@ -134,7 +132,19 @@ class TestFan:
             payload=GroupValueWrite(DPTArray(140)),
         )
         await fan.process(telegram)
+        # Setting a speed for a fan that has no dedicated switch GA,
+        # should turn on the fan.
+        assert fan.is_on is True
         assert fan.current_speed == 55
+
+        # Now set a speed of zero which should turn off the fan.
+        telegram = Telegram(
+            destination_address=GroupAddress("1/2/3"),
+            payload=GroupValueWrite(DPTArray(0)),
+        )
+        await fan.process(telegram)
+        assert fan.is_on is False
+        assert fan.current_speed == 0
 
     async def test_process_speed_wrong_payload(self):
         """Test process wrong telegrams. (wrong payload type)."""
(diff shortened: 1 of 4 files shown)


xknx/devices/fan.py
```diff
@@ -58,11 +60,29 @@ def __init__(
         self.mode = FanSpeedMode.STEP if max_step else FanSpeedMode.PERCENT
         self.max_step = max_step
 
+        # If there is no dedicated speed GA, then the main GA of the fan serves
+        # as speed GA (and implicitely on/off as well). If however there is a dedicated GA,
+        # then the regular state GA becomes an on/off switch.
+        if group_address_speed is not None:
+            self.switch = RemoteValueSwitch(
```
farmio, line 67: just initialize it with `None` group addresses. Otherwise things like `return self.switch.initialized` would raise `AttributeError`.

farmio (changes requested):

xknx/devices/fan.py
spacegaier: Change pushed

xknx/devices/fan.py
```diff
@@ -58,11 +60,28 @@ def __init__(
         self.mode = FanSpeedMode.STEP if max_step else FanSpeedMode.PERCENT
         self.max_step = max_step
 
+        # If there is no dedicated speed GA, then the main GA of the fan serves
+        # as speed GA (and implicitely on/off as well). If however there is a dedicated
+        # speed GA, then the regular state GA becomes an on/off switch.
```
farmio, line 65: Tbh I don't think this is a good concept. The `group_address` attribute would now take GAs with different DPTs depending on wether or not the `group_address_speed` is set. ([Zen of Python](https://peps.python.org/pep-0020/))
It would be easier to understand if they would always expect the same DPT.

In HA I see 2 options then
- we can make it a breaking change and use `address` (optional) for switch and add a new required key `speed_address`
- we can add a new optional key `switch_address` and call it a day

spacegaier: In the end I would be fine with both options.
1. If we use that one, I'd say the new key `speed_address` should be optional, since I assume there can be fans that can only be switched on and off, but have not speed control.
2. Fine with me as well. I just proposed the current PR with the intention of keeping a consistent feel across the platforms (with `address` controlling the main on/off state), but I do see the DPT issue with this dynamic approach.

I will adjust the coding to go with option 2, since then we have neither a breaking change nor a dynamic DPT.

spacegaier: But they might still be able to oscillate which the fan platform would already support. But since we are not going with option 2, that point from option is moot by now 😄.

farmio: Oh dear, I seem to have no clue about fans. Please ignore my comments 🤣

test/devices_tests/fan_test.py
```diff
@@ -60,6 +60,42 @@ async def test_sync_state_address(self):
...
+        assert xknx.telegrams.qsize() == 1
+        telegram = xknx.telegrams.get_nowait()
+        assert telegram == Telegram(
+            destination_address=GroupAddress("1/2/3"),
+            payload=GroupValueWrite(DPTBinary(0)),
+        )
+
+        # Turn the fan on again this time with a provided speed, which however has no
+        # additional effect, since in the switch GA mode, the speed during switching on
+        # will not be considered since that is controlled independently in this mode
```
spacegaier, line 90: That might lead to a bit of confusion, since this behavior is not completely obvious, so needs to be documented in the HA KNX docs. Otherwise users might be surprised that in their service calls the percentage/speed gets ignored, although it completely makes sense if they configure a fan with a dedicated switch GA.

farmio: With light entities we have a similar situation - this could indeed be somewhat confusing. We don't send to a binary GA when receiving a turn_on call with brightness value there. 
But for fans there is the additional FAN.SET_PERCENTAGE service so this could be ok.

test/devices_tests/fan_test.py
```diff
@@ -60,6 +60,73 @@ async def test_sync_state_address(self):
...
+    #
+    async def test_switch_on_off(self):
+        """Test switching on/off of a Fan."""
+        xknx = XKNX()
+        fan = Fan(xknx, name="TestFan", group_address_speed="1/2/3")
+
+        # Turn the fan on via speed GA. First try without providing a speed,
+        # which will not work for fans without a switch GA.
+        await fan.turn_on()
+        assert xknx.telegrams.qsize() == 0
```
farmio, line 73: Should we just use a default value (eg. 50%) then?

xknx/devices/fan.py
```diff
@@ -93,23 +108,54 @@ def __init__(
...
+        else:
+            await self.set_speed(0)
+
+    async def set_on(self) -> None:
+        """Switch on fan."""
+        await self.switch.on()
+
+    async def set_off(self) -> None:
+        """Switch off fan."""
+        await self.switch.off()
```
farmio, line 145: I guess we don't need set_on and turn_on methods doing essentially the same.
```suggestion
    async def turn_on(self, speed: int | None = None) -> None:
        """Turn on fan."""
        if self.switch.initialized:
            await self.switch.on()
        else:
            await self.set_speed(speed)

    async def turn_off(self) -> None:
        """Turn off fan."""
        if self.switch.initialized:
            await self.switch.off()
        else:
            await self.set_speed(0)
```

test/devices_tests/fan_test.py
spacegaier: That is what the KNX HA integration does, but I left that logic there rather then moving into xknx, to not break the xknx behavior. Although having it in xknx would have one big benefit:

Currently, if you have a fan with a switch GA and `turn_on()` is called, we have to ignore the provided speed percentage value, since we do not know whether it is the default or a dedicated percentage value. If xknx were to handle it the default logic, we could support the percentage value also for switch GA driven fans, since xknx would be able to differentiate between a real percentage value (sent by the user or an HA automation) and just the default (with the later being used for purely speed GA fans, but not for switch GA ones).

farmio: Move it to xknx then! 
It's ok to break things here if we get benefits from it (even if just for readability).

spacegaier: Done

xknx/devices/fan.py
```diff
@@ -93,23 +110,49 @@ def __init__(
...
+            await self.set_speed(speed or DEFAULT_TURN_ON_PERCENTAGE)
+
+    async def turn_off(self) -> None:
+        """Turn off fan."""
+        if self.switch.initialized:
+            await self.switch.off()
+        else:
+            await self.set_speed(0)
+
+    async def set_speed(self, speed: int | None) -> None:
```
farmio, line 144: Do we need to accept `None` here?
I think it would be better to guard in the calling methods to avoid an no-op-await.

farmio (approved):
Great! Thank you very much 👍

PR #1019, opened 2022-08-23
Properly ignore repeated TunnellingRequests from UDP

farmio:
## Description

From 3/8/4 Tunnelling §2.6 Frame confirmation

> If a KNXnet/IP Tunnelling Client receives a frame with a sequence number that equals the expected sequence number then it shall reply with a TUNNELLING_ACK (Status = E_NO_ERROR) frame and process the received frame.

> If a KNXnet/IP Tunnelling Client receives a frame with a sequence number that is one less than the expected sequence number then it shall reply with a TUNNELLING_ACK (Status = E_NO_ERROR) message and discard the received frame.

> If a KNXnet/IP Tunnelling Client receives a frame with a sequence number that is not equal to the expected sequence number and not equal to one less than the expected sequence number, the KNXnet/IP Tunnelling Client shall not reply and shall discard the received frame.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -11,6 +11,7 @@ nav_order: 2
 ### Bugfixes
 
 - Correctly retry sending a TunnellingRequest if no TunnellingAck was received for the first time for UDP tunnelling connections.
+- Ignore repeated TunnellingRequests received from UDP tunnelling connections.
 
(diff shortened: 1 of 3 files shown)


marvin-w (approved):
Looks good! Thanks! Also thanks for fixing the conflicts ;).

PR #1063, opened 2022-11-02
Use upper case L for liter

elupus:
This matches unit used in home assistant.

## Description

The unit used for liters in home assistant is upper case L. So let's provide that as default.

## Type of change

## Checklist:

### DIFF
--- a/xknx/dpt/dpt_2byte_float.py
+++ b/xknx/dpt/dpt_2byte_float.py
@@ -244,21 +244,21 @@ class DPTPower2Byte(DPT2ByteFloat):
 
 
 class DPTVolumeFlow(DPT2ByteFloat):
-    """DPT 9.025 DPT_Value_Volume_Flow (l/h)."""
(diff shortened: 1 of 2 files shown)


farmio:
🤔 should we attempt to move device class definitions and units (overrides as they are defined in KNX specs as well) to the HA integration so such changes can be done in batches in the core repo?

We could also add "convenience types" to eg. have state_class set. 
cc @marvin-w

elupus:
Yup, we likely should have the mapping of this inside home assistant instead of in the KNX library. Do you feel we should stick to KNX spec here and try to work around it inside home assistant for now? I don't think we have the arch in place at the moment?

farmio:
Nah, we can always revert back to Knx spec specifics later when we have such thing in place. 

I'm just preparing a bugfix PR - we can release this probably next week then.

marvin-w:
I don't see any issue with this PR. We don't use this anywhere anyway except HA.

marvin-w (approved):
Looks good!

PR #1096, opened 2022-12-06
Decouple CEMI parsing from KNXIPFrame parsing

farmio:
## Description

- Decouple KNXIPFrame parsing from CEMIFrame parsing. TunnellingRequest and RoutingIndication now carry the raw cemi frame payload as bytes. This allows decoupled CEMIFrame parsing at a later time (in Interface class rather than in KNXIPTransport class).
- Remove some tests for RoutingIndication that actually tested CEMIFrame / Telegram functions.

We are now able to inject data into the cemi parser more easily without having to pass an xknx instance throughout the IPFrame parsers. This paves the way to pass DataSecure keys.
It would also be an option to add a separate CEMI-Handler in front of `Tunnel / Routing` classes - eg. for USB support (@kistlin may be interested).

A noticeable change of this is that the `xknx.knx` logger now doesn't show the CEMI contents anymore, but the hex representation of the raw bytes 😬
```log
<KNXIPFrame <KNXIPHeader HeaderLength="6" ProtocolVersion="16" KNXIPServiceType="TUNNELLING_REQUEST" Reserve="0" TotalLength="23" />
 body="<TunnellingRequest communication_channel_id="2" sequence_counter="34" cemi="2900bcd0111529150300800c5b" />" />
```

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -8,6 +8,10 @@ nav_order: 2
 
 # Unreleased changes
 
+### Internal
+
+- Decouple KNXIPFrame parsing from CEMIFrame parsing. TunnellingRequest and RoutingIndication now carry the raw cemi frame payload as bytes. This allows decoupled CEMIFrame parsing at a later time (in Interface class rather than in KNXIPTransport class) for better error handling and upcoming features.
+
 # 2.1.0 Enhance notification device
 
 ### Devices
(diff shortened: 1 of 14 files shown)


marvin-w (approved):
Looks good, as mentioned maybe introducing xknx.cemi makes sense, but we can do this in a follow-up as well :).

PR #1174, opened 2023-02-04
Fix wrong string length in keyfile signature verification

envy:
## Description

Attribute values during keyfile signature verification are prefixed with their length. However this was wrong, when the UTF-8 encoded string contained multi-byte characters. This fixes that. Keyfiles that contain project names with umlauts (äüö) should pass signature verification now.

Fixes # (issue)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/xknx/secure/keyring.py
+++ b/xknx/secure/keyring.py
@@ -414,12 +414,12 @@ class KeyringSAXContentHandler(ContentHandler):
 
     def append_string(self, value: str | bytes) -> None:
         """Append a string to a byte array for signature verification."""
-        self.output.append(len(value))
 
         if isinstance(value, str):
-            self.output.extend(value.encode("utf-8"))
-        if isinstance(value, bytes):
-            self.output.extend(value)
+            value = value.encode("utf-8")
+
+        self.output.append(len(value))
+        self.output.extend(value)
 
 
 def verify_keyring_signature(path: str, password: str) -> bool:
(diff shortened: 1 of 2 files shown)


xknx/secure/keyring.py
```diff
@@ -414,11 +414,12 @@ def endElement(self, name: str) -> None:
 
     def append_string(self, value: str | bytes) -> None:
         """Append a string to a byte array for signature verification."""
-        self.output.append(len(value))
 
         if isinstance(value, str):
+            self.output.append(len(value.encode("utf-8")))
             self.output.extend(value.encode("utf-8"))
```
farmio, line 420:
```suggestion
            byte_value = value.encode("utf-8")
            self.output.append(len(byte_value))
            self.output.extend(byte_value)
```
We should call `encode` only once.

xknx/secure/keyring.py
```diff
@@ -414,11 +414,12 @@ def endElement(self, name: str) -> None:
...
         if isinstance(value, bytes):
+            self.output.append(len(value))
             self.output.extend(value)
```
farmio, line 423:
```suggestion
        if isinstance(value, str):
            value = value.encode("utf-8")

        self.output.append(len(value))
        self.output.extend(value)
```
Or make it even shorter 😄

marvin-w (approved):
Nice, looks good :).

Adding a test would probably make sense if possible.

farmio:
We'd have to change one of the keyfiles to really test that. Otherwise we'd not have a signature to test against. I could add an Umlaut to the keyfile added in #1167

farmio (approved):
Thank you very much 🎉

PR #1199, opened 2023-02-16
Support different connection types in `ConnectRequest` and `ConnectResponse` frames

### LINKED ISSUE #1197: Support for KNXnet/IP Device Management Connection
Currently `ConnectRequest` only supports tunnelling (basic or extended tunnel CRI), would it be possible to add support for opening device management connections?

pdgendt:
## Description

This PR adds support for a Device Management Connection to an KNXnet/IP server.

Fixes #1197

## Type of change

## Checklist

### DIFF
--- a/xknx/io/tunnel.py
+++ b/xknx/io/tunnel.py
@@ -19,6 +19,7 @@ from xknx.exceptions import (
 )
 from xknx.knxip import (
     HPAI,
+    ConnectRequestInformation,
     DisconnectRequest,
     DisconnectResponse,
     HostProtocol,
@@ -190,7 +191,7 @@ class _Tunnel(Interface):
         connect = Connect(
             transport=self.transport,
             local_hpai=self.local_hpai,
-            individual_address=self._requested_address,
+            cri=ConnectRequestInformation(individual_address=self._requested_address),
         )
         await connect.start()
         if connect.success:
--- a/xknx/io/tunnel.py
+++ b/xknx/io/tunnel.py
@@ -203,7 +203,7 @@ class _Tunnel(Interface):
                 else connect.data_endpoint.addr_tuple
             )
             # Use the individual address provided by the tunnelling server
-            self._src_address = IndividualAddress(connect.identifier)
+            self._src_address = connect.crd.individual_address or IndividualAddress(0)
             self.xknx.current_address = self._src_address
             logger.debug(
                 "Tunnel established. communication_channel=%s, address=%s",
--- a/xknx/knxip/connect_response.py
+++ b/xknx/knxip/connect_response.py
@@ -8,6 +8,7 @@ assigns a communication channel and an individual address for the client.
 from __future__ import annotations
 
 from xknx.exceptions import CouldNotParseKNXIP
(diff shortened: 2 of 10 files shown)


xknx/knxip/connect_response.py
```diff
@@ -90,6 +68,79 @@ def __repr__(self) -> str:
...
+class ConnectResponseData:
+    """Representation of a KNX Connect Response Data block (CRD)."""
+
+    CRD_LENGTH = 2
+    CRD_TUNNEL_LENGTH = 4
+
+    def __init__(
+        self,
+        request_type: ConnectRequestType = ConnectRequestType.TUNNEL_CONNECTION,
+        identifier: int = 0,
```
farmio, line 84: Since this is only used for tunnelling, I guess it always represents an `IndividualAddress`. Then we could just use that type instead of `int` here.

pdgendt: Updated the PR.

xknx/io/tunnel.py
```diff
@@ -202,7 +203,7 @@ async def _connect_request(self) -> bool:
                 else connect.data_endpoint.addr_tuple
             )
             # Use the individual address provided by the tunnelling server
-            self._src_address = IndividualAddress(connect.identifier)
+            self._src_address = connect.crd.individual_address or IndividualAddress(0)
```
pdgendt, line 206: Not sure about the `or IndividualAddress(0)` but `connect.crd.individual_address` can be `None`.

farmio: Ah, in theory it can, in practice it can't. 
When we receive a non-tunneling CRD here it should probably raise some error. And a tunneling-CRD always has an identifier (defaulted to 0 before). 

Not sure how to implement this the best way. Maybe we can do a function/property in CRD that returns the individual address or raises when there is none.

pdgendt: Change it like this?
```suggestion
            if connect.crd.individual_address is None:
                raise CommunicationError("No individual address.")
            self._src_address = connect.crd.individual_address
```

pdgendt:
@farmio I don't know why the CI keeps failing, any suggestions?

farmio:
Use `black` 😉
Or our `pre-commit` hooks (which invoke black).

xknx/io/tunnel.py
farmio: Right... we should then also close it 🫤
I guess that's a lot of code for a non-issue...

farmio: Maybe just revert to defaulting to 0. If this ever yields problems it can be worked out then.

pdgendt: Right, I didn't change it yet so it's fine. The behavior is also the same as before the PR.

farmio:
Oh look at that! The feature I wanted to implement next - changing the tunnel IA for UDP connections - actually need this 😃
I always thought this would use a transport layer connection 🙃
Bildschirm­foto 2023-02-16 um 17 11 11

farmio (approved):
Thank you very much!

Issue #1155, opened 2023-01-22
Light: last used mode


farmio:
Maybe we can find a way to store the last used mode of lights supporting white with color temperatures and also colored light. 
Like a light that can be controlled by HSB and also tunable white DPT 5 or 7 GAs.

A possible way would be to store the latest non-state addresses that received a GroupValueWrite. This could be mapped to a color mode in the HA integration. 

We'd need to find out if there is consistent behaviour among different actuator models when they eg. have an internal defined turn-on state. So if only a `address` DPT 1 is received and it switches its internal mode to something different than what was chosen before there should probably be some writes on state_addresses. In this case we shouldn't only observe non-state addresses.

This would enable us to support multiple modes in a single light entity in HA. Currently 2 independent entitites are needed for such a setup - but that can lead to issues with duplicated unique_ids.

PR #1202, opened 2023-02-20
Add support of color temperature brightness for lights

sthade:
## Description

This pull requests add support for color_temperature_brightness which some KNX dimmers have
additional to the brightness. In this case changing the brightness will bring the KNX dimmer into the color mode and changing the color_temperature_brightness switches the dimmer into the color_temperature mode.
So, this change is about adding a pair of additional group addresses, not changing the existing functions.

Target is to get full support for the KNX dimmers which support a mixed mode of RGBW+CT/TW.

The modifications for home assistant are also already written and tested (Incl. multi-mode KNX light),
but there is no chance of a pull request in home assistant if the library below (xknx) don't support the color_temperature_brightness.

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,10 @@ nav_order: 2
 
 - When `ConnectionConfig.individual_address` is set and a Keyring is given `ConnectionType.AUTOMATIC` will try to connect to the host of this address. If not found (in keyfile or discovery) it will raise.
 
+### Feature added
+
+- Support for color temperature brightness added. This is needed for dimmers having additionaly to the brightness a seperate brightness for the color temperature.
+
 # 2.5.0 Request IA 2023-02-14
 
 ### Connection
--- a/docs/light.md
+++ b/docs/light.md
@@ -33,6 +33,8 @@ The Light object is either a representation of a binary or dimm actor, LED-contr
 - `group_address_tunable_white_state` KNX group address for the current relative color temperature. *DPT 5.001*
 - `group_address_color_temperature` KNX group address to set absolute color temperature. *DPT 7.600*
 - `group_address_color_temperature_state` KNX group address for the current absolute color temperature. *DPT 7.600*
+- `group_address_color_temperature_brightness` KNX group address to set absolute color temperature brightness. *DPT 5.001*
+- `group_address_color_temperature_brightness_state` KNX group address for the current color temperature state. *DPT 5.001*
 
 - `group_address_switch_red` KNX group address to switch the red component. *DPT 1.001*
 - `group_address_switch_red_state` KNX group address for the state of the red component. *DPT 1.001*
@@ -71,7 +73,9 @@ light = Light(xknx,
               group_address_tunable_white='1/2/9',
               group_address_tunable_white_state='1/2/10',
               group_address_color_temperature='1/2/11',
-              group_address_color_temperature_state='1/2/12')
+              group_address_color_temperature_state='1/2/12',
+              group_address_color_temperature='1/2/13',
+              group_address_color_temperature_state='1/2/14')
 
 # Switching light on
 await light.set_on()
@@ -94,12 +98,16 @@ await set_tunable_white(25)
 # Set absolute color temperature (Kelvin)
 await set_color_temperature(3300)
 
+# Set color temperature brightness
+await light.set_color_temperature_brightness(23)
+
 # Accessing light via 'do'
 await light.do('on')
 await light.do('off')
 await light.do('brightness:80')
 await light.do('tunable_white:75')
 await light.do('color_temperature:5000')
+await light.do('color_temperature_brightness:80')
 
 # Update current state via KNX GroupValueRead
 await light.sync(wait_for_result=True)
@@ -115,6 +123,8 @@ print(light.supports_tunable_white)
 print(light.current_tunable_white)
 print(light.supports_color_temperature)
 print(light.current_color_temperature)
+print(light.supports_color_temperature_brightness)
+print(light.current_color_temperature_brightness)
 ```
 
 ## [](#header-2)Example: RGBW light with individual group addresses for red, green, blue and white
--- a/test/devices_tests/light_test.py
+++ b/test/devices_tests/light_test.py
@@ -275,6 +275,26 @@ class TestLight:
         assert not light.supports_color_temperature
 
     #
+    # TEST SUPPORT COLOR TEMPERATURE BRIGHTNESS
+    #
+    def test_supports_color_temperature_brightness_true(self):
+        """Test supports_color_temperature_brightness with a light with color temperature brightness function."""
+        xknx = XKNX()
+        light = Light(
+            xknx,
+            "Diningroom.Light_1",
+            group_address_switch="1/6/4",
+            group_address_color_temperature_brightness="1/6/6",
+        )
+        assert light.supports_color_temperature_brightness
+
+    def test_supports_color_temperature_brightness_false(self):
+        """Test supports_color_temperature_brightness attribute with a Light without color temperature brightness function."""
+        xknx = XKNX()
+        light = Light(xknx, "Diningroom.Light_1", group_address_switch="1/6/4")
+        assert not light.supports_color_temperature_brightness
+
+    #
     # SYNC
     #
     async def test_sync(self):
@@ -290,9 +310,10 @@ class TestLight:
             group_address_xyy_color_state="1/2/4",
             group_address_tunable_white_state="1/2/7",
             group_address_color_temperature_state="1/2/8",
+            group_address_color_temperature_brightness_state="1/2/10",
             group_address_rgbw_state="1/2/9",
         )
-        expected_telegrams = 7
+        expected_telegrams = 8
 
         await light.sync()
         assert xknx.telegrams.qsize() == expected_telegrams
@@ -319,6 +340,9 @@ class TestLight:
             ),
             Telegram(
                 destination_address=GroupAddress("1/2/8"), payload=GroupValueRead()
+            ),
+            Telegram(
+                destination_address=GroupAddress("1/2/10"), payload=GroupValueRead()
             ),
         ]
         assert len(set(telegrams)) == expected_telegrams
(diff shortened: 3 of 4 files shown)


farmio:
Hi 👋! 
Thank you for your contribution!

What devices support this additional brightness object?
How is it different from the current `group_address_tunable_white` attribute?

farmio:
Looking at your [commit](https://github.com/home-assistant/core/commit/13afa42ec6ca0d0ab0c7318f940513b3913a3aad) it seems that `color_mode` is updated whenever it is set from HA (`async_turn_on()`), but it isn't updated when changed from the KNX bus. 
Would it make sense to move that `color_mode` attribute to xknx and update it internally (eg. in `process_group_write()`)?

sthade:
Hello,

the device I like to support is an MDT AKD-0424V.02 see:
https://www.mdt.de/produkte/produktdetail/aktorik/dimmaktoren/led-controller-akd-up.html

For the tuneable white (TW) / correlated color temperature (CCT) feature there are 3 absolute values to write:
1. LED TW Colortemperatur (%) Absolute (DPT 5.001)
2. LED TW Colortemperatur (Kelvin) Absolute (DPT 7.600)
3. LED TW Brightness Absolute (DPT 5.001)

And of course there are 3 status objects too.

For the RGBW color mode there are different communication objects offered incl. a separate value / brightness object. Depending which color or brightness signal will be sent to the device it is switching to the color mode or tunable white mode.

The color temperature in percent or in Kelvin are normally not used at the same time / alternatively used.

Let’s have a short look how home assistant is using it:
```
knx:
  light:
  - name: My RGBW+TW light
    address: 1/2/1
    state_address: 1/2/2
    brightness_address: 1/2/3
    brightness_state_address: 1/2/4
    rgbw_address: 1/2/5
    rgbw_state_address: 1/2/6
    color_temperature_address: 1/2/7
    color_temperature_state_address: 1/2/8
    color_temperature_mode: relative
```
Based on the “color_temperature_mode” the group address from “color_temperature_address” is given to the xknx objects: (see the following home assistant code)
```
    if config[LightSchema.CONF_COLOR_TEMP_MODE] == ColorTempModes.ABSOLUTE:
        group_address_color_temp = config.get(LightSchema.CONF_COLOR_TEMP_ADDRESS)
        group_address_color_temp_state = config.get(
            LightSchema.CONF_COLOR_TEMP_STATE_ADDRESS
        )
    elif config[LightSchema.CONF_COLOR_TEMP_MODE] == ColorTempModes.RELATIVE:
        group_address_tunable_white = config.get(LightSchema.CONF_COLOR_TEMP_ADDRESS)
        group_address_tunable_white_state = config.get(
            LightSchema.CONF_COLOR_TEMP_STATE_ADDRESS
        )
```
So, the existing configuration is only covering two out of the three communication objects for the TW feature.
What I need is a home assistant configuration like:
```
knx:
  light:
  - name: My RGBW+TW light
    address: 1/2/1
    state_address: 1/2/2
    brightness_address: 1/2/3
    brightness_state_address: 1/2/4
    rgbw_address: 1/2/5
    rgbw_state_address: 1/2/6
    color_temperature_address: 1/2/7
    color_temperature_state_address: 1/2/8
    color_temperature_brightness_address: 1/2/9
    color_temperature_brightness_state_address: 1/2/10
    color_temperature_mode: relative
```
Which I realized on my local installation and try to bring to the projects.

Of cause the naming which I propose for the new group address “color_temperature_brightness” is not perfect, because someone who is using the kelvin mode (which is here called absolute) would like to have it named “tunable_white_brightness”, but with the current naming there is no good alternative.
 
Please allow me a note on the naming:
In "KNX world" there are terms which looks like not applied at that position in xknx:
- Absolute normally means a fixed target value (go to brightness 213)
- Relative normally means a delta to be applied (apply -10, means current value 213-10=203)

Here this is used to switch between the percentage and kelvin setting, which are all absolute values.
In a clean implementation I would name the group addresses for TW:
```
tw_color_temperature_kelvin   (instead of color_temperature)
tw_color_temperature_percent  (instead of tunable_white)
tw_brightness

```
(And the states according to it)
Alternatively following the naming in home assistant, also the following naming might be ok (but not perfect):
```
color_temperature_kelvin      (instead of color_temperature)
color_temperature_percent     (instead of tunable_white)
color_temperature_brightness

```
But this would be a breaking change… So that is why I proposed the naming which is close to the home assistant naming.

(The same problem is in home assistant with the relative / absolute mode switch, which should be kelvin / percent, but this is a different project I know…)

sthade:
Regarding the idea get the color_mode from the KNX status objects. I would love to that the status from a state communication object, unfortunately the dimmer I'm using is not giving any information about if he is in color mode or tunable white mode.

This is by the way the same for other devices. My Philips HUE light bulbs have the same issue in HA. If they are switched on with a tunable white color, HA is showing the color selector and not the color temperature selector. So, the mode is also not preserved.

But please note that in daily operation there is no issue with it.

That’s why I implemented the mode switch in HA in this way.

farmio:
Aha, interesting - this is a 4 channel controller providing a 5 channel function 🙃. I had a look at Enertex 5-channel Led sequencer application and it seems to also have a dedicated TW-brightness object.

> Alternatively following the naming in home assistant, also the following naming might be ok (but not perfect):
color_temperature_kelvin      (instead of color_temperature)
color_temperature_percent     (instead of tunable_white)
color_temperature_brightness

I like this as `tw_color_temperatur` kind of describes the function doubly. I'm ok with changing names of these in xknx to more reasonable names, but in HA I think breaking users configs just for the sake of it is not a good idea. 


Still the mode should also be updated when changed from outside HA.

farmio:
> unfortunately the dimmer I'm using is not giving any information about if he is in color mode or tunable white mode.

thinking about that, we don't even have to rely on the last received status GA. We could just act on the last received command-GA. If it was a color-temperature thing (kelvin or brightness) it would be mode TW, if it was a color GA it would be in a color mode.
I don't know how these devices behave when eg. in TW mode and the "normal" brightness GA is written to. Or what happens when just turned on via DPT 1 (I guess that's configurable?).

sthade:
Regarding the naming:
OK, then I will update the pull request and will use the color_temperature_kelvin, color_temperature_percent, color_temperature_brightness naming. Which is also consistent with the HA naming.
(But I disagree that tw_color_temperature_relative would be a double naming, tw is the mode, color_temperature the function)

Regarding the mode:
I checked the manual of the "Enertex KNX LED Dimmsequenzer 20A/5x DK" and there is also no status were a mode can easy derived. So, there will be no direct way to get the mode as status.
This results in the options:
1.	Don’t track the status. Which means that the sliders for controlling brightness should be arranged differently in the UI and it would be a perfect solution. But it’s not an option for HA, as this is only needed for KNX devices and HA support different devices which can handle this.
2.	Handle the status in HA. This is what I proposed, which works without issues as long as triggering the done by KNX or HA. If switching on from a KNX button and continuing in HA, there might be an unexpected change from TW to RBG (or vice versa) if changing the brightness. Finally, the result is not too bad and in most cases it works fine.
3.	Handle the mode in the xknx. Your proposal #1155 to get the status from the group writes is also a good option. There are two situations where this might not work correctly.
a) As you told if the device after switch on has a different state than when switched of.
b) The other situation is a restart of xknx, in this situation the status cannot read from the KNX bus.
This leads to the situation that also in few situations the might be an unexpected mode change if changing the brightness. But overall, it will work well.

So, #2 and #3 have some limitations. But #3 can be improved:

With my MDT dimmer I tried a way to understand the status after switch on from the status, reading the write communication objects, … but it didn’t find a way to get the mode. From the colors it’s also not possible to conclude the mode, as even a color which means TW can be set in color mode.

The only stable solution I can imagine is to let the user set the behaviour of his device by configuration. Like:
turnon_color_mode_hint: LAST_MODE | COLOR_TEMPERATURE | RGB | RGBW | HSV

E.g.  If I configured my device to always start with a cold white in TW.
I will set turnon_color_mode_hint: COLOR_TEMPERATURE

E.g.  If I configured my device to always start with a nice light green in RGBW.
I will set turnon_color_mode_hint: RGBW

By this there will be no issue with the mode after turning on the device. Of cause if the device is reconfigured the configuration might need to change too.

As default I would go for LAST_MODE

This solution would be also extendable. If a good algorithm to determine the color code would be found the option AUTO can be added (and get the default). Or if solutions are found for a group of of KNX devices, also a fitting option can be added.

So #3 with the turnon_color_mode_hint will work best in my opinion.

farmio:
> OK, then I will update the pull request and will use the color_temperature_kelvin, color_temperature_percent, color_temperature_brightness naming. Which is also consistent with the HA naming.
> (But I disagree that tw_color_temperature_relative would be a double naming, tw is the mode, color_temperature the function)

Ah, I see. Well maybe we should just leave it as is and not introduce any breaking. These names `group_address_color_temperature_state` etc. are already quite unwieldy - making them even longer isn't really solving a problem. I guess they could be renamed at some point when already heading for a breaking change release in a separate PR.

Ad 3a: does your device always report both, kelvin and rbg status, on turn-on?

> as even a color which means TW can be set in color mode.

That is probably due to it being a 4-channel controller, isn't it? 

Going for 2 for a start is the safest option imho. Turning to 3 - with its added configuration complexity, could always be done later.

sthade:
Behavior on turn on of the MDT device is configurable.

If in TW mode and a normal (value) brightness arrive the dimmer switch to color mode. Same for RGB or RGBW signals.

If in color mode and a TW brightness or TW temp is received, it is switching to the tunable white mode.

As written in the previous comment, getting the mode from the command to group address, plus the hint what is configured for the KNX device might be a solution.

sthade:
The color mode is only used in the UI to my knowledge. But even for devices where the color mode is saved (Philips HUE) the UI of HA do not jump to the right mode (select the correct tab)

So the only need for the mode is know to which GA to sent the brightness.

As the effect, if one time wrong,  is low. I'm still fine with #2.

farmio:
I'd suggest to go with # 2 . We don't have to implement all ideas in a single HA release. Let's see how this works out with users having different devices like DALI gateways etc.

farmio:
From the first preview, it seems that color_mode isn't that important for the new light cards anymore. <https://rc.home-assistant.io/blog/2023/02/22/release-20233/#new-dialogs-for-lights-switches-and-siren-entities>

farmio:
Hi 👋! Can you make CI pass?

sthade:
Hello, sorry for being inactive for some time. I will make the CI pass and still plan get all in xknx and home assistant, but I need some more days, as some unexpected family issue occurred which currently limiting my free time.

farmio:
Hi @sthade 👋! I hope your family is alright.
Are you still planning on finishing this PR and the changes in HA core?

PR #1210, opened 2023-03-01
Support Device Management CEMI frames

pdgendt:
## Description

This PR adds classes/types used to send and receive Device Configuration Requests. 
According to the KNX specifications, these requests use CEMI M_Prop type frames for reading/writing device properties.

## Type of change

## Checklist


xknx/profile/const.py
```diff
@@ -0,0 +1,317 @@
...
+
+
+class ResourceTextCataloguePropertyId(ResourcePropertyId):
+    """Extended class for Text Catalogue Object (Object Type 16) Property Ids."""
+
+    PID_LOCALE_LIST = 51  # PDT_GENERIC_04[]
+    PID_LOCALE_SELECTION = 52  # PDT_GENERIC_04
+    PID_ACTIVE_LOCALE = 53  # PDT_GENERIC_04
+    # PID_STRING_001 to PID_STRING_141 is [60-200]
+    PID_STRING_001 = 60  # PDT_UTF_8
```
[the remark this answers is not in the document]
farmio: 🤔 something like this comes to my mind, but if that is any better than adding them explicitly I don't know
```py
Enum("PidString", [f"PID_STRING_{num}" for num in range(1, 142)], start=60)
```

xknx/cemi/cemi_frame.py
```diff
@@ -195,58 +220,459 @@ def from_knx_data_link_layer(cemi: bytes, code: CEMIMessageCode) -> CEMIFrame:
...
+        return CEMIMPropInfo.LENGTH + len(self.data)
+
+    def to_knx(self) -> bytes:
+        """Serialize to CEMI raw data."""
+        return self.property_info.to_knx() + self.data
+
+    @classmethod
+    def from_knx(cls, raw: bytes) -> CEMIData:
+        """Parse/deserialize from KNX/IP raw data."""
+        if len(raw) < 7:
```
farmio, line 378:
```suggestion
        if len(raw) <= CEMIMPropInfo.LENGTH:
```
We should use constants for readability when possible.
You could also define a MIN_LENGTH class attribute if you prefer.

xknx/cemi/cemi_frame.py
```diff
@@ -195,58 +220,459 @@ def from_knx_data_link_layer(cemi: bytes, code: CEMIMessageCode) -> CEMIFrame:
...
+        _data = (
+            f'data="{self.data.hex()}" '
+            if self.property_info.number_of_elements != 0
+            else ""
+        )
+        return (
+            f"CEMIMPropReadResponse("
+            f"{self.property_info._property_repr()}"
+            f'error_code="{self.error_code}" '
+            f"{_data} )"
```
farmio, line 404:
```suggestion
            f"{_data})"
```
`_data` already has ends with a space

xknx/cemi/cemi_frame.py
```diff
@@ -13,150 +13,176 @@
...
+    def from_knx(cls, raw: bytes) -> CEMIData:
+        """Parse/deserialize from KNX/IP raw data."""
+
+    @abstractmethod
+    def __repr__(self) -> str:
+        """Return object as readable string."""
+
+    def __eq__(self, other: object) -> bool:
+        """Equal operator."""
+        return self.__dict__ == other.__dict__
```
farmio, line 83:
```suggestion
    def __eq__(self, other: object) -> bool:
        """Equal operator."""
        if self.__class__ != other.__class__:
            return False
        return self.__dict__ == other.__dict__
```
Otherwise eg. `CEMIMPropReadResponse` would equal `CEMIMPropWriteRequest` with same data and property_info.

xknx/cemi/cemi_frame.py
```diff
@@ -195,58 +220,459 @@ def from_knx_data_link_layer(cemi: bytes, code: CEMIMessageCode) -> CEMIFrame:
...
-            npdu_len = self.payload.calculated_length()
+        """Serialize to CEMI raw data."""
+        return (
+            self.object_type.value.to_bytes(2, "big")
+            + self.object_instance.to_bytes(1, "big")
+            + self.property_id.to_bytes(1, "big")
+            + ((self.number_of_elements << 12) + self.start_index).to_bytes(2, "big")
+        )
 
+    def _property_repr(self) -> str:
```
farmio, line 307: why not use `__str__` or `__repr__`?
We could use f"{self.property_info}" instead of `f"{self.property_info._property_repr()}"` below then.

xknx/knxip/device_configuration_request.py
```diff
@@ -0,0 +1,64 @@
...
+from xknx.exceptions import CouldNotParseKNXIP
+
+from .body import KNXIPBody
+from .knxip_enum import KNXIPServiceType
+
+
+class DeviceConfigurationRequest(KNXIPBody):
+    """Representation of a KNX Device Configuration Request."""
+
+    SERVICE_TYPE = KNXIPServiceType.DEVICE_CONFIGURATION_REQUEST
```
farmio, line 13: I think this (and the ACK) could share a common parent class with `TunnellingRequest`. They are exactly the same (when using `self.__class__.__name__` in `__repr__`).
But I'm fine with this as well, its just some duplicate code - we have a lot of that 😬

farmio (changes requested):
Thanks for your contribution 👍

xknx/secure/data_secure.py
```diff
@@ -131,12 +131,15 @@ def check_sequence_number(
 
     def received_cemi(self, cemi: CEMIFrame) -> CEMIFrame:
         """Handle received CEMI frame."""
+        # No data frame
+        if not isinstance(cemi.data, CEMILData):
```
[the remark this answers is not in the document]
pdgendt: A deep copy is taken somewhere down the road, I wasn't comfortable changing it this way. Maybe this can be done in a separate PR?

xknx/knxip/device_configuration_request.py
pdgendt: How can a "private" parent class be created in python? But we could move that to a new PR.

xknx/secure/data_secure.py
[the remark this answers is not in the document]
farmio: Sure 👍

xknx/knxip/device_configuration_request.py
farmio: Yeah, a separate PR would be fine. 
I guess technically there is no "private class", but it can be prepended with `_` and not assign a `SERVICE_TYPE`. Maybe even not inherit from `KNXIPBody`, but use it as Mixin class.

farmio (approved):
Very nice, thank you! 👍

PR #1260, opened 2023-05-02
Add support for Tunnelling Feature service messages

pdgendt:
## Description

This PR adds support for (de)serialization of Tunnelling Feature requests and responses. 

## Type of change

## Checklist

### DIFF
--- a/xknx/knxip/knxip_enum.py
+++ b/xknx/knxip/knxip_enum.py
@@ -167,3 +167,24 @@ class SearchRequestParameterType(Enum):
     SELECT_BY_SERVICE = 0x03
     # The Client shall include this SRP to indicate that it is interested in the listed DIBs
     REQUEST_DIBS = 0x04
+
+
+class TunnellingFeatureType(Enum):
+    """Enum class for KNX/IP Tunnel Features services."""
+
+    # Getting the supported EMI type(s) (only cEMI allowed)
+    SUPPORTED_EMI_TYPE = 0x01
+    # Getting the local Device Descriptor Type 0 for possible local device management
+    DEVICE_DESCRIPTOR_TYPE_0 = 0x02
+    # Getting and informing on the bus connection status
+    BUS_CONNECTION_STATUS = 0x03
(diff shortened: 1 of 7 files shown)


xknx/knxip/knxip_enum.py
```diff
@@ -167,3 +167,26 @@ class SearchRequestParameterType(Enum):
     SELECT_BY_SERVICE = 0x03
     # The Client shall include this SRP to indicate that it is interested in the listed DIBs
     REQUEST_DIBS = 0x04
+
+
+class TunnellingFeatureType(Enum):
+    """Enum class for KNX/IP Tunnel Features services."""
+
+    # Invalid placeholder
+    INVALID = 0x00
```
farmio, line 176: is this used in the KNX specs too?

farmio (commented):
LGTM so far 👍
Please add some tests and we are good to go!

xknx/knxip/knxip_enum.py
pdgendt: It's not in the specs, I added it to have a default argument for init functions. I can remove it and make the argument required if that's preferred?

farmio: Making it required doesn't really work with the current implementation of KNXIPFrame. See https://github.com/XKNX/xknx/blob/bf69e38792d61ec2e84c8ec175c3a61a3f20c4c6/xknx/knxip/knxip.py#L122-L123
This would require a quite big refactoring - like done for APCI (see https://github.com/XKNX/xknx/pull/1154) but that is out of scope.

You can leave it there and maybe add a check to raise if it was encountered in `to_knx` or `from_knx` (just like any other, not listed enum-item would do). 
Or just use a different `TunnellingFeatureType` as default (we do this in other KNXIPBody arguments too (`sequence_counter`, `communication_channel_id`, etc.) - not beautiful, but works fine 😬

pdgendt: I will change the default.

farmio (approved):
LGTM! Thank you very much! 😀

PR #1395, opened 2024-01-26
Serial number-based individual address read/write

opieters:
## Description

Adds two new management procedures: `nm_individual_address_serial_number_read` and `nm_individual_address_serial_number_write`, which can be used to set and read the individual address based on the unique KNX serial number (if available).

Fixes # (issue)

## Type of change

## Checklist


xknx/management/procedures.py
```diff
@@ -184,3 +184,56 @@ async def nm_invididual_address_write(
...
+    """Write individual address to device with specified serial number."""
+    individual_address = IndividualAddress(individual_address)
+    await xknx.management.send_broadcast(
+        Telegram(
+            destination_address=GroupAddress("0/0/0"),
+            payload=apci.IndividualAddressSerialWrite(
+                address=individual_address, serial=serial
+            ),
+        )
+    )
```
farmio, line 221: The specs say 
> The procedure shall ensure that the assigned Individual Address is unique.

, but don't have any means of doing so in the procedure. Is it expected that the management client knows and uses `nm_individual_address_check` by themselves before?

test/management_tests/procedures_test.py
```diff
@@ -594,3 +594,152 @@ async def test_nm_individual_address_write_address_found_other_in_programming_mo
...
+            address=individual_address, serial=serial_number
+        ),
+    )
+
+    await time_travel(0)
+    xknx.management.process(address_reply)
+    await time_travel(0.5 * management.management.MANAGAMENT_CONNECTION_TIMEOUT)
+    assert xknx.cemi_handler.send_telegram.call_args_list == [
+        call(read_address),
+    ]
```
farmio, line 631:
```suggestion
    await time_travel(0)
    assert xknx.cemi_handler.send_telegram.call_args_list == [
        call(read_address),
    ]
    xknx.management.process(address_reply)
```
we don't need to await the timeout when the reply was received.

test/management_tests/procedures_test.py
```diff
@@ -594,3 +594,152 @@ async def test_nm_individual_address_write_address_found_other_in_programming_mo
...
+    read_address = Telegram(
+        destination_address=GroupAddress("0/0/0"),
+        payload=apci.IndividualAddressSerialRead(serial=serial_number),
+    )
+
+    await time_travel(management.management.MANAGAMENT_CONNECTION_TIMEOUT)
+
+    assert xknx.cemi_handler.send_telegram.call_args_list == [
+        call(read_address),
+    ]
```
farmio, line 658:
```suggestion
    await time_travel(0)
    assert xknx.cemi_handler.send_telegram.call_args_list == [
        call(read_address),
    ]
    await time_travel(3)
```
the telegram is sent before any timeout - we could test that

test/management_tests/procedures_test.py
```diff
@@ -594,3 +594,152 @@ async def test_nm_individual_address_write_address_found_other_in_programming_mo
...
+
+    assert not await task
```
farmio, line 660:
```suggestion
    assert await task is None
```

test/management_tests/procedures_test.py
```diff
@@ -594,3 +594,152 @@ async def test_nm_individual_address_write_address_found_other_in_programming_mo
...
+    xknx.management.process(address_reply)
+
+    assert xknx.cemi_handler.send_telegram.call_args_list == [
+        call(write_address),
+        call(read_address),
+    ]
+
+    await time_travel(management.management.MANAGAMENT_CONNECTION_TIMEOUT)
+
+    await task
```
farmio, line 708:
```suggestion
    await time_travel(0)
    assert xknx.cemi_handler.send_telegram.call_args_list == [
        call(write_address),
        call(read_address),
    ]
    xknx.management.process(address_reply)
    await task
```
no need to wait for anything 😃

test/management_tests/procedures_test.py
```diff
@@ -594,3 +594,152 @@ async def test_nm_individual_address_write_address_found_other_in_programming_mo
...
+    await time_travel(0)
+    await time_travel(management.management.MANAGAMENT_CONNECTION_TIMEOUT)
+
+    assert xknx.cemi_handler.send_telegram.call_args_list == [
+        call(write_address),
+        call(read_address),
+    ]
+
+    with pytest.raises(ManagementConnectionError):
+        await task
```
farmio, line 745:
```suggestion
    await time_travel(0)
    assert xknx.cemi_handler.send_telegram.call_args_list == [
        call(write_address),
        call(read_address),
    ]
    await time_travel(3)
    with pytest.raises(ManagementConnectionError):
        await task
```
we can wait for the timeout after the requests were sent.

farmio (changes requested):
That is a great addition! Thank you!

farmio:
Have you tested this on real world devices? I am wondering if it can yield any race conditions / awkward responses when we send `A_IndividualAddressSerialNumber_Write-PDU` and `A_IndividualAddressSerialNumber_Read-PDU` without any wait time in between to let the devices process. The specs don't suggest to do so, I am just curious if this actually works the way it should 🙃.

examples/example_write_serial_number.py
```diff
@@ -0,0 +1,37 @@
...
+from xknx import XKNX
+from xknx.management import procedures
+from xknx.telegram import IndividualAddress
+
+
+async def main(argv: list[str]) -> int:
+    """
+    Write the individual address to a device in programming mode.
+
+    This fails if multiple devices are in programming mode and/or when there is no device found in programming mode.
```
pdgendt, line 14: Update comment

xknx/management/procedures.py
```diff
@@ -184,3 +184,56 @@ async def nm_invididual_address_write(
...
+    async with xknx.management.broadcast() as bc_context:
+        broadcast_telegram = Telegram(
+            destination_address=GroupAddress("0/0/0"),
+            source_address=xknx.current_address,
+            payload=apci.IndividualAddressSerialRead(serial=serial),
+        )
+        await xknx.management.send_broadcast(broadcast_telegram)
+        async for result in bc_context.receive(timeout=timeout):
+            if isinstance(result.payload, apci.IndividualAddressSerialResponse):
+                return result.source_address
```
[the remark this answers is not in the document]
opieters: I also added a test for this check.

xknx/management/procedures.py
opieters: What I implemented is based on the sequence diagram contained in the spec: 
image. I think that the device issuing this procedure should indeed first check the availability with `nm_individual_address_check`.

xknx/management/procedures.py
```diff
@@ -184,3 +184,58 @@ async def nm_invididual_address_write(
...
+        broadcast_telegram = Telegram(
+            destination_address=GroupAddress("0/0/0"),
+            source_address=xknx.current_address,
+            payload=apci.IndividualAddressSerialRead(serial=serial),
+        )
+        await xknx.management.send_broadcast(broadcast_telegram)
+        async for result in bc_context.receive(timeout=timeout):
+            if isinstance(result.payload, apci.IndividualAddressSerialResponse):
+                if result.serial != serial:
+                    raise ManagementConnectionError("Transmitted and received serial numbers do not match.")
```
pdgendt, line 205: Do we need to `raise` or simply ignore this response?

opieters: I would raise an error, since there was a response, but not the expected one. The correct response might still appear on the bus later. If we return None, the issuer might incorrectly conclude that the device is not present on the bus. If we raise an error, the issuer can reattempt to access the individual address of the device.

examples/example_write_serial_number.py
opieters: Fixed in d5e2a7247180e5531be50748e29f6b1c45a93b2d

xknx/management/procedures.py
pdgendt: I would argue the opposite and simply ignore if we are not the destination (which we know), the caller should not need to care and wait for a timeout IMO.

opieters:
> Have you tested this on real world devices? I am wondering if it can yield any race conditions / awkward responses when we send `A_IndividualAddressSerialNumber_Write-PDU` and `A_IndividualAddressSerialNumber_Read-PDU` without any wait time in between to let the devices process. The specs don't suggest to do so, I am just curious if this actually works the way it should 🙃.

So far, this was only tested on the KNX Sentido v4 from Basalte, which did not generate any weird error and/or response in the absence of a delay between writing and reading the individual address. Not sure about other devices though, but the spec does not indicate the need for a delay. Maybe it is assumed that the delay between frames should be sufficient for the device to process the write request before the read is transmitted?

xknx/management/procedures.py
opieters: Fair point, I updated the code to ignore other broadcast messages. A timeout error will not be thrown as passed to the management procedure. I also modified the tests and error messages reported to better reflect what happened.

xknx/management/procedures.py
```diff
@@ -184,3 +184,59 @@ async def nm_invididual_address_write(
...
+    )
+
+    address = await nm_individual_address_serial_number_read(xknx=xknx, serial=serial)
+
+    if address != individual_address:
+        if address is None:
+            raise ManagementConnectionError(f"No reply received from {serial!r}.")
+        raise ManagementConnectionError(
+            f"Failed to write serial address {individual_address} to device with serial {serial!r}. Detected {address}"
+        )
```
pdgendt, line 236:
```suggestion
    if address is None:
        raise ManagementConnectionError(f"No reply received from {serial!r}.")
        
    if address != individual_address:
        raise ManagementConnectionError(
            f"Failed to write serial address {individual_address} to device with serial {serial!r}. Detected {address}"
        )
```

xknx/management/procedures.py
```diff
@@ -184,3 +184,59 @@ async def nm_invididual_address_write(
...
+        broadcast_telegram = Telegram(
+            destination_address=GroupAddress("0/0/0"),
+            source_address=xknx.current_address,
+            payload=apci.IndividualAddressSerialRead(serial=serial),
+        )
+        await xknx.management.send_broadcast(broadcast_telegram)
+        async for result in bc_context.receive(timeout=timeout):
+            if isinstance(result.payload, apci.IndividualAddressSerialResponse):
+                if result.payload.serial == serial:
+                    return result.source_address
```
pdgendt, line 205:
```suggestion
            if (
                isinstance(result.payload, apci.IndividualAddressSerialResponse)
                and result.payload.serial == serial
            ):
                return result.source_address
```

pdgendt (approved):

farmio (approved):
Nice! Thank you very much 👍

PR #1596, opened 2024-10-29
P2 p rate limit

opieters:
Added a rate limit option to P2PConnections.

This is not related to the global `rate_limit` parameter passed when creating the xknx object, which only affects data sent by xknx device objects on the bus.

Since different devices might need different limits, I added this parameter on a per-connection level in place of using the global `xknx.rate_limit` value.

Also fixes a typo in one of the management procedures `nm_invididual_address_write` was renamed to `nm_individual_address_write`. This is a breaking change.

## Type of change

## Checklist

### DIFF
--- a/xknx/management/management.py
+++ b/xknx/management/management.py
@@ -6,6 +6,7 @@ import asyncio
 from collections.abc import AsyncGenerator, AsyncIterator, Callable, Generator
 from contextlib import asynccontextmanager
 import logging
+import time
 from typing import TYPE_CHECKING
 
 from xknx.exceptions import (
@@ -81,11 +82,13 @@ class Management:
         logger.debug("Unhandled management telegram: %r", telegram)
         return
 
-    async def connect(self, address: IndividualAddress) -> P2PConnection:
+    async def connect(
+        self, address: IndividualAddress, rate_limit: int = 0
+    ) -> P2PConnection:
         """Open a point-to-point connection to a KNX device."""
         if address in self._connections:
             raise ManagementConnectionError(f"Connection to {address} already exists.")
-        p2p_connection = P2PConnection(self.xknx, address)
+        p2p_connection = P2PConnection(self.xknx, address, rate_limit)
         try:
             await p2p_connection.connect()
         except ManagementConnectionError as exc:
@@ -176,15 +179,20 @@ class BroadcastContext:
 class P2PConnection:
     """Class to manage a point-to-point connection with a KNX device."""
 
-    def __init__(self, xknx: XKNX, address: IndividualAddress) -> None:
+    def __init__(
+        self, xknx: XKNX, address: IndividualAddress, rate_limit: int = 0
+    ) -> None:
         """Initialize P2PConnection class."""
         self.xknx = xknx
         self.address = address
         self.disconnect_hook: Callable[[], None]
+        self.rate_limit = rate_limit
 
         self.sequence_number = self._sequence_number_generator()
         self._expected_sequence_number = 0
         self._connected = False
+
+        self._last_sent_time = time.time()
 
         self._ack_waiter: asyncio.Future[TAck | TNak] | None = None
         self._response_waiter: asyncio.Future[Telegram] = (
@@ -359,5 +367,15 @@ class P2PConnection:
             raise ManagementConnectionRefused(
                 "Management connection disconnected by the peer."
             )
+
+        if self.rate_limit:
+            # time in seconds since the last request operation
+            time_diff = time.time() - self._last_sent_time
+            wait_time = 1 / self.rate_limit
+            if time_diff < wait_time:
+                await asyncio.sleep(wait_time - time_diff)
+
         await self._send_data(payload)
-        return await self._receive(expected)
+        response = await self._receive(expected)
+        self._last_sent_time = time.time()
+        return response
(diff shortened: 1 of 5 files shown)


xknx/management/management.py
```diff
@@ -359,5 +367,15 @@ async def request(self, payload: APCI, expected: type[APCI] | None) -> Telegram:
...
+            # time in seconds since the last request operation
+            time_diff = time.time() - self._last_sent_time
+            wait_time = 1 / self.rate_limit
+            if time_diff < wait_time:
+                await asyncio.sleep(wait_time - time_diff)
+
         await self._send_data(payload)
-        return await self._receive(expected)
+        response = await self._receive(expected)
+        self._last_sent_time = time.time()
```
farmio, line 380:
```suggestion
        self._last_sent_time = time.time()
        response = await self._receive(expected)
```
The current implementation stores the "`_last_response_time`", not the time of sending the last payload. I'm not sure if the variable name is misleading or you actually want to rate limit only on sending.

test/management_tests/management_test.py
```diff
@@ -223,3 +223,62 @@ async def test_broadcast_message():
...
+            expected=apci.DeviceDescriptorResponse,
+        )
+    )
+    await asyncio.sleep(0)
+    send_responses(1)
+
+    if rate_limit > 0:
+        # this should raise a timeout, since the data was not yet sent
+        with pytest.raises(ManagementConnectionTimeout):
+            await task
```
farmio, line 281: I haven't used management for a while, so I'm not really following this.

What I see here:
L270 creates a task to request a response from a device. It isn't sent because still in the rate limit of the previous task - so it would theoretically wait 1 second before sending.

L276 `send_responses(1)` sends responses for a request it hasn't even received yet.
Since not yet waiting for an ACK, it is discarded (`process()`).
Same for the `DeviceDescriptorResponse` since not yet waiting for that sequence_number. (I guess `Management.process()` sends an ACK for that).

L281 awaiting the response we previously discarded - we block for a second waiting for nothing.

...
If this is correct, I guess we should at least `time_travel` in the `pytest.raises` block. And probably don't send the response beforehand, but look for the sent request - before and after the `time_travel`.

xknx/management/procedures.py
```diff
@@ -106,7 +106,7 @@ async def nm_individual_address_read(
     return addresses
 
 
-async def nm_invididual_address_write(
+async def nm_individual_address_write(
```
farmio, line 109: oops. Shall we add a `nm_invididual_address_write = nm_individual_address_write` line with some comment to avoid a breaking change for that?

farmio (commented):
Thank you very much for your contribution 😃
I really like the idea of a per-connection rate limit.

farmio:
Could you share some e example of devices that would need such rate limit? I would have guessed that when they ACK + respond, they are ready to receive a next request 🙃

Also, from your experience, what are typically needed rate limit values for those?

xknx/management/management.py
opieters: I failed to rename this variable. The idea was indeed to store the last response time as the whole exchange is considered.

opieters:
We are using these P2P connections to flash device parameters for e.g. the basalte Sentido. When flashing with specific KNX/IP bridges, we noticed that sometimes follow-up packets are sent within a few ms of the ACK and are never handled by the device. ETS and several other bridges seem to automatically add some time between a new frame and the previous ACK. If this is not done, nearly 100% of the frames timeout once, making everything very slow. Adding this feature makes it much more reliable and faster. A typical value is 20 frames per second, which results in similar performance as ETS.

test/management_tests/management_test.py
opieters: That's an alternative. The idea of the test was to do exactly the same sequence with and without the rate limit option. The new version checks for the sent messages.

opieters:
The tests I performed were with tunnelling always. Okay, I'll set it to 20 frames per second. All tests are still working, so that's good :smile:

test/management_tests/management_test.py
```diff
@@ -273,12 +275,30 @@ def send_responses(index):
...
+                        payload=apci.DeviceDescriptorRead(descriptor=0),
+                    )
+                ),
+            ]
+
             await task
     else:
         # should work as expected
+        send_responses(1)
+
```
farmio: The test now takes 6 seconds (you can check that by calling  `pytest --durations=10`) waiting for L299 `await task` since before we only make sure that the request is sent, but the timeout only starts then.
But we don't need to test the timeout here - that's done in a different test already. It can be refactored to only test the rate limit and be finished when the responses did arrive - just like the 0-rate_limit parameter does.
```py
    if rate_limit > 0:
        await time_travel(0.5 / rate_limit)
        # the request is still queued
        assert xknx.cemi_handler.send_telegram.call_args_list == []

        await time_travel(0.5 / rate_limit)
        # the request should now be sent
        assert xknx.cemi_handler.send_telegram.call_args_list == [
            call(
                Telegram(
                    destination_address=ia,
                    tpci=tpci.TDataConnected(1),
                    payload=apci.DeviceDescriptorRead(descriptor=0),
                )
            ),
        ]

    send_responses(1)

    await task
```

farmio (approved):
Thank you very much! 👍

PR #1620, opened 2024-12-12
Add support for making connections over unix domain sockets

steffen-kiess:
## Description

Add support for making connections over unix domain sockets.

The protocol used for the unix domain sockets is the same as the TCP protocol. The primary advantage of using unix domain sockets is that permissions can be applied if the server is running on the same host as the xknx application. This can be used to run knxd (e. g. to connect to a USB interface) and expose the socket only for certain users/groups. knxd supports TCP and unix domain sockets starting with version 0.14.68 (or PR  <https://github.com/knxd/knxd/pull/573>).

This also adds support for connecting to arbitrary streams, which can be used to connect to a knxd on another host using ssh, see `examples/example_telegram_monitor_unix.py` for an example.

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -20,6 +20,10 @@ nav_order: 2
 - Fix typo in management procedure (`nm_invididual_address_write` was renamed to `nm_individual_address_write`)
 - Fix TunnellingFeatureResponse missing `return_code`
 
+### Connection
+
+- Add support for making connections over unix domain sockets.
+
 # 3.3.0 Climate humidity 2024-10-20
 
 ### Devices
--- a/examples/example_telegram_monitor_unix.py
+++ b/examples/example_telegram_monitor_unix.py
@@ -1,0 +1,86 @@
+"""Example for the telegram monitor callback over unix domain socket."""
+
+import asyncio
+import getopt
+import socket
+import subprocess
+import sys
+
+from xknx import XKNX
+from xknx.io import ConnectionConfig, ConnectionType
+from xknx.telegram import AddressFilter, Telegram
+
+
+def telegram_received_cb(telegram: Telegram) -> None:
+    """Do something with the received telegram."""
+    print(f"Telegram received: {telegram}")
+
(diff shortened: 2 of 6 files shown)


farmio:
Hi 👋! 
Thanks for your contribution efforts!

I'd have a couple of questions regarding this. 

- Why do you use the tcp tunnelling protocol and not routing protocol?
- Are there any other apps / devices, apart from knxd, that support Knx IP over Unix domain sockets?
- What are the real world use cases for that?
- I don't understand this part:
  > support for connecting to arbitrary streams, which can be used to connect to a knxd on another host using ssh

  Wouldn't providing the sockets be responsibility of the app using xknx, instead of xknx itself? Why do we need a `connect_cb`?

steffen-kiess:
> Hi 👋! Thanks for your contribution efforts!
> 
> I'd have a couple of questions regarding this.
> 
>     * Why do you use the tcp tunnelling protocol and not routing protocol?

The routing protocol does not hand out individual addresses, and also I'm not sure how to transport that over a unix domain socket.

>     * Are there any other apps / devices, apart from knxd, that support Knx IP over Unix domain sockets?

No, not that I'm aware of.

>     * What are the real world use cases for that?

In my case I want to use knxd to connect to a KNX USB interface. Connecting from xknx to knxd over TCP or UDP is possible, but means that everyone on the current host can access the KNX bus, which I'd like to avoid. With unix domain sockets, I can ask systemd to set the permissions in a way that only certain users can access the `/run/knxnet` socket.

>     * I don't understand this part:
>       > support for connecting to arbitrary streams, which can be used to connect to a knxd on another host using ssh
>       
>       
>       Wouldn't providing the sockets be responsibility of the app using xknx, instead of xknx itself? Why do we need a `connect_cb`?

Providing a socket e.g. in the `ConnectionConfig` class would break reconnecting, because a new socket has to be used for the second connection attempt. That's why I decided to use a callback here.

PR #1630, opened 2025-01-05
Add pre-commit hook for pylint

philippwaller:
## Description

This PR adds a pre-commit hook for pylint to ensure code quality checks are performed automatically before each commit.

## Type of change

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,6 +63,7 @@ repos:
       language: python
       types: [python]
       pass_filenames: true
+      require_serial: true
       files: ^(xknx/|examples/)
 
     - id: pylint-test
@@ -71,5 +72,6 @@ repos:
       language: python
       types: [python]
       pass_filenames: true
+      require_serial: true
       files: ^test/
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -54,3 +54,15 @@ repos:
         types: [python]
         require_serial: true
         files: ^xknx/.+\.py$
+  - repo: local
+    hooks:
+      - id: pylint
+        name: pylint
+        entry: pylint
+        language: system
+        types: [python]
+        args:
+          [
+            "-rn", # Only display messages
+            "-sn", # Don't display the score
+          ]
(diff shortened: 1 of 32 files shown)


farmio:
This was not part of pre-commit since pylint was uncomfortably slow - and ruff covered most anyway. Is this faster nowadays?

Also, have a look at our tox pylint invocation. We run it with different arguments for tests. Not sure what default `jobs` is, but setting to 0 should speed things up a little. 
https://github.com/XKNX/xknx/blob/905b246592c9add976bfda5f305f13b6f1cbad01/tox.ini#L31-L32
Maybe this can be configured in pyproject.toml for a global pylint invocation.

philippwaller:
I initially added it because I noticed inconsistencies in the test results when running tox versus executing pylint directly. At first, I suspected this was due to different virtual environments being used, but I’ve now realized the root cause is different:

It turns out that modules are only recognized if an `__init__.py` file is present.

When running:
`pylint --disable=protected-access,abstract-class-instantiated test`

No issues are reported. However, when specifying a file explicitly:

`pylint --disable=protected-access,abstract-class-instantiated test test/dpt_tests/payload_test.py`

```
************* Module payload_test
test/dpt_tests/payload_test.py:76:15: C2801: Unnecessarily calls dunder method __repr__. Use repr built-in function. (unnecessary-dunder-call)
test/dpt_tests/payload_test.py:77:15: C2801: Unnecessarily calls dunder method __repr__. Use repr built-in function. (unnecessary-dunder-call)
```

`pre-commit` still reports all errors because it runs pylint on each file individually (actually in batches). However, with the latest update, performance is ok-ish since it only lints the changed files.

I’ve also added the missing `__init__.py` files to fix the tox workflow.

farmio:
https://pylint.pycqa.org/en/latest/user_guide/installation/pre-commit-integration.html explicitly mentions 
> We however discourage it as pylint -- due to its speed -- is more suited to a continuous integration job or a git pre-push hook

thus I asked.

Further
>  If only some of the files are given, pylint might miss a particular value's type and produce inferior inference for the subset. Since pre-commit slices the files given to it in order to parallelize the processing, the result can be degraded. It can also be unexpectedly different when the file set changes because the new slicing can change the inference. Thus the require_serial option should be set to true (...)

So should we `require_serial: true`?

PR #1642, opened 2025-01-19
Add swing and horizontal swing to climate device

### LINKED ISSUE #466: Feature request: Add Fan mode and Swing mode to climate KNX devices
Hi All,
Would it be possible to add Fan Mode and Swing Mode to KNX Climate devices? It is supported by Home Assistant in other Climate integrations (e.g. Sensibo Climate). If needed I can help with testing, I have a KNX modules for Mitsubishi A/C units which enable to control all features of the A/C unit including Fan modes and Swing modes. If needed I can also share which DPTs are used to control those modes.

richardpolzer:
## Description
Add swing and horizontal swing support to climate device

Fixes #466

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,10 @@ nav_order: 2
 
 # Changelog
 
+### Features
+
+- Added swing and horizontal swing support to climate device
+
 # 3.4.0 8 byte energy and 4 byte pressure 2024-11-20
 
 ### Devices
--- a/docs/climate.md
+++ b/docs/climate.md
@@ -22,6 +22,10 @@ Climate are representations of KNX HVAC/Climate controls.
 - `group_address_setpoint_shift_state` KNX address to read current setpoint_shift. *DPT 6.010* or *DPT 9.002*
 - `group_address_fan_speed` KNX address for the fan speed. *DPT 5.001 / 5.010*
 - `group_address_fan_speed_state` KNX address for reading fan speed. *DPT 5.001 / 5.010*
+- `group_address_swing` KNX address for fan swing. If `group_address_horizontal_swing` is present this should refer to vertical swing *DPT 1*
+- `group_address_swing_state` KNX address for reading fan swing. If `group_address_horizontal_swing` is present this should refer to vertical swing *DPT 1*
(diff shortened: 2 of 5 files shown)


farmio (approved):
Hi 👋! 
Thank you very much for your contribution!

farmio:
Do you plan to add this to the HA KNX integration too?

The code looks perfectly fine, but I have a little concern about it in general: None of the added attributes or methods interact in any way with the rest of the `Climate` class. 

These 2 swing attributes stand for their own. Thus, I think in HA we could just add 2 extra `xknx.devices.Switch` instances to `KNXClimate(ClimateEntity)` on demand, instead of increasing size and complexity of the xknx Climate class (and its tests).
This could be handled like `ClimateMode` https://github.com/home-assistant/core/blob/96e40334b179d77d669035251716179a83571fa8/homeassistant/components/knx/climate.py#L416-L428
, but in a more generalised way. (As having `ClimateMode` as an attribute of `Climate` is also just tech-dept imho).

What do you think about that?

richardpolzer:
Yes I plan on adding it to HA too.

I understand your concern. However if we want to abstract a device with the climate class, we should include all possible functions within that class and not rely on logic within systems that use the library. At least thats my opinion on that.

farmio:
> if we want to abstract a device with the climate class, we should include all possible functions within that class and not rely on logic within systems that use the library

Currently, I think the system using the lib - HA in this case - dictates the functions it can consume. So any future extension to a HA entity model (eg. `homeassistant.components.climate.ClimateEntity`) would need a reflection in xknx (`xknx.devices.Climate`). Leaving `homeassistant.components.knx.KNXClimate` as an intermediary layer without much functionality - it's mostly a translation layer between HA and xknx method/property names. But still we have to touch both, xknx and the HA knx integration, to add a feature. Including tests and code maintenance in both repos.
Using multiple xknx device instances per HA Entity (one per independent feature) instead of a single xknx device per HA Entity could (maybe) eliminate the need to change xknx code to add / change features to HA entities.

Anyway - we didn't do it that way until now, so let's leave it this way... can always be broken up later if we wanted to.

> Yes I plan on adding it to HA too.

Just let me know when you are ready - so I can do an xknx release when you tested your feature with the integration code.

PR #1678, opened 2025-04-11
Split multicast socket for UDPTransport

pdgendt:
Create a separate socket when creating an UDPTransport with a remote multicast address to receive and send messages.
This allows us to distinguish messages we sent ourselves.

## Description

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/test/integration_tests/routing_multicast_test.py
+++ b/test/integration_tests/routing_multicast_test.py
@@ -1,0 +1,51 @@
+"""Tests for KNX/IP routing indications using multicast."""
+
+import asyncio
+from unittest.mock import Mock
+
+from xknx import XKNX
+from xknx.dpt.payload import DPTBinary
+from xknx.io.connection import ConnectionConfig, ConnectionType
+from xknx.telegram import GroupAddress, IndividualAddress, Telegram
+from xknx.telegram.apci import GroupValueWrite
+from xknx.telegram.telegram import TelegramDirection
+from xknx.telegram.tpci import TDataGroup
+from xknx.util import asyncio_timeout
+
+
+async def test_routing_indication_multicast() -> None:
+    """Test exchanging routing indication using multicast."""
+    routing1_mock = Mock()
+    data_received = asyncio.Event()
+    received_data: Telegram | None = None
+
+    def routing2_callback(telegram: Telegram) -> None:
+        """Set test data from receiving xknx instance."""
+        nonlocal received_data
+        received_data = telegram
+        data_received.set()
+
+    msg = Telegram(
+        destination_address=GroupAddress("1/2/2"),
+        source_address=IndividualAddress("1.1.1"),
+        payload=GroupValueWrite(DPTBinary(1)),
+        tpci=TDataGroup(),
+    )
+
+    async with (
+        XKNX(
+            connection_config=ConnectionConfig(connection_type=ConnectionType.ROUTING),
+            telegram_received_cb=routing1_mock,
+        ) as xknx1,
+        XKNX(
+            connection_config=ConnectionConfig(connection_type=ConnectionType.ROUTING),
+            telegram_received_cb=routing2_callback,
+        ),
+    ):
+        await xknx1.telegrams.put(msg)
+        async with asyncio_timeout(1):
+            await data_received.wait()
+
+        msg.direction = TelegramDirection.INCOMING
+        routing1_mock.assert_not_called()
+        assert received_data == msg
--- a/test/integration_tests/routing_multicast_test.py
+++ b/test/integration_tests/routing_multicast_test.py
@@ -1,6 +1,7 @@
 """Tests for KNX/IP routing indications using multicast."""
 
 import asyncio
+from typing import Final
 from unittest.mock import Mock
 
 from xknx import XKNX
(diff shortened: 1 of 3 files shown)


xknx/io/transport/udp_transport.py
```diff
@@ -72,10 +76,16 @@ def __init__(
...
 
         self.callbacks = []
+        self.local_addr_assigned: tuple[str, int] | None = None
         self.transport: asyncio.DatagramTransport | None = None
+        self.mcast_transport: asyncio.DatagramTransport | None = None
 
     def data_received_callback(self, raw: bytes, source: tuple[str, int]) -> None:
         """Parse and process KNXIP frame. Callback for having received an UDP packet."""
+        if self.multicast and source == self.local_addr_assigned:
+            knx_logger.debug("Received echoed KNXIPFrame")
```
farmio, line 86:
```suggestion
            raw_socket_logger.debug("Discarded echoed datagram")
```
I think this is better handled by `raw_socket_logger` since we haven't parsed it to a KNXIPFrame yet.

test/integration_tests/routing_multicast_test.py
```diff
@@ -0,0 +1,42 @@
...
+            connection_config=ConnectionConfig(connection_type=ConnectionType.ROUTING),
+            telegram_received_cb=routing2_mock,
+        ),
+    ):
+        await xknx1.telegrams.put(msg)
+        await asyncio.sleep(0.1)
+
+        msg.direction = TelegramDirection.INCOMING
+        routing1_mock.assert_not_called()
+        routing2_mock.assert_called_with(msg)
```
farmio, line 42:
```suggestion
from xknx.telegram.tpci import TDataGroup
from xknx.util import asyncio_timeout


async def test_routing_indication_multicast() -> None:
    """Test exchanging routing indication using multicast."""
    routing1_mock = Mock()
    data_received = asyncio.Event()
    received_data: Telegram | None = None

    def routing2_callback(telegram: Telegram) -> None:
        """Set test data from receiving xknx instance."""
        nonlocal received_data
        received_data = telegram
        data_received.set()

    msg = Telegram(
        destination_address=GroupAddress("1/2/2"),
        source_address=IndividualAddress("1.1.1"),
        payload=GroupValueWrite(DPTBinary(1)),
        tpci=TDataGroup(),
    )

    async with (
        XKNX(
            connection_config=ConnectionConfig(connection_type=ConnectionType.ROUTING),
            telegram_received_cb=routing1_mock,
        ) as xknx1,
        XKNX(
            connection_config=ConnectionConfig(connection_type=ConnectionType.ROUTING),
            telegram_received_cb=routing2_callback,
        ),
    ):
        await xknx1.telegrams.put(msg)
        async with asyncio_timeout(1):
            await data_received.wait()

        msg.direction = TelegramDirection.INCOMING
        routing1_mock.assert_not_called()
        assert received_data == msg
```
We should avoid `sleep` in tests.

farmio (changes requested):

farmio:
I've tested this locally (with Gira and Enertex IP Secure Routers) and it works fine.

One thing I found is that the integration test could be problematic eventually. It sends real KNX telegrams out to the network - so if someone runs the `pytest` locally with a non-knx-secure environment - this may trigger actual things in their KNX installation. Using `local_ip="127.0.0.1"` doesn't seem to work at my machine. So there are 3 workarounds I can think of:
- use a non-standard multicast group (not ideal - some users may have test environments using the same)
- use secure routing - but in the current implementation this adds quite a bit of time (>3 seconds) to the tests when trying to synchronize the secure timer. Maybe this can be mock-patched to work better, but I haven't succeeded in some quick trys.
- exclude it from running locally and only run it in CI (or locally when called manually)

pdgendt:
> I've tested this locally (with Gira and Enertex IP Secure Routers) and it works fine.

Nice!

> One thing I found is that the integration test could be problematic eventually. It sends real KNX telegrams out to the network - so if someone runs the `pytest` locally with a non-knx-secure environment - this may trigger actual things in their KNX installation. Using `local_ip="127.0.0.1"` doesn't seem to work at my machine. So there are 3 workarounds I can think of:

I also tried `local_ip="127.0.0.1"` which seems to work as expected?

> * use a non-standard multicast group (not ideal - some users may have test environments using the same)
> * use secure routing - but in the current implementation this adds quite a bit of time (>3 seconds) to the tests when trying to synchronize the secure timer. Maybe this can be mock-patched to work better, but I haven't succeeded in some quick trys.
> * exclude it from running locally and only run it in CI (or locally when called manually)

If the local IP isn't a valid solution, I'd prefer option 1 here.

farmio:
CI seems to work fine with 127.0.0.1
On what OS did you test it? Didn't work on macOS for me.

pdgendt:
> CI seems to work fine with 127.0.0.1 On what OS did you test it? Didn't work on macOS for me.

Ubuntu, but the previous test already failed on my mac with `No route to host`.

farmio:
> WARNING  xknx.log:udp_transport.py:53 Error received: [Errno 49] Can't assign requested address

here

farmio:
Hm... what do you think about 224.0.0.254 ? https://www.rfc-editor.org/rfc/rfc4727.html

farmio (approved):
Nice 👍 Thanks for your contribution!

PR #1681, opened 2025-04-25
Support explicitly passing a keyring in the secure config

cpha-trifork:
## Description

Allow a keyring to be passed explicitly through a secure config rather than always being read from a file.

The motivation is that for our usage we have the desired contents of the keyring in memory and it would be great if it wasn't necessary to write it to a file only to have `xknx` read it back into memory immediately after.

## Type of change

- [X] Code quality improvements to existing code or addition of tests

(This isn't really any of the above)

## Checklist

(I'm not sure any of these are relevant but please let me know if they are.)


farmio:
Hi 👋! Thank you very much for your contribution! 
Can you please add a small test for this, so we can make sure future changes don't break this feature?

cpha-trifork:
Thanks for the fast response. Have added a test.

PR #1685, opened 2025-05-11
Fix flipped DPT 235.001 data validity bits

NHanser:
## Description

Fix DPT 235.001 validation check.
When we look at the specifications : 
DPT235.001 spec

We can see that validation bits are at 1 when data is **_not_** valid.
Current check does the opposite, checking if it has 1 value to consider them as valid

Fixes # (issue)

## Type of change

## Checklist


farmio (changes requested):
Hi 👋! 
Thanks for your contribution! 
Looking at the code, it seems outgoing payloads - `to_knx` - also have this wrong.

NHanser:
@farmio thanks for the review, I definitely did it too quickly....should be fixed now

farmio (approved):
Thank you very much 👍

PR #1708, opened 2025-07-25
Cover: add invert_stop

emarj:
## Description

In my KNX installation cover stopping signal is inverted. I don't know how standard it is, but since we already have flags to invert up/down, angle and position I think it will be good to have this as well.

I did not add tests since I saw that was not done for other invert flags like `invert_angle`.
I did not edit the changelog, don't want to mess it up.

Thanks for the awesome project!

## Type of change

## Checklist

### DIFF
--- a/docs/cover.md
+++ b/docs/cover.md
@@ -27,6 +27,7 @@ Shutters are simple representations of blind/roller cover actuators. With XKNX y
 - `travel_time_down` seconds to reach lower end position. Default: 22
 - `travel_time_up` seconds to reach upper end position. Default: 22
 - `invert_updown` invert up/down binary payload. Default: False
+- `invert_stop` invert stop binary payload. Default: False
 - `invert_position` invert position percentage from / to KNX bus. Default: False
 - `invert_angle` invert angle from / to KNX bus. Default: False
 - `device_updated_cb` Callback for each update.
--- a/xknx/devices/cover.py
+++ b/xknx/devices/cover.py
@@ -58,6 +58,7 @@ class Cover(Device):
         travel_time_down: float = DEFAULT_TRAVEL_TIME,
         travel_time_up: float = DEFAULT_TRAVEL_TIME,
         invert_updown: bool = False,
+        invert_stop: bool = False,
         invert_position: bool = False,
         invert_angle: bool = False,
         device_updated_cb: DeviceCallbackType[Cover] | None = None,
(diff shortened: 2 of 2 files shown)


farmio:
Hi 👋! 

What actuator are you using that only supports `0` for stopping? 

There is a DPT that only has `1` defined (1.017 Trigger) - there is no DPT that only defines a `0` payload, but no `1`.

In 07_50_01 Shutters and blinds sensors v01.03.02 AS there is §1.2.4 and §1.5.3.3 Output Dedicated Stop
> 0, 1: Requests to stop the movement.

In addition to that, there is §1.5.3.2 Output StopStep Up Down which defines what our `group_address_short` does - move in a direction and stop when the same payload is received - revert direction if the other was received.

emarj:
Hi farmio!

Unfortunately the KNX programming was done by a contractor I do not have full visibility on that. The blinds are managed by a separate system by [Nice](https://niceforyou.com/) with a proprietary KNX bridge (DMKNX). Not sure if this is a bug of this KNX bridge or a misconfiguration in its ETS programming. I will try to dig further, but could take time.

I monitored the bus to see that the payload of the stop command triggered by buttons is 0 and, if I put 1, the cover does not stop. Moreover I can say that in the KNX proj (parsed by xknx) the payload 0 it is labeled as "stop" and 1 as "start".

If this is non-standard I totally understand you might not want it in the library. In the worst case I will keep using the patched library until I find a fix for this.

emarj:
Thanks a lot for your time.

We are talking about a lot of covers, so that would be a lot of automations to maintain. I solved by using `knx_event`:
```yaml
knx:
    event:
        - address:
            - "2/1/*"
```
and
```yaml
alias: Invert covers stop signal
triggers:
  - event_type: knx_event
    trigger: event
conditions:
  - condition: template
    value_template: |-
      {{ trigger.event.data.direction == 'Outgoing'
               and trigger.event.data.data == true
               and trigger.event.data.destination.startswith('2/1/') }}
actions:
  - data:
      address: "{{ trigger.event.data.destination }}"
      payload: 0
    action: knx.send
mode: single
```

With a trigger it would have been much cleaner, but apparently you can't use a pattern to match several GAs, maybe that's a PR I could submit instead! What do you think?
Also it would be nice if for every entry in `knx_event` one could specify an `id` that is then injected in the `event.data`. What do you think?

Thanks again

PR #1721, opened 2025-08-27
secure: use typing.ByteString

### LINKED ISSUE #1719: xknx 3.9.0 breaks compatibility with cryptography<45
This commit https://github.com/XKNX/xknx/commit/ba93c123d125eecd2f0f12dc239fef577360bf14 imports cryptography.utils.Buffer, which was added in cryptography 45.0.0 here: https://github.com/pyca/cryptography/commit/9cd7902d2cc090af22d15ec833834430e14a7fef, however, the declared minimum version is much lower (cryptography>=35.0.0). cryptography.utils.Buffer also seems like a private detail (it isn't documented, in the documentation all references show up as `bytes` instead).

Since xknx targets Python 3.10+, typing.ByteString can be used directly.

enkore:
restores compatibility with cryptograph<45
fixes #1719

## Description

Fixes #1719
## Type of change

## Checklist

- [-] The documentation has been adjusted accordingly
- [-] Tests have been added that prove the fix is effective or that the feature works
- [?] The changes are documented in the changelog (docs/changelog.md)


enkore:
collections.abc.Buffer has only been added in Python 3.12, so I suppose we could define it manually like cryptography does (`Buffer = typing.Union[bytes, bytearray, memoryview]` or rather `Buffer = bytes | bytearray | memoryview`).

farmio:
Since `collections.abc.Buffer` was added 3.12, we'd still have a gap.
So I guess we could define our own in xknx.typing `bytes | bytearray | memoryview` just do `if TYPE_CHECKING` when importing Buffer from cryptography.

xknx/secure/util.py
```diff
@@ -1,7 +1,9 @@
 """Utilities for KNX Secure."""
 
 from cryptography.hazmat.primitives import hashes
-from cryptography.utils import Buffer
+
+
+Buffer = bytes | bytearray | memoryview
```
farmio, line 6:
```suggestion
# py3.12 collections.abc.Buffer can be used instead
Buffer = bytes | bytearray | memoryview
```
Just added a note to remind removing this when we require Python 3.12

farmio (approved):

PR #1802, opened 2026-02-20
feat: add custom interface injection via ConnectionConfig

matssun:
> **Note:** This PR is based on #1803 (Interface ABC enhancement). Please review/merge #1803 first.

## Summary

- Add `ConnectionType.CUSTOM` and `interface_factory` parameter to `ConnectionConfig`
- Enable dependency injection of custom `Interface` implementations without subclassing `KNXIPInterface`
- The factory callable receives the `XKNX` instance and a CEMI received callback, and returns an `Interface` implementation

## Motivation

Currently, users who need custom transport implementations (e.g., wrapping KNX/IP communication in a different networking stack, proxying through a message bus, or testing with mock transports) must subclass `KNXIPInterface` and override internal `_start_*` methods. This is fragile and couples to implementation details.

This PR adds a clean injection point: pass an `interface_factory` callable to `ConnectionConfig` with `connection_type=ConnectionType.CUSTOM`, and `KNXIPInterface` will use it to create the interface instead of its built-in factory methods.

## Usage

```python
from xknx import XKNX
from xknx.io import ConnectionConfig, ConnectionType

def my_interface_factory(xknx, cemi_received_callback):
    return MyCustomInterface(xknx, cemi_received_callback)

xknx = XKNX(
    connection_config=ConnectionConfig(
        connection_type=ConnectionType.CUSTOM,
        interface_factory=my_interface_factory,
    )
)
```

## Design

- `InterfaceFactoryType = Callable[[XKNX, CEMIBytesCallbackType], Interface]` — typed factory signature
- `ConnectionType.CUSTOM` — new enum value, checked first in `_start()`
- `_start_custom()` — calls the factory then `await interface.connect()`
- Error raised if `CUSTOM` is used without providing `interface_factory`

## Backward Compatibility

Fully backward-compatible:
- `interface_factory` defaults to `None`
- All existing `ConnectionType` values work unchanged
- No changes to existing method signatures
- Full test suite passes (2880/2880)

## Test plan

- [x] All 2880 existing tests pass unchanged
- [x] Unit tests for `ConnectionType.CUSTOM` with mock interface factory
- [x] Test for error case: `CUSTOM` without `interface_factory`

### DIFF
--- a/test/io_tests/knxip_interface_test.py
+++ b/test/io_tests/knxip_interface_test.py
@@ -4,7 +4,7 @@ from collections.abc import AsyncGenerator
 from pathlib import Path
 import threading
 from typing import Any
-from unittest.mock import DEFAULT, Mock, patch
+from unittest.mock import AsyncMock, DEFAULT, Mock, patch
 
 import pytest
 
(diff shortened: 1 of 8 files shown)


matssun:
> Hi 👋!
> 
> > Currently, users who need custom transport implementations (e.g., wrapping KNX/IP communication in a different networking stack, proxying through a message bus, or testing with mock transports)
> 
> I'd be curious what's the actual use case of those is.
> 
> I think `xknx` and `cemi_received_callback` could be added to `Interface` abstract type. Maybe also add abstractions for setting connected / disconnected and `xknx.current_address`. That way users of the Interface type wouldn't need to know what to do with the xknx object in the first place.

Hi farmio,                                                                                         
                                                                                                     
  Thanks for the thoughtful feedback! Let me clarify our use case:

  We have a centralized network stack that's used throughout our entire codebase for all
  inter-component communication. This custom stack provides several features beyond low-level KNX/IP
  communication:

  - Unified routing and message handling across all communication protocols (not just KNX)
  - Built-in observability (logging, tracing, metrics) integrated with our monitoring infrastructure
  - Request/response correlation and distributed tracing
  - Cross-cutting concerns like encryption, authentication, and audit logging
  - Consistent error handling and resilience patterns across the organization

  By allowing dependency injection of this network stack into the KNX module, we can:

  1. Maintain consistency across our entire system—KNX communication uses the same infrastructure as
  everything else
  2. Avoid duplicating observability and networking features in the KNX adapter
  3. Keep our codebase maintainable by having a single, well-tested networking layer

  Regarding your suggestion about expanding the Interface abstract type: that's a great idea and
  worth exploring. Adding xknx, cemi_received_callback, and lifecycle methods to the interface would
  definitely reduce boilerplate. However, we'd still need flexibility to inject our custom network
  stack, since the lifecycle and transport concerns in our system extend beyond what KNX/IP typically
   requires.

  Would it be feasible to support both approaches—enhancing the Interface while also allowing for
  custom transport injection?

PR #1803, opened 2026-02-25
Enhance Interface ABC with common connection management

matssun:
## Summary

Following @farmio's suggestion in #1802, this PR enhances the `Interface` ABC to formalize what all concrete implementations already share:

- Added `xknx` and `cemi_received_callback` to `__slots__` and type annotations on `Interface`
- Added `connection_state_changed()` concrete method (wraps `xknx.connection_manager`)
- Added `current_address` property (wraps `xknx.current_address`)
- Updated `_Tunnel` and `Routing` to use these Interface methods instead of direct `xknx` access
- Removed duplicate slot declarations from subclasses

## Motivation

Every `Interface` implementation (`_Tunnel`, `Routing`) already stores `xknx` and `cemi_received_callback`, and uses the same patterns for connection state changes and address management. Moving these to the base class:

1. Eliminates duplicated code across implementations
2. Establishes a clear contract for what an `Interface` provides
3. Prepares the ABC for custom transport implementations (as discussed in #1802)

## Breaking change

`xknx` and `cemi_received_callback` are now declared in `Interface.__slots__`. Third-party `Interface` subclasses that redeclare these names in their own `__slots__` will need to remove them to avoid a `ValueError` conflict. This is a one-line fix per subclass.

## Changes

| File | Change |
|------|--------|
| `xknx/io/interface.py` | Added slots, type annotations, `connection_state_changed()`, `current_address` property |
| `xknx/io/tunnel.py` | Removed duplicate `xknx`/`cemi_received_callback` slots, use Interface methods |
| `xknx/io/routing.py` | Removed duplicate `xknx`/`cemi_received_callback` slots, use Interface methods |

## Testing

Pure refactor — no behavior changes. All 2876 existing tests pass with zero regressions.

### DIFF
--- a/xknx/io/interface.py
+++ b/xknx/io/interface.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Callable
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, ClassVar
 
 from xknx.cemi import CEMIFrame
 from xknx.core import XknxConnectionState, XknxConnectionType
@@ -29,6 +29,7 @@ class Interface(ABC):
 
     __slots__ = ("cemi_received_callback", "transport", "xknx")
 
+    connection_type: ClassVar[XknxConnectionType]
     cemi_received_callback: CEMIBytesCallbackType
     transport: KNXIPTransport
     xknx: XKNX
@@ -45,14 +46,10 @@ class Interface(ABC):
     async def send_cemi(self, cemi: CEMIFrame) -> None:
         """Send CEMIFrame to KNX bus."""
 
-    def connection_state_changed(
-        self,
-        state: XknxConnectionState,
-        connection_type: XknxConnectionType = XknxConnectionType.NOT_CONNECTED,
-    ) -> None:
+    def connection_state_changed(self, state: XknxConnectionState) -> None:
         """Update connection state via connection manager."""
         self.xknx.connection_manager.connection_state_changed(
-            state, connection_type
+            state, self.connection_type
         )
 
     @property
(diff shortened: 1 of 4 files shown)


xknx/io/interface.py
```diff
@@ -36,3 +44,23 @@ async def disconnect(self) -> None:
...
+
+    @property
+    def current_address(self) -> IndividualAddress:
+        """Get current individual address."""
+        return self.xknx.current_address
+
+    @current_address.setter
+    def current_address(self, address: IndividualAddress) -> None:
+        """Set current individual address."""
+        self.xknx.current_address = address
```
farmio: I don't think we need to read it from an Interface, but only set it so I don't think we need a property contract here. A private method should suffice.

```suggestion
    def _set_individual_address(self, address: IndividualAddress) -> None:
        """Set current individual address."""
        self.xknx.current_address = address
```

xknx/io/interface.py
```diff
@@ -36,3 +44,23 @@ async def disconnect(self) -> None:
...
+
+    def connection_state_changed(
+        self,
+        state: XknxConnectionState,
+        connection_type: XknxConnectionType = XknxConnectionType.NOT_CONNECTED,
+    ) -> None:
+        """Update connection state via connection manager."""
+        self.xknx.connection_manager.connection_state_changed(
+            state, connection_type
+        )
```
farmio: We could annotate `connection_type` as ClassVar and use it here for simplicity.

```suggestion
    def connection_state_changed(self, state: XknxConnectionState) -> None:
        """Update connection state via connection manager."""
        self.xknx.connection_manager.connection_state_changed(state)
```
And maybe use `XknxConnectionType.NOT_CONNECTED` when `XknxConnectionState.DISCONNECTED` in `xknx.connection_manager.connection_state_changed`

What do you think?

farmio (changes requested):

PR #1831, opened 2026-05-16
refactor(management): split procedures.py into per-procedure package

kewde:
## Description

`xknx/management/procedures.py` (single file, 6 procedures) becomes the `xknx/management/procedures/` package. Each procedure now lives in its own file under a subdirectory named after its KNX spec family, with the family prefix in the filename so `grep` / IDE search lands on it directly:

```
xknx/management/procedures/
├── __init__.py                                  # re-exports
├── device/
│   └── dm_restart_rco.py
└── network/
    ├── nm_individual_address_check.py
    ├── nm_individual_address_read.py
    ├── nm_individual_address_serial_number_read.py
    ├── nm_individual_address_serial_number_write.py
    └── nm_individual_address_write.py
```

If we want to add ~160 procedures across KNX 03.05.02 (Management Procedures) and 03.05.03 (Configuration Procedures). The single-file `procedures.py` would not scale. This PR is the scaffold; subsequent PRs add procedures as `<family>/<prefix>_<name>.py` files into the new layout.

Pure code move — function bodies are byte-identical to the pre-split versions; the logger name is preserved, and the public API is unchanged (`from xknx.management import procedures` and `procedures.dm_restart` still resolve via `__init__.py` re-exports). The legacy typo alias `nm_invididual_address_write` is preserved.

Tests mirror the layout under `test/management_tests/procedures/<family>/test_<prefix>_<name>.py`. All 25 management tests pass unchanged.

## Type of change

## Checklist


farmio:
For readability I'd use full-names for the family package directory instead of `nm` and `dm`. Or `device` and `network`. 
The modules are fine using `nm_*.py` imho. 
What do you think?

kewde:
@farmio agreed, applied 👍

xknx/management/procedures/device/dm_restart_r_co.py
```diff
@@ -0,0 +1,36 @@
...
+from xknx.telegram import Telegram, apci, tpci
+from xknx.telegram.address import IndividualAddress, IndividualAddressableType
+
+if TYPE_CHECKING:
+    from xknx import XKNX
+
+logger = logging.getLogger("xknx.management.procedures")
+
+
+async def dm_restart(xknx: XKNX, individual_address: IndividualAddressableType) -> None:
```
kewde, line 17: IMHO: replace xknx parameter with a conn: P2PConnection

For connectionless procedures we can have a different calling convention but for procedures that require dm_connect we can just accept a P2PConnection, which implements dm_connect but lacks the device read request.

xknx/management/procedures/network/nm_individual_address_check.py
```diff
@@ -0,0 +1,49 @@
...
+from xknx.telegram.address import IndividualAddress, IndividualAddressableType
+
+if TYPE_CHECKING:
+    from xknx import XKNX
+
+logger = logging.getLogger("xknx.management.procedures")
+
+
+async def nm_individual_address_check(
+    xknx: XKNX, individual_address: IndividualAddressableType
```
kewde, line 19: ditto, could just use a P2PConnection here
looking at the other network procedures, it would also need something for broadcasts though, I assume that is why XKNX is being passed instead.

farmio: Especially `nm_individual_address_read` would need broadcast. `nm_individual_address_write` would even need both.
And I guess (have not checked in specs) that there are multiple procedures that would want to do the broadcast-read before doing connection stuff.

The original idea was to have a user-friendly interface like 
```py
    async with XKNX() as xknx:
        individual_address = IndividualAddress(address)
        await procedures.nm_individual_address_write(xknx, individual_address)
```
for all the procedures - without the user needing to know implementation details about broadcast, connection-less or connection-oriented communication modes.

xknx/management/procedures/device/dm_restart_r_co.py
farmio: Is `dm_connect` what we do in `xknx.management.connect` - sending a `TConnect` payload? In this case, we don't automatically do the `DeviceDescriptor_Read`, but leave that to the procedure implementation (example `nm_individual_address_check`).

I think Connection-less communication wasn't taken into account when implementing this, as all procedures that would allow it have to support connection-oriented communication too.

xknx/management/procedures/__init__.py
```diff
@@ -0,0 +1,36 @@
+"""
+Management procedures grouped by KNX spec family.
+
+Subpackages are created when their first procedure lands. Naming mirrors
+the KNX spec prefix:
+
+  - ``network/`` for NM_* procedures (KNX 03.05.02 Network Management)
+  - ``device/`` for DM_* procedures (KNX 03.05.02 Device Management)
+  - ``dmp/`` for DMP_* procedures (KNX 03.05.02 Device Management Procedures)
```
farmio: `dmp` seems to be a subtype of `dm`. Maybe we can use `device/` for both?

xknx/management/procedures/device/dm_restart_r_co.py
kewde: > Is dm_connect what we do in xknx.management.connect - sending a TConnect payload? In this case, we don't automatically do the DeviceDescriptor_Read, but leave that to the procedure implementation (example nm_individual_address_check).

Correct, but the spec seems to mandate a `DeviceDescriptor_Read` in `dm_connect`. The question is, do we want to mandate that in `P2PConnection` too? We could have a `dm_connect` procedure that just does `DeviceDescriptor_Read` over the `P2PConnection`.

 The problem seems to be chaining individual procedures becomes impossible with the `xknx.management.connect` pattern since every individual procedure would be opening up its own connection if it were to start with a `xknx.management.connect`.

I suppose we could always cache connections inside `xknx.management` - but that could easily get tricky (e.g figuring out which connection to retrieve etc).

I'm partial to making all procedure accept an ABC the shape of `P2PConnection` and another ABC the shape for broadcasting? Consumers can bring their own connection/broadcasting implementation or use the ones of XKNX directly and we can still provide higher level/user friendly ways to execute these procedures on the `XKNX` object.

xknx/management/procedures/__init__.py
kewde: [8375e48](https://github.com/XKNX/xknx/pull/1831/commits/8375e487964dae655d6b8e515fae4f1cb39eadfe)

xknx/management/procedures/network/nm_individual_address_write.py
```diff
@@ -0,0 +1,97 @@
...
+            "Checking if device exists at %s and restarting it.", individual_address
+        )
+
+        try:
+            await connection.request(
+                payload=apci.DeviceDescriptorRead(descriptor=0),
+                expected=apci.DeviceDescriptorResponse,
+            )
+        except ManagementConnectionTimeout as ex:
+            # if nothing is received (-> timeout) IA is free
```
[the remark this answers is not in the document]
kewde: I'm keeping it as-is for now, I have a PR that prefixes this procedure with the text from the specification, we can see what copilots thinks in that PR 👍

xknx/management/procedures/network/nm_individual_address_check.py
kewde: let's continue that discussion [here](https://github.com/XKNX/xknx/pull/1831#discussion_r3255252363)

xknx/management/procedures/device/dm_restart_r_co.py
farmio: I agree we'll have to run multiple procedures in one connection context. So we should model this in a way that is mostly readable and easy to understand. Imho that's more important than backwards compatibility at this point - as xknx management procedures are more of a proof of concept currently anyway (and we can always up to xknx 4.0 if we like).

`dm_restart` for example seems not to use the `dm_connect` procedure (with reading DeviceDescriptor), but still use a connection-oriented context. So adding it to `P2PConnection` also doesn't seem perfect.
There are some occasions of explicit
> DM_Connect shall be executed before executing this Management Procedure.

in the specs.

I could imagine something like
```py
individual_address = IndividualAddress(address)
async with XKNX() as xknx:
    async with xknx.management.connection(address=individual_address) as conn:
        await dm_connect(conn)  # this does DeviceDescriptor, but omits the actual connection as it is passed in already
        await dm_authorize(conn, key="akey")
        # ... do other things in connection context
```

It's not the most beautiful solution, but if mixed connection and broadcast is needed, we can use the `xknx` object of a `P2PConnection` to just send that broadcast.

Maybe you find a nicer solution, that's just what came to my mind now.

kewde: Sounds good to me, I think that pattern should suffice and hold up reasonably against many of the procedures. I'll give this a shot in follow up PR.

farmio:
Great 👍 Thank you very much!

PR #1832, opened 2026-05-19
refactor(management): add composable connection-based procedure cores

kewde:
## Description

Adds composable "conn" variants of two management procedures, so several procedures can be chained over one already-open `P2PConnection` instead of each one opening and closing its own:

- `dm_restart_r_co(conn)` — the actual KNX v02.01.02 - Management Procedures 03.05.02 - §3.7.3 procedure name for the connection-based variant of DM_Restart. `dm_restart(xknx, address)` now just opens a connection and delegates to it.
- `nm_individual_address_check_conn(conn)` — an xknx-only naming convention (not a KNX spec name) for the connection-based core of `nm_individual_address_check`. `nm_individual_address_write` now calls this directly on its own already-open connections instead of duplicating the DeviceDescriptorRead/timeout handling inline.

All existing top-level procedures (`dm_restart`, `nm_individual_address_check`, `nm_individual_address_write`, `nm_individual_address_read`, `nm_individual_address_serial_number_read`/`_write`) keep their familiar `(xknx, ...)` signature — no breaking change to any of those call sites.

Also:
- Add `P2PConnection.send_data(payload, wait_for_ack=True)`, replacing the near-duplicate `send_data_no_ack`/private `_send_data` pair, so `dm_restart_r_co` and the request-based procedures share one send path instead of two copies of the same `TDataConnected` construction.
- Remove the `nm_invididual_address_write` typo alias — the misspelled name is no longer exported. This is the only breaking change in this PR.
- Add `CLAUDE.md` documenting how to cite the KNX Standard in code (document + version, since different KNX documents version independently), the management procedure naming convention, changelog conventions, and PR template usage.

Fixes # (issue)

## Type of change

## Checklist


test/management_tests/procedures/network/test_nm_individual_address_write.py
```diff
@@ -139,7 +139,7 @@ async def test_nm_individual_address_write_two_devices_in_programming_mode(
 
     task = asyncio.create_task(
         nm_individual_address_write(
-            xknx=xknx, individual_address=individual_address_new
+            xknx.management, xknx.management, individual_address_new
```
kewde: I don't like it, but there's no union & for python 😢

xknx/management/procedures/network/nm_individual_address_write.py
```diff
@@ -3,38 +3,46 @@
...
 
-if TYPE_CHECKING:
-    from xknx import XKNX
-
 logger = logging.getLogger("xknx.management.procedures")
 
 
 async def nm_individual_address_write(
-    xknx: XKNX, individual_address: IndividualAddressableType
+    manager: ConnectionManager,
```
kewde: ## Problem: procedures that need multiple connections

Some procedures require sequential A_Connects to different addresses:

- **NM_IndividualAddress_Write** — 2 serial A_Connects (occupancy check + verify/restart).
- **NM_Router_Scan** — loop of up to 256 A_Connects, one per subnet address 0–255.
- **NM_DomainAddress_Scan / NM_DomainAddress_Scan2** — same pattern.

These can't take `conn: P2PConnection` — they need a connection factory.

## Options

#### 1. Move `address` into `connect()`

`P2PConnection` becomes a reusable session object; `address` is passed at connect-time:

```python
# procedure signature
async def nm_router_scan(conn: P2PConnection) -> list[IndividualAddress]:
    for i in range(256):
        async with conn.connect(IndividualAddress(f"1.{i}.0")):
            ...

# caller
conn = P2PConnection(xknx)
await nm_router_scan(conn)
```

#### 2. Pass `ConnectionManager` to scanning procedures

Remove the `raise` guard in `Management.connect()` — allow reconnecting to the same or a new
address — and let scanning procedures own their loop:

```python
async def nm_router_scan(manager: ConnectionManager) -> list[IndividualAddress]:
    for i in range(256):
        async with manager.connection(IndividualAddress(f"1.{i}.0")) as conn:
            ...

# caller
await nm_router_scan(xknx.management)
```

This is already the pattern for `nm_individual_address_write` — scanning procedures are just
another case of a procedure that manages its own connection lifecycle.

xknx/management/management.py
```diff
@@ -304,6 +310,21 @@ def process(self, telegram: Telegram) -> None:
         self._response_waiter.set_result(telegram)
         self._expected_sequence_number = self._expected_sequence_number + 1 & 0xF
 
+    async def send_data_no_ack(self, payload: APCI) -> None:
```
kewde: to provide a clean way to do restarts

test/management_tests/procedures/network/test_nm_individual_address_write.py
farmio: Why do we need to separate P2P from Broadcast here when we pass `management` anyway - which contains both. Imho the separation of those 2 in different Protocols looks like preemptive optimisation to me. We have the `Management` class we can just use for that, can't we?

So I wonder, why the Protocols when we have the real class definitions we can use for type annotation anyway?

kewde: > Why do we need to separate P2P from Broadcast here when we pass management anyway - which contains both. Imho the separation of those 2 in different Protocols looks like preemptive optimisation to me.

Not an optimisation but more a big API design decision 😁 

The idea behind it is there are multiple procedures that require different sets of "capabilities":
* some only need broadcast -> can just take broadcaster as protocol (aka Management class)
* some are connection oriented ->can just take p2p protocol (result of Management.connection())
* some are connectioned oriented but need multiple conns -> need the connection manager (aka Management class)

the choice for protocols was mainly to allow any third-party library to bring their own version of the `Management` class - we're not forcing the usage of that class - want to manage your own connections? Go ahead.

farmio: I see, but I'm wondering if this is a real-world scenario. A third-party could just inherit from Management and change/add whatever it needs, couldn't it? Also, Imho this can at any point be transformed into protocol usage if really needed. Right now, I think it adds unnecessary indirection and is just another thing to think about when we wanted to do changes in the Management class (eg. changing return type of `request()`).

Anyway, if you think Protocols need to be there, I'd make `Management` explicitly inherit from them, so it is super obvious that there is correlation.
Also a combined broadcast + connection protocol could be added so we don't need the duplicate arguments. 

Or it could just be a single Protocol / ABC - if a third-party wants to bring it's own management class, I guess it's fine to just have it manage both, connection and broadcast.
The 2nd point - the P2P connection makes perfect sense... this is different from the factory / communication type.

farmio:
Claude Code 🤖:

I like where this is heading — splitting "procedure logic" from "connection lifecycle" is the right move, and `dm_restart(conn)` reads much better than the old telegram-crafting inline. A few things I'd change before this lands, all with "what does the call site look like for a user?" as the yardstick.

### 1. Drop `protocols.py` — annotate with the real classes

Re: the Protocol discussion above — I'm with @farmio here. Concretely:

* `xknx.management.protocols.P2PConnection` and `xknx.management.management.P2PConnection` are two different things with the same name. Anyone reading `conn: P2PConnection` now has to check *which* import it is. That alone is a strong argument.
* The protocols are `@runtime_checkable`, but nothing does `isinstance()` on them — and for `P2PConnection` (which has a data member `address`) `issubclass()` wouldn't work anyway. So the decorator buys nothing.
* "third parties can bring their own Management" is a real use case, but it's already served by subclassing `Management` — and if we ever *do* need structural typing, converting `Management` → `Protocol` later is a typing-only change. It's not a decision we have to make now, and making it now costs indirection on every procedure and one more file to keep in sync whenever `request()` or `connection()` changes.

So: `conn: P2PConnection`, `management: Management` — the actual classes.

### 2. Never pass the same object twice

```python
nm_individual_address_write(xknx.management, xknx.management, individual_address_new)
```

This is the clearest signal that the split is at the wrong seam. `Management` *is* both the connection manager and the broadcaster; the two-argument form exists only to describe the protocol split from (1). With concrete types this collapses to:

```python
await nm_individual_address_write(xknx.management, "1.2.3")
```

Same for `nm_individual_address_read(xknx.management, ...)` etc. — one `management` argument everywhere a procedure owns its own connections, one `conn` argument where it operates on an established one. That's a rule a user can hold in their head.

### 3. No legacy layer, please — let's just break it

The management procedures have very few users, and we're happy to cut a major release for an API change here. That's much cheaper for everyone than what's currently in the PR:

* the `_legacy` modules are more code than the procedures themselves, and we'd have to carry and test them for a couple of releases;
* they force a second public name for every procedure;
* and today `from xknx.management.procedures import dm_restart` and `from xknx.management.procedures.device import dm_restart` are two different functions with the same name and incompatible signatures, differing only by import path — that's a trap no IDE or traceback will help with.

So: drop `device/legacy.py`, `network/legacy.py` and the module-level aliasing, keep one public name per procedure with the new signature, and describe the change in `docs/changelog.md` with before/after snippets. While we're at it, the `nm_invididual_address_write` typo alias can go too.

### 4. `nm_individual_address_check(conn)` leaks its own semantics

The procedure is specified as "is `IA_new` occupied?", and per 03.05.02 §2.3 an `A_Disconnect` counts as occupied. After the split, part of that answer lives *outside* the function — the `ManagementConnectionRefused` from `connect()`/`disconnect()` — which is why the legacy wrapper has to re-implement it:

```python
except ManagementConnectionRefused:
    return True
```

and why `nm_individual_address_write` needs this:

```python
try:
    async with manager.connection(individual_address) as conn:
        address_found = await nm_individual_address_check(conn)
except ManagementConnectionRefused:
    pass
```

Two callers, two different handlings of the same condition, and in the second one `address_found` is only bound if the exception came from `disconnect()` in the `finally` — if it ever comes from anywhere earlier, the next line raises `UnboundLocalError`. Even if that path isn't reachable today, it shouldn't depend on *where* inside the context manager the exception originates.

Suggestion: `nm_individual_address_check` keeps owning its connection (`management, address`), so "refused == occupied" stays in one place. If a raw "does this device answer on an open connection" primitive is useful for scanning, give it its own name (`device_responds(conn)` / `dd_read_check(conn)`) rather than overloading the spec-named procedure. At minimum, set `address_found = True` in that `except` and drop the reliance on the earlier assignment.

### 5. `P2PConnection.send_data_no_ack()` — 👍

Good addition, this is exactly the kind of thing that belongs on the connection rather than being open-coded in two procedures. Only bikeshed: `send_data_no_ack` next to the private `_send_data` reads a bit like an internal; `send_unacknowledged()` or `send_data(payload, wait_for_ack=False)` might age better. Your call.

### Suggested target shape

```python
# procedures that talk to one device and own the connection
await dm_restart(xknx.management, "1.2.3")
await nm_individual_address_check(xknx.management, "1.2.3")
await nm_individual_address_write(xknx.management, "1.2.3")

# broadcast-only
await nm_individual_address_read(xknx.management)

# power users / multiple procedures on one connection
async with xknx.management.connection(IndividualAddress("1.2.3")) as conn:
    await dm_restart_r_co(conn)
```

i.e. the `conn`-taking functions stay (they're the composable core, and scanning procedures like `NM_Router_Scan` will want them), but the spec-named public procedures keep the one-liner ergonomics. On your `NM_Router_Scan` question above: option 2 (pass `Management`, loop over `manager.connection(...)`) — a connection object that can be re-pointed at another address (option 1) means `conn.address` is no longer stable, and every procedure taking `conn` would have to care about that.

xknx/management/procedures/network/nm_individual_address_write.py
farmio: In my opinion Option 2 is a cleaner API.
Regarding removing the raise:
```py
if address in self._connections:
            raise ManagementConnectionError(f"Connection to {address} already exists.")
```
We remove the adddress from `self._connections` on disconnect. Parallel connections aren't possible iirc. So I think we can - and should - keep this. Connections to new addresses are possible anyway.

I could imagine to have some kind of backoff retry mechanism if an address is already connected by another procedure. But that's out of scope for this PR. And maybe better handled by the user of the API.

farmio:
Personally I'd drop the Protocols and the Legacy shims. It should be fine to break these APIs and do a major release when we are ready.

xknx/management/procedures/network/nm_individual_address_check.py
```diff
@@ -15,35 +16,52 @@
...
-    except ManagementConnectionRefused as ex:
-        # if Disconnect is received immediately, IA is occupied
-        logger.debug("Device does not support transport layer connections. %s", ex)
-        return True
+            IndividualAddress(individual_address)
+        ) as conn:
+            address_found = await nm_individual_address_check_conn(conn)
+    except ManagementConnectionRefused:
+        address_found = True
+    return address_found
```
farmio:
```suggestion
try:
async with xknx.management.connection(
            IndividualAddress(individual_address)
        ) as conn:
            return await nm_individual_address_check_conn(conn)
    except ManagementConnectionRefused:
        return True
```
we could simplify a bit here

farmio (approved):
I think this looks very clean! Like it 👍

xknx/management/procedures/network/nm_individual_address_check.py
kewde: [c58c45e](https://github.com/XKNX/xknx/pull/1832/commits/c58c45e7dc7a801362843f0fd7e5f205d80bdabe)

PR #1833, opened 2026-05-19
feat(management): implement dm_authorize and dm_connect procedures

kewde:
## Description

Adds `dmp_authorize_r_co`, `dmp_authorize2_r_co` and `dm_connect` to
`xknx/management/procedures/device/`, alongside the existing
`dm_restart`/`dm_restart_r_co`.

- **`dmp_authorize_r_co(connection, key)`** — DM_Authorize (KNX v02.01.02 -
  Management Procedures 03.05.02 - §3.5.1): sends a single
  `A_Authorize_Request` with the given key and returns the granted access
  level.
- **`dmp_authorize2_r_co(connection, client_key)`** — DM_Authorize2 (§3.5.2):
  authorizes with `FREE_ACCESS_KEY` first, then `client_key`, and keeps
  whichever gave the better (lower) access level — for the case where the
  caller doesn't know in advance whether the device is already locked with
  `client_key` (e.g. a mix of provisioned and factory-fresh devices).
- **`dm_connect(conn)`** — DMP_Connect_RCo (§3.2.1): reads Device Descriptor
  Type 0 on an already-open connection to confirm it's live, returning the
  mask version.

All three take an already-open `P2PConnection` and have no `xknx`-opening
counterpart, by design: unlike `dm_restart`, an authorization's access level
(and DM_Connect's whole point of existing) only has meaning for the
lifetime of the connection it ran on, so opening a connection just to
authorize or connect and immediately closing it again would discard the
result. They're meant to be composed with other `_r_co`/`_conn` procedures
under a caller-managed `xknx.management.connection()`.

Rebased onto current `main` (this predates the composable connection-based
procedure cores refactor in #1832 - `P2PConnection` now lives in
`xknx.management.management`, not the old `xknx.management.protocols`).

Fixes # (issue)

## Type of change

## Checklist


xknx/management/procedures/device/dm_authorize.py
```diff
@@ -0,0 +1,55 @@
...
+
+from xknx.management.protocols import P2PConnection
+from xknx.telegram import apci
+
+FREE_ACCESS_KEY = 0xFFFFFFFF
+
+__all__ = ["FREE_ACCESS_KEY", "dmp_authorize2_r_co", "dmp_authorize_r_co"]
+
+
+async def dmp_authorize_r_co(connection: P2PConnection, key: int) -> int:
```
farmio: should we default to `FREE_ACCESS_KEY` or use `key: int | None` and use FREE_ACCESS_KEY if None then?

kewde: [438d564](https://github.com/XKNX/xknx/pull/1833/commits/438d5642633ae11bc34e62b30f3793ddb5ae57eb)

xknx/management/procedures/device/dm_connect.py
```diff
@@ -0,0 +1,28 @@
...
+
+from __future__ import annotations
+
+from xknx.management.management import P2PConnection
+from xknx.telegram import apci
+
+__all__ = ["dm_connect"]
+
+
+async def dm_connect(conn: P2PConnection) -> int:
```
kewde: `dm_connect_conn`?

xknx/management/procedures/device/dm_authorize.py
```diff
@@ -0,0 +1,59 @@
...
+    response = await connection.request(
+        payload=apci.AuthorizeRequest(key=key),
+        expected=apci.AuthorizeResponse,
+    )
+    # `expected` guarantees this via `P2PConnection._receive`
+    assert isinstance(response.payload, apci.AuthorizeResponse)
+    return response.payload.level
+
+
+async def dmp_authorize2_r_co(connection: P2PConnection, client_key: int) -> int:
```
kewde: `dmp_authorize2_r_co` already has an `_r_co` extension which makes it somewhat naturally that it is connection oriented but that may not be what consumers of the API expect. 

How do we want to deal with xknx vs p2pconnection arg differences?

farmio: hm... maybe we should have defaulted (no-suffix) to the `P2PConnection` variant 😬
I mean, it's not a big deal as we do have type annotations anyway - so whatever we do, it shouldn't come to surprise any user...

I guess in the long run, we probably want to add some sequences of connect, authorize, read memory, read properties, write something, read again, disconnect ... so chaining procedures will probably be a common thing, wouldn't it?

farmio:
Claude Code  🤖:

Checked out the branch (based on `35cc2867`), ran `pytest test/management_tests/` (27 passed), mypy and ruff — all green. Verified the `dmp_authorize2_r_co` algorithm against `03_05_02 Management Procedures v02.01.02 AS.pdf` §3.5.2: the free-first / client-key / re-auth-if-worse flow and the `free_level == 0` short-circuit ("If the free access level is not the highest level (lowest numerical value, maximum access rights)") match the spec sequence exactly. That core is right — the points below are around it.

### Blocking

**1. The new procedures are not re-exported from `xknx.management.procedures`**

`device/__init__.py` was updated, but `procedures/__init__.py:40` still reads `from .device import dm_restart, dm_restart_r_co`. On the branch:

```
>>> from xknx.management.procedures import dmp_authorize_r_co
ImportError: cannot import name 'dmp_authorize_r_co' from 'xknx.management.procedures'
```

The changelog entry added here says the functions are added "to `xknx.management.procedures`", and the package docstring promises "This package re-exports every implemented procedure". `dm_connect` and `FREE_ACCESS_KEY` are unreachable there too. Please add them to that import line.

**2. The busy-wait loops in the tests hang the suite instead of failing it**

`while xknx.cemi_handler.send_telegram.call_count < N: await asyncio.sleep(0)` (lines 150, 168, 205, 223, 263, 281, 299) has no timeout. Reproduced: change `if free_level == 0:` to `if free_level <= 15:` in `dmp_authorize2_r_co` — so the second authorize is never sent — and `test_dmp_authorize2_r_co_client_key_is_better` spins forever: the procedure returns, then `await responder` never completes because the responder is stuck at `call_count < 3`. pytest had to be killed externally (exit 124, no output at all). The `# noqa: ASYNC110` comments suppress exactly the lint rule that flags this. Please bound each wait with `asyncio_timeout(...)`, or drive the responder off an `asyncio.Event`, so a regression fails fast rather than hanging CI.

**3. The `dmp_authorize2_r_co` tests never assert *which key* was sent**

They assert only the returned level. Reproduced: replacing the final re-authorization

```python
await dmp_authorize_r_co(connection, FREE_ACCESS_KEY)
```

with `await dmp_authorize_r_co(connection, client_key)` — the exact step §3.5.2 exists to specify ("If free access gave higher access level (lower numerical value), a new authorisation with free access level is done") — leaves all 6 tests passing. The two single-shot tests do assert `call_args_list`; the authorize2 tests should do the same, e.g. compare the final `send_telegram.call_args_list` against the expected `AuthorizeRequest` telegrams with their sequence numbers.

### Worth fixing before merge

**4. `dm_connect` breaks the package's documented naming convention** — `procedures/__init__.py` (lines 20–29) reserves the unsuffixed `<spec_name>(xknx: XKNX, ...)` for the form that "opens (and closes) whatever P2P connections or broadcasts it needs on its own", with `dm_restart` / `dm_restart_r_co` as the in-tree example. `dm_connect(conn: P2PConnection)` inverts that, so a caller following the documented contract writes `await dm_connect(xknx, IndividualAddress("1.1.1"))` and gets a `TypeError` inside `conn.request`. Since §3.2.1's spec name is DMP_Connect_RCo, `dmp_connect_r_co` in `dm_connect_r_co.py` (mirroring `dm_restart_r_co.py`) fits both the spec and the convention. Same for `dm_authorize.py`, which holds `dmp_authorize_r_co`, and for the parameter name — `connection` here vs `conn` in `dm_connect` and `dm_restart_r_co`.

**5. `key: int = FREE_ACCESS_KEY` as the default contradicts §3.5.1** — the spec guards the whole A_Authorize exchange with `if authorization is required (key != FFFF FFFFH) ... endif`, and §3.5 adds that authorization "shall be executed only when it is required by the Management Server". So `await dmp_authorize_r_co(conn)` — which reads as "no authorization needed" — sends a request the spec says must not be sent; against a device that doesn't implement A_Authorize at all that means `_receive` blocking for `MANAGAMENT_CONNECTION_TIMEOUT` (6 s) and raising `ManagementConnectionTimeout` where the spec wants a no-op. `dmp_authorize2_r_co` genuinely must send the free key (§3.5.2: the client "re-authorizes explicitly with key FFFFFFFFh"), so that belongs on an internal helper; the public §3.5.1 entry point should either drop the default or document the deviation.

**6. `dm_connect` doesn't check the descriptor type of the response** — `expected=apci.DeviceDescriptorResponse` matches the class only, not `descriptor`, so a `DeviceDescriptorResponse(descriptor=2, value=...)` is returned as what the docstring and changelog both call "the mask version (DD0 value)", and a caller comparing it against known mask versions mis-identifies the device. §3.2.2 explicitly requires the connectionless sibling to be "capable of handling both DD-types expected and unexpected"; one `if response.payload.descriptor != 0: raise ManagementConnectionError(...)` covers it.

**7. The re-auth response level is discarded** (`dm_authorize.py:57`) — after re-authorizing with `FREE_ACCESS_KEY` the function returns the cached `free_level` from *before* the client-key attempt, not the level the device is granting now. `return await dmp_authorize_r_co(connection, FREE_ACCESS_KEY)` reports what actually happened, at no extra cost.

**8. Misleading comment** (`test_dm_authorize.py:247`) — `# connect + 2x authorize` is wrong: `reset_mock()` on line 201 cleared the connect telegram, and the 4 is 2 requests + the 2 ACKs `Management.process` backgrounds for the incoming responses. Someone adding a step later will reason from the comment and adjust the number wrongly, silently weakening the only assertion that checks no re-authorization happened.

kewde:
Addressed in f7c8d3a4:

**Blocking**
1. `dmp_authorize_r_co`, `dmp_authorize2_r_co`, `dmp_connect_r_co`, `FREE_ACCESS_KEY` now re-exported from `xknx.management.procedures` (`procedures/__init__.py`).
2. The three responder busy-wait loops in `test_dm_authorize.py` are now bounded with `asyncio_timeout(1)`. Verified: reverting `free_level == 0` to `free_level <= 15` (so the second request never gets sent) now fails `test_dmp_authorize2_r_co_client_key_is_better` in ~1s with a clear `TimeoutError`, instead of hanging.
3. All three multi-step tests now assert the exact outgoing telegrams (including which key each `AuthorizeRequest` carries), not just the returned level. Verified: swapping the re-auth call's key back to `client_key` now fails `test_dmp_authorize2_r_co_free_access_is_better`.

**Worth fixing before merge**
4. `dm_connect.py`/`dm_connect` renamed to `dm_connect_r_co.py`/`dmp_connect_r_co` (matches DMP_Connect_RCo, mirrors `dm_restart_r_co.py`). `dm_authorize.py`'s `connection` param renamed to `conn` to match.
5. Dropped `dmp_authorize_r_co`'s `key: int = FREE_ACCESS_KEY` default - §3.5.1 only calls for the exchange when authorization is required at all, so defaulting to the free key would send a request the spec says shouldn't be sent, and would time out against a device without A_Authorize. `dmp_authorize2_r_co` already passed the key explicitly everywhere, so this doesn't change its behavior.
6. `dmp_connect_r_co` now checks `response.payload.descriptor == 0` and raises `ManagementConnectionError` otherwise, with a test covering a DD-type-2 response.
7. The final re-auth branch of `dmp_authorize2_r_co` now returns the actual response from the re-authorization instead of the cached `free_level`.
8. Fixed the misleading comment - it's 2 `AuthorizeRequest`s + 2 background ACKs `Management.process` sends for the incoming responses, not "connect + 2x authorize".

All of `test/management_tests/` (38 tests), mypy, pylint and ruff are green on the branch.

PR #1834, opened 2026-05-20
feat(management): implement dmp_interface_object_*_r

kewde:
## Description

 Add `dmp_interface_object_read_r` (KNX 03.05.02 §3.27.2), `dmp_interface_object_write_r` (§3.25.2), `dmp_interface_object_scan_r` (§3.28.2), and `dmp_interface_object_verify_r` (§3.26.2) to the `device/` procedure package. All procedures take a `P2PConnection` directly. Read, write, and verify handle chunking transparently for payloads exceeding 15 elements (4-bit `nr_of_elem` limit); verify compares per block as required by the spec. Scan enumerates all interface objects and their property descriptions via `PropertyDescriptionRead`/`PropertyValueRead` loops, returning a list of `ScannedInterfaceObject`

Fixes # (issue)

## Type of change

## Checklist


xknx/management/procedures/device/dmp_interface_object_write_r.py
```diff
@@ -0,0 +1,112 @@
...
+
+from __future__ import annotations
+
+from xknx.exceptions import ManagementConnectionError
+from xknx.management.management import P2PConnection
+from xknx.telegram import apci
+
+# nr_of_elem field is 4 bits (KNX v02.01.01 - Application Layer 03.03.07 -
+# §3.4.4 Figure 48), max value 2^4 - 1
+MAX_ELEMENTS_PER_REQUEST = (1 << 4) - 1
```
kewde: dedupe

xknx/management/procedures/device/dmp_interface_object_read_r.py
```diff
@@ -0,0 +1,87 @@
...
+
+from __future__ import annotations
+
+from xknx.exceptions import ManagementConnectionError
+from xknx.management.management import P2PConnection
+from xknx.telegram import apci
+
+# nr_of_elem field is 4 bits (KNX v02.01.01 - Application Layer 03.03.07 -
+# §3.4.4 Figure 46), max value 2^4 - 1
+MAX_ELEMENTS_PER_REQUEST = (1 << 4) - 1
```
kewde: dedupe

xknx/management/procedures/device/dmp_interface_object_scan_r.py
```diff
@@ -0,0 +1,103 @@
...
+
+from dataclasses import dataclass, field
+
+from xknx.exceptions import ManagementConnectionError
+from xknx.management.management import P2PConnection
+from xknx.telegram import apci
+
+# KNX v02.01.01 - Application Interface Layer 03.04.01 - §4.3.2.3: the
+# Property with property_id = 1 is always PID_OBJECT_TYPE
+_PID_OBJECT_TYPE = 0x01
```
kewde: another place we should put this?

xknx/management/procedures/device/dmp_interface_object_read_r.py
```diff
@@ -0,0 +1,87 @@
...
+    :param start_index: Start element index (1-based, 1-4095)
+    :return: The property data read from device
+    :raises ManagementConnectionError: If device returns error (nr_of_elem = 0)
+    """
+    # The spec's sequence (A_PropertyDescription_Read) is intentionally skipped:
+    # callers supply object_index/property_id/count/start_index explicitly, so the
+    # property is always known at this level. Discovery belongs to the Configuration
+    # Procedure above (per spec: "shall not interpret ... at the level of this
+    # Management Procedure"). Use dmp_interface_object_scan_r to discover properties
+    # (type, max_count, access) before calling this function.
```
kewde: I'm going a bit off spec here. I suppose we can make some arguments optional in the function to make it fully spec compliant version, but that muddies the types a bit.

IMHO i like the type checker catching "None" in calls to arguments, and I think `dmp_interface_object_scan_r` is the better primitive tbh.

But it's not a hill i'll die upon, happy to change this, accept optional values and do a scan variant here.

xknx/management/procedures/device/dmp_interface_object_scan_r.py
farmio: we already have this at `xknx.profile.const`

xknx/management/procedures/device/dmp_interface_object_read_r.py
farmio: so could we add a convenience function that takes a property_id , scans all the object interface, looks for the passed id and pipes its parameters to read - returning bytes or raising if not found?

xknx/management/procedures/device/dmp_interface_object_write_r.py
kewde: Moved to a shared `device/const.py` in 07e4bb84 — the two copies also cited different spec figures (46 vs 48) for the same 4-bit field, now folded into one comment covering both.

xknx/management/procedures/device/dmp_interface_object_read_r.py
kewde: Same fix — see 07e4bb84, moved to `device/const.py`.

xknx/management/procedures/device/dmp_interface_object_scan_r.py
kewde: done

farmio (commented):
Claude Code  🤖:

Checked out locally, verified against `03_05_02 Management Procedures v02.01.02`, `03_03_07 Application Layer v02.01.01` and `03_04_01 Application Interface Layer v02.01.01`.

**Clean:** full suite passes (3867), ruff / ruff format / mypy clean. Spec fidelity is genuinely good — `scan_r` follows the §3.28.2 sequence exactly (PID=0/property_index=0 probe, `PID <> 0` existence test, PID=01h/start_index=01h/count=01h type read), `verify`'s per-chunk comparison correctly reflects that §3.26.2 puts the comparison *inside* the loop, the `verify` kwarg is a faithful mapping of §3.25.2's flags bit 1, and the citation format/versions match AGENTS.md. Changelog placement is right.

## Correctness

**1. Silent wire corruption when `start_index + count - 1 > 4095`** — the one I'd block on.

`start_index` is 12 bits packed into the same octet as `count`, and `to_knx` bounds-checks only `count`. Nothing in the three chunking procedures validates `start_index`, even though all three docstrings promise "1-4095". Verified:

```
PropertyValueWrite(object_index=3, property_id=78, count=5, start_index=4105, data=b"\xaa"*5)
→ decodes as count="6" start_index="9"
```

No exception. So `dmp_interface_object_write_r(..., count=20, start_index=4090)` sends its second chunk to elements 9–14 of the property instead of 4105+ — a write landing on entirely the wrong data, with a response the `response_count != chunk_count` guard won't reliably catch. Fix in the procedures (`1 <= start_index` and `start_index + count - 1 <= 0xFFF`), and/or add the missing bounds check to `to_knx` next to the existing count check — that's the natural sibling and fixes it library-wide.

**2. Chunking handles the element-count limit but not the APDU length limit.** 15 elements of a 4-byte property is 60 data bytes; xknx serializes that happily (I get a 75-byte cEMI frame, no complaint), and the device answers `nr_of_elem=0` — AL §3.4.4.1 lists "the data does not fit in a PDU" as exactly that error. The procedure then raises instead of retrying smaller. So "transparently chunks" only really holds for 1-byte elements. The write path knows `element_size` and could cap `chunk_count` accordingly; the read path would need an `element_size`/`max_bytes` argument (or PID_MAX_APDULENGTH). At minimum say so in the docstrings, because the PR body and changelog both claim it's transparent.

**3. The partial-response guard rejects a spec-legal response.** AL §3.4.4.1 error handling: a read with `start_index = 0` and `nr_of_elem > 1` *shall* be answered with `start_index = 0, nr_of_elem = 1` carrying the current element count. `read_r`/`verify_r` raise on that. Narrow (defaults are `start_index=1`, and `count=1` still works), but if `start_index=0` isn't meant to be supported, reject it up front rather than surfacing it as a device error.

## API surface / conventions

**4. `scan_r` and `verify_r` are never exported.** `device/__init__.py` imports only read and write, and none of the four reaches `procedures/__init__.py` — whose docstring says the package "re-exports every implemented procedure", and whose changelog entry says all four land in `xknx.management.procedures.device`. This fails worse than an ImportError: `from xknx.management.procedures.device import dmp_interface_object_scan_r` silently binds the *module* of the same name, so calling it gives `TypeError: 'module' object is not callable`. `ScannedInterfaceObject` isn't exported anywhere either.

**5. No `__all__`** in any of the four modules; `dm_authorize.py` and `dm_connect_r_co.py` both have one.

**6. Parameter is `connection`**, every existing procedure uses `conn`.

**7. Naming vs. the documented convention.** Per AGENTS.md, `_conn` marks the form that takes an open connection, with `dm_restart_r_co` the single exception because `RCo` genuinely is the KNX name for DM_Restart's connection-based variant. `_R` isn't that — §3.25.2/§3.27.2 explicitly say "connection oriented **or connectionless**". By the convention these should be `dmp_interface_object_read_r_conn(conn, ...)`. And unlike authorize/connect — where the changelog explains why no `(xknx, ...)` form exists — read/write/scan/verify are perfectly meaningful as one-shot operations on an address, so the `(xknx, individual_address, ...)` counterpart is the obvious missing half.

## Tests

**8. The chunking is never verified on the wire.** No test asserts on outgoing telegrams; the chunked tests only check the assembled return value, which the responder itself supplies. Nothing pins the second request to `start_index=16, count=5`, so an off-by-one in `current_index` passes today. `test_dm_authorize.py` already does this with `send_telegram.assert_has_calls([call(...)])` — this is the core logic of the PR and it's the one thing untested.

**9. Responder loops have no `asyncio_timeout`.** The sibling file defines `RESPONDER_TIMEOUT = 1` and wraps every `while call_count < N: await asyncio.sleep(0)`. Without it, a regression that stops sending spins forever and hangs the suite instead of failing it.

Also, ~1300 lines of test for ~400 of code, because each file re-inlines full `Telegram` construction instead of using `_ack()`/`_response()` builders like the neighbours.

## Minor

- `count <= 0` silently returns `b""` / `None` in all three — a `ValueError` matches the `len(data) % count` handling right below it.
- `read_r`'s `:raises:` doesn't mention the partial-response error.
- `scan_r`'s loops are unbounded, but `object_index`/`property_index` are single octets: a device that never answers PID=0 dies at 256 with a raw `struct.error` instead of `ManagementConnectionError`. Cap at 255.
- `int.from_bytes(type_resp.payload.data, "big")` accepts any length; PID_OBJECT_TYPE is 2 octets, so a short or long response yields a silently wrong `object_type`.
- The comment "PID_OBJECT_TYPE is always property_id 1, index 0 … (always 1 element, 2 bytes)": §4.3.2.3's "index 0" is the *property index*, not the array element index, and PID_OBJECT_TYPE's `max_nr_of_elem` may exceed 1 for array Interface Objects. Reading element 1 is what §3.28.2 prescribes — it's the parenthetical that's off.
- `verify_r` signals a mismatch — its normal negative outcome — with the same `ManagementConnectionError` as a transport failure, so callers can't tell them apart.

xknx/management/procedures/device/dmp_interface_object_read_r.py
kewde: After reading the specification again, I think this is fine. The spec for this management procedure explicitly passes all the values as arguments.

farmio:
Claude Code  🤖:

Re-checked at b24b59ae — suite green (3876), ruff/format/mypy clean. Most of the review is addressed; putting the `start_index` bounds check in `to_knx` rather than in the procedures was the right call. Two things left from my list, plus one that's now moot.

### Naming (my point 7) — moot

Ignore it. The `_conn` suffix / `(xknx, individual_address, ...)` question is farmio's and yours to settle, not something you should have to guess at from a review comment.

### Still open: the APDU length limit

Nothing in the four modules mentions it. The chunking caps at 15 elements because `nr_of_elem` is 4 bits, but the other limit is the APDU, and it bites first for anything wider than a 1-byte element: 15 elements of a 4-byte property is 60 data bytes, which xknx serializes without complaint (I get a 75-byte cEMI frame), and the device then answers `nr_of_elem=0` — KNX v02.01.01 - Application Layer 03.03.07 - §3.4.4.1 lists "the data does not fit in a PDU" as exactly that error. So the procedure raises instead of retrying smaller, and the caller gets a generic device error with no hint that a smaller `count` would work.

The write path already knows `element_size` and could cap `chunk_count` by a byte budget; the read path can't infer it, so it would need an `element_size`/`max_bytes` argument (or PID_MAX_APDULENGTH). Even just a docstring sentence would help — as it stands the changelog promises the chunking is transparent, and for multi-byte properties it isn't.

### New nit, from the fix to #1

The bounds check lives only in `to_knx`, so the procedures still let `start_index + count - 1` run past 4095 and only fail once the offending chunk is encoded:

```
dmp_interface_object_write_r(conn, ..., count=20, start_index=4090)
→ chunk 1 (start_index=4090, count=15) is written to the device
→ chunk 2 (start_index=4105) raises ConversionError: Start index out of range.
```

Two edges there: the write is left half-applied, and `ConversionError` is an `XKNXException` but *not* a `ManagementConnectionError`, so it slips past callers catching the type the docstring documents. Validating `start_index + count - 1 <= 0xFFF` alongside the existing `start_index < 1` check would fail before any telegram goes out. The `to_knx` guard is still worth keeping as the backstop.

Related: `write_r` doesn't validate `start_index >= 1` the way `read_r`/`verify_r` now do. That's probably deliberate — writing element 0 resets the array per KNX v02.01.01 - Application Interface Layer 03.04.01 - §4.3.2.3 — but its docstring still says "Start element index (1-based, 1-4095)", so either the check or the docstring should move.

farmio:
Claude Code  🤖:

072f517e crossed my last comment — it answers both points in it, and then some. Checked it over: `max_apdu_length` on all three procedures with the chunk sized as `min(15, max(1, (max_apdu_length - 5) // element_size))`, `read_r` probing one element first because it can't know the size upfront, and the `start_index + count - 1 <= 0xFFF` pre-check in all three so an out-of-range transfer fails before the first telegram instead of halfway through. `read_r` now *supporting* `start_index=0` (restricted to `count=1`, so it's the only loop iteration) is a nicer resolution than the rejection I'd suggested.

I checked the byte arithmetic against the spec, since that's where this would quietly be off by two: the `PID_MAX_APDU_LENGTH` citation matches the real document (KNX v01.10.01 - Resources 03.05.01 - §4.3.7), §4.3.7.2.1 does say a client without that Property manages the device "with L_Data_Standard-frames with an APDU-length of maximal 15 octets", §4.3.7.1 puts the Property's range at 15–254, and those units line up with the LG field and therefore with `calculated_length()` — so the 15 default and the `- 5` header subtraction are both right, and `PROPERTY_VALUE_HEADER_OCTETS = 5` matches `calculated_length()` exactly. Locally: 3889 passed, ruff/format/mypy clean (the one `routing_multicast_test` failure reproduces on `main` — no multicast route on this machine, unrelated).

The implementation I'd call ready. What's left is around it.

### Blockers

**1. Conflicts with `main`.** GitHub reports the PR as conflicting; it's only `docs/changelog.md`, but not a trivial resolve — `main` has since cut 3.20.0 and shortened the changelog, so the entry has to move into a fresh `# Unreleased changes` section that didn't exist when it was written.

**2. The changelog entry is stale on substance, not just position.** It still describes only the original four procedures. Missing since then: the `max_apdu_length` parameter on `read_r`/`write_r`/`verify_r`, `start_index=0` support in `read_r`, `PropertyVerificationError` (now public in `xknx.exceptions.__all__`), and that `PropertyValue*.to_knx()` now raises `ConversionError` for an out-of-range `start_index` where it previously emitted a silently corrupt frame — that last one is a behaviour change in existing API, not just an addition. The entry also still places the procedures in `xknx.management.procedures.device` although they're re-exported a level up now, and its "transparently chunk … 15-element limit" line now undersells what the code does, since chunking is APDU-aware.

**3. `Resources 03.05.01` is missing from the AGENTS.md table.** This PR already added the `03.04.01` row when it introduced that citation; the new `PID_MAX_APDU_LENGTH` references introduce another document and want the same row: `| v01.10.01 | Resources | 03.05.01 |`. I verified that version against the document itself.

### Worth a look, not blocking

**`read_r` now costs an extra round trip for every `count > 1`.** The 1-element probe is the right default given the PDU carries no PDT, but a caller coming from `dmp_interface_object_scan_r` is already holding the `PropertyDescriptionResponse` and knows the element size — an optional `element_size` argument that skips the probe would pay off for exactly the callers the docstrings point at `scan_r`.

**`max(1, ...)` sends an oversized frame rather than refusing one.** If `element_size` exceeds the budget — a 14-octet element against the default 15, giving `(15 - 5) // 14 == 0` — the clamp yields 1 element and a 19-octet PDU that a standard-frame-only device can't accept. The clamp is right to prevent a zero-size infinite loop, but a `ValueError` saying the element doesn't fit in `max_apdu_length` would beat knowingly putting an over-long frame on the bus.

PR #1857, opened 2026-07-20
feat: A_SystemNetworkParameter_Read & A_SystemNetworkParameter_Response

### LINKED ISSUE #1021: A_ADC_Response-PDU or A_SystemNetworkParameter_* not parsed correctly
**Description of problem:**
A_ADC_Response-PDU or A_SystemNetworkParameter_* not parsed correctly.

This was received via a tunnelling connection with an unrelated interface once when first installing a new KNX IP Secure Router to the system. 
I failed to reproduce the exception after that. It probably resulted from a system broadcast message.

- [x] using xknx standalone

**Version information:**
- xknx release with the issue: 1.0.0

```py
  File "/usr/local/lib/python3.10/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/local/lib/python3.10/asyncio/selector_events.py", line 1027, in _read_ready
    self._protocol.datagram_received(data, addr)
  File "/usr/local/lib/python3.10/site-packages/xknx/io/transport/udp_transport.py", line 49, in datagram_received
    self.data_received_callback(data, addr)
  File "/usr/local/lib/python3.10/site-packages/xknx/io/transport/udp_transport.py", line 82, in data_received_callback
    knxipframe.from_knx(raw)
  File "/usr/local/lib/python3.10/site-packages/xknx/knxip/knxip.py", line 115, in from_knx
    self.init(self.header.service_type_ident).from_knx(
  File "/usr/local/lib/python3.10/site-packages/xknx/knxip/tunnelling_request.py", line 59, in from_knx
    pos += self.cemi.from_knx(raw[pos:])
  File "/usr/local/lib/python3.10/site-packages/xknx/knxip/cemi_frame.py", line 127, in from_knx
    return self.from_knx_data_link_layer(raw)
  File "/usr/local/lib/python3.10/site-packages/xknx/knxip/cemi_frame.py", line 187, in from_knx_data_link_layer
    self.payload.from_knx(apdu)
  File "/usr/local/lib/python3.10/site-packages/xknx/telegram/apci.py", line 472, in from_knx
    channel, self.count, self.value = struct.unpack("!BBH", raw[1:])
struct.error: unpack requires a buffer of 4 bytes
```

kewde:
## Description

This PR adds two APCI's for read and responses for system network parameters.

Fixes #1021 (issue)

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -8,6 +8,10 @@ nav_order: 2
 
 # Unreleased changes
 
+### Protocol
+
+- Add A_SystemNetworkParameter_Read and A_SystemNetworkParameter_Response APCI service parsing.
+
 ### DPT
 
 - Add DPT 2 definitions
--- a/test/telegram_tests/apci_test.py
+++ b/test/telegram_tests/apci_test.py
@@ -39,6 +39,7 @@ from xknx.telegram.apci import (
     PropertyValueWrite,
     Restart,
     SystemNetworkParameterRead,
+    SystemNetworkParameterResponse,
     UserManufacturerInfoRead,
     UserManufacturerInfoResponse,
     UserMemoryRead,
@@ -313,7 +314,7 @@ class TestSystemNetworkParameterRead:
             object_type=0, property_id=0, operand=bytes.fromhex("00b001")
         )
 
-        assert payload.calculated_length() == 5
+        assert payload.calculated_length() == 6
 
     def test_from_knx(self) -> None:
         """
@@ -356,6 +357,68 @@ class TestSystemNetworkParameterRead:
         assert str(payload) == (
             '<SystemNetworkParameterRead object_type="0" property_id="0" '
             'operand="00b001" />'
+        )
+
+
+class TestSystemNetworkParameterResponse:
+    """Test class for SystemNetworkParameterResponse objects."""
+
+    def test_calculated_length(self) -> None:
+        """Test the test_calculated_length method."""
+        payload = SystemNetworkParameterResponse(
+            object_type=11,
+            property_id=5,
+            test_info_and_result=bytes.fromhex("0099aabbcc"),
+        )
+
+        assert payload.calculated_length() == 8
+
+    def test_from_knx(self) -> None:
+        """Test the from_knx method."""
+        payload = APCI.from_knx(bytes.fromhex("01c90b050099aabbcc"))
+
+        assert payload == SystemNetworkParameterResponse(
+            object_type=11,
+            property_id=5,
+            test_info_and_result=bytes.fromhex("0099aabbcc"),
+        )
+
+    def test_from_knx_does_not_strip_reserved_bits(self) -> None:
+        """
+        Test from_knx keeps the reserved 4 bits as-is in test_info_and_result.
+
+        Unlike SystemNetworkParameterRead, the reserved nibble here is not
+        masked out yet - see the NOTE in the class docstring.
+        """
+        payload = APCI.from_knx(bytes.fromhex("01c90b05f099aabbcc"))
+
+        assert payload == SystemNetworkParameterResponse(
+            object_type=11,
+            property_id=5,
+            test_info_and_result=bytes.fromhex("f099aabbcc"),
+        )
+
+    def test_to_knx(self) -> None:
+        """Test the to_knx method."""
+        payload = SystemNetworkParameterResponse(
+            object_type=11,
+            property_id=5,
+            test_info_and_result=bytes.fromhex("0099aabbcc"),
+        )
+
+        assert payload.to_knx() == bytes.fromhex("01c90b050099aabbcc")
+
+    def test_str(self) -> None:
+        """Test the __str__ method."""
+        payload = SystemNetworkParameterResponse(
+            object_type=11,
+            property_id=5,
+            test_info_and_result=bytes.fromhex("0099aabbcc"),
+        )
+
+        assert str(payload) == (
+            '<SystemNetworkParameterResponse object_type="11" property_id="5" '
+            'test_info_and_result="0099aabbcc" />'
         )
 
 
--- a/test/telegram_tests/apci_test.py
+++ b/test/telegram_tests/apci_test.py
@@ -340,6 +340,13 @@ class TestSystemNetworkParameterRead:
             object_type=0, property_id=0, operand=bytes.fromhex("00b001")
         )
 
+    def test_from_knx_wrong_length(self) -> None:
+        """Test from_knx raises ConversionError for a too-short APDU."""
+        # only 4 octets (TPCI/APCI, APCI, object_type, property_id) - the
+        # reserved/operand octet is missing.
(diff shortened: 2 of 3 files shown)


farmio:
I guess this might fix https://github.com/XKNX/xknx/issues/1021 !?

farmio (changes requested):
Review by Claude code 🤖:

Thanks for the PR! Unfortunately I think the PDU layout is not right, so this needs another round.

## The ASDU layout

Both new classes parse the ASDU as `object_type` = 1 octet, `property_id` = 1 octet, followed by a "reserved upper nibble". Per 03_03_07 §3.3.8 the layout is:

```
raw[0:2]                          APCI
raw[2:4]                          Interface Object Type (16 bit)
raw[4] + high nibble of raw[5]    PID (12 bit)
low nibble of raw[5]              reserved (4 bit)
raw[6:]                           test_info / test_result
```

Cross-checking with Calimero: it encodes `putShort(objectType).putShort(pid << 4).put(value)` and decodes with `iot = apdu[2]<<8 | apdu[3]`, `pid = apdu[4]<<4 | (apdu[5] & 0xf0)>>4`, `operand = apdu[6]` — see [`ManagementClientImpl.java`](https://github.com/calimero-project/calimero-core/blob/master/src/io/calimero/mgmt/ManagementClientImpl.java#L558-L616). So both the field widths and the position of the reserved nibble differ from this PR.

The real-world frame in the test docstring shows it nicely: `01c8 0000 00b0 01` decodes as **object type 0 (Device Object), PID 11 (`PID_SERIAL_NUMBER`), reserved 0, test_info `01`** — the standard "which devices are in programming mode" system broadcast. This PR decodes it as `object_type=0, property_id=0, operand=00b001`, so the PID is silently swallowed into the operand and the test passes with every field wrong.

Concretely:

1. `object_type` truncates at 255. It is a 16 bit field, and `struct.pack("!BB", ...)` would raise a bare `struct.error` rather than `ConversionError` for larger values.
2. `property_id` is 12 bit: `(raw[4] << 4) | (raw[5] >> 4)`.
3. The reserved nibble is `raw[5] & 0x0F`, not `raw[4] & 0xF0`. `test_from_knx_strips_reserved_bits` therefore checks the wrong nibble of the wrong octet — with the correct layout `01c8 0000 f0b0 01` is PID 15 with payload `b0 01`.
4. The minimum APDU length is 6, not 5 (4 octet ASDU). Note that exactly 6 is valid for the response: a device signals "unsupported object type / PID" by echoing `iot=0xFFFF` resp. `pid=0xFFF` with no test data, so please don't require test_info to be present.

## Other remarks

- **`from_knx` / `to_knx` are not symmetric.** `from_knx` masks the reserved bits away, but `to_knx` re-emits `operand` verbatim, so a frame with non-zero reserved bits doesn't reserialize to itself. Whichever way you go, both directions should agree — the other classes here (eg. `MemoryExtendedWriteResponse`) round-trip exactly.
- **The two classes intentionally disagree.** The `NOTE:` in `SystemNetworkParameterResponse` documents that it does *not* strip the reserved bits while `SystemNetworkParameterRead` does. The headers are identical, so please parse them identically; a small shared helper or common base class for the header would be nice.
- **Range validation is missing in `to_knx`.** Convention in this module is to raise `ConversionError` (see `MemoryExtendedWriteResponse.to_knx`); please add `0 <= object_type <= 0xFFFF` and `0 <= property_id <= 0xFFF`.
- `calculated_length` follows the `len(APDU) - 1` convention correctly, but it becomes `4 + len(test_info)` once the header is 4 octets — worth re-checking when reworking.
- Optional / follow-up: `A_SystemNetworkParameter_Write` (0x1CA) is the missing third of the set.

## Things that are fine

- Keeping test_info and test_result in one `test_info_and_result` field is a good call — Calimero needs the original request's operand length to split them, which a stateless parser doesn't have. The docstring explains this well.
- Dispatching 0x1C8/0x1C9 inside the `ADC_RESPONSE` branch makes `ADCResponse` unparseable for channels 8 and 9, but that collision comes from the spec and the existing `MEMORY_EXTENDED_*` cases set the precedent.
- Changelog entry and test layout follow the project style.

kewde:
Welp, I botched the PID lol. I've fixed that, should be aligned, I was looking too much at the variable length operands that I missed to check correctness on the static length bits. I'm kinda stumped there's no length prefix in the encoding on these requests; do you know if calimero has some tricks up its sleeve to split the variable length operands vs test result?

I think that indeed fixes that issue, you can reproduce it by opening up ETS and going to the connections tab. That will emit search requests, which when responded to by a router (potentially a virtual router) will trigger a system network parameter read for the serial number.

farmio:
Claude Code  🤖:

No trick, unfortunately - Calimero relies on out-of-band knowledge on both sides, because there really is no length indicator on the wire.

**Responder side** ([`ManagementServiceNotifier.onNetworkParamRead`](https://github.com/calimero-project/calimero-device/blob/master/src/main/java/io/calimero/device/ManagementServiceNotifier.java#L391)): it takes everything after the 4 octet header as `info`, passes it to the application, and then just concatenates:

```java
asdu.putShort((short) objectType);
asdu.putShort((short) (pid << 4));
asdu.put(info).put(res);   // test_info echoed verbatim, then test_result
```

So the boundary in the response is exactly "length of the test_info from the request" - by construction, not by any marker.

**Client side** ([`ManagementClientImpl.readSystemNetworkParameter`](https://github.com/calimero-project/calimero-core/blob/master/src/io/calimero/mgmt/ManagementClientImpl.java#L558)): it knows what it sent, so it just slices it off:

```java
final int prefix = 2 + 4 + 1 + additionalTestInfo.length;  // APCI + header + operand + extra test info
return responders.stream().map(r -> Arrays.copyOfRange(r, prefix, r.length)).collect(toList());
```

It additionally sanity-checks the echo (`apdu[6] == operand`, plus IOT and PID) and treats a 6 octet response as "unsupported object type / PID" rather than a result.

Spec-wise this is by design: the operand is the first octet of test_info and selects which test is being requested, and the format and length of the rest of test_info are fixed per interface object type / PID / operand combination. So a stateless decoder genuinely cannot split the two - it isn't just inconvenient, the information isn't in the frame. Keeping them together as `test_info_and_result` in the APCI layer is the right call. If we ever add a management client procedure on top (which sends the request and correlates the response), that layer knows the test_info length and can do the split there.

One small related data point: Calimero bounds the ASDU to 4..12 octets, and for a *read* it requires at least one test_info octet (the operand), so it would never emit a 6 octet `A_SystemNetworkParameter_Read`. Our parser accepting it is harmless leniency, no need to change.

And yes, I think you're right that this covers #1021 - thanks for digging into it! I verified the new encoder produces byte-identical frames to Calimero's `putShort(objectType).putShort(pid << 4)` across the whole PID range, so the layout looks correct now.

farmio (approved):

farmio:
Looks good to me, thanks! 👍
I personally don't know Calimero code, so can't really help you there... been a couple of years since I have looked there 🙃 but I think it should still be a good source!

PR #1859, opened 2026-07-22
feat: add  `A_FunctionPropertyExt*`

kewde:
## Description

This PR adds a bunch of the `A_FunctionPropertyExt*` as defined in the specification.
Had to move the APCI return code to prevent circular imports, added the missing two too.

Fixes # (issue)

## Type of change

## Checklist


farmio (changes requested):
Claude Code  🤖:

Really nice PR - the wire format is spot on. I checked `_pack_/_unpack_function_property_ext_header` against Calimero's `functionProperty` encoder and its `extractFunctionPropertyExtData` / `onFunctionPropertyExtCommandOrState` decoders and the 16 bit IOT + 12 bit object-instance + 12 bit PID packing matches byte-for-byte; I round-tripped the full field ranges (`0xFFFF,0xFFF,0xFFF` and `0x1234,0xABC,0xDEF`) to be sure. Treating the reserved/service-id/input tail as opaque `data` is the right level of abstraction (Calimero does the same at the ASDU layer). Dispatch placement is correct (`0x1D4..0x1D6 & 0x03C0 == 0x1C0`), and the `ReturnCode` move leaves no dangling references. `pytest`/`ruff`/`mypy` all clean.

One thing I'd like fixed before merging, plus a couple of minor notes.

## Should fix: bare `ValueError` on unknown return codes

`FunctionPropertyExtStateResponse.from_knx` does `return_code=ReturnCode(raw[7])`. Any return code not in the enum raises an unhandled `ValueError` that escapes `APCI.from_knx`, instead of the `ConversionError` the parser contract uses everywhere else. The spec reserves ranges (`01-1F`, `20-5F`, `A0-DF`, "None proposed") that a device could still put on the wire, and I confirmed a response with `return_code=0x01` crashes with `1 is not a valid ReturnCode`. Calimero avoids this with `ReturnCode.of(code)`.

The same pattern already exists at `xknx/knxip/tunnelling_feature.py:191` (`ReturnCode(raw[5])`), so at least it's consistent - but this is a good moment to wrap both (`try/except ValueError -> ConversionError`, or preserve the raw byte).

## Minor

- **Test parity with the SystemNetworkParameter classes.** The `_pack_function_property_ext_header` range-validation branches (IOT/OI/PID out of range) are untested, and there's no round-trip test - whereas the `SystemNetworkParameter*` classes added both `test_to_knx_*_out_of_range` and `test_round_trip`. Codecov will likely flag the three uncovered lines. Cheap to add for parity.
- `__str__` renders `return_code="ReturnCode.E_SUCCESS"` via the enum repr; `self.return_code.name` -> `"E_SUCCESS"` would match the int/hex style used by the other `__str__` methods. Cosmetic.
- The new length-error messages (`f"... : {len(raw)} bytes."`) differ in format from the `SystemNetworkParameter` ones (`"... in CEMI: {raw.hex()}"`). Trivial.

## Nice touches

- Adding the missing `E_COMMAND_INVALID` (0xF2) and `E_COMMAND_IMPOSSIBLE` (0xF3) while moving the enum - good catch.
- Breaking change documented under its own heading. If you'd like to soften it, a re-export shim in the old `application_layer_enum` module (importing from `apci`) wouldn't reintroduce the circular import - but documenting it is defensible too.

kewde:
Thanks @farmio - I've addressed the feedback (minus the breaking change in nice to have, le me know if you want that).

farmio:
Thanks!

PR #1860, opened 2026-07-22
fix: A_Restart Master Reset serialization

kewde:
## Description

Add RestartMasterReset (APCI 0x381) for the request and RestartMasterResetResponse (APCI 0x3A1, error_code + process_time) for its confirmation - Basic Restart has no application-layer response per spec, so that case is intentionally not modeled.

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -16,6 +16,7 @@ nav_order: 2
 
 - Add A_SystemNetworkParameter_Read, A_SystemNetworkParameter_Response and A_SystemNetworkParameter_Write APCI service parsing.
 - Add A_FunctionPropertyExtCommand, A_FunctionPropertyExtState_Read and A_FunctionPropertyExtState_Response APCI service parsing.
+- Fix A_Restart parsing to distinguish Basic Restart from Master Reset (restart_type bit was ignored, silently dropping erase_code/channel_number on relay). Add A_Restart_Response parsing for the Master Reset confirmation.
 
 ### DPT
 
--- a/test/telegram_tests/apci_test.py
+++ b/test/telegram_tests/apci_test.py
@@ -41,6 +41,8 @@ from xknx.telegram.apci import (
     PropertyValueResponse,
     PropertyValueWrite,
     Restart,
+    RestartMasterReset,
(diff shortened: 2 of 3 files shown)


farmio (commented):
Claude Code  🤖:

Looks correct - nice work on a fiddly one. The subtlety here is that the restart-type bit folds into the APCI octet, and this gets it right.

I cross-checked against Calimero: it builds the master-reset request as a length-optimized APDU of `RESTART` (0x380) with SDU `{0x01, eraseCode, channel}`, so the `0x01` restart-type byte folds into the APCI low 6 bits (`0x80 | 0x01 = 0x81`), leaving `03 81 EC CH` - exactly what this PR emits. The response `03 A1 ER TH TL` (error at raw[2], 2-byte process_time at raw[3:5]) matches Calimero's decode too. I ran both directions across value ranges and they're byte-identical. Dispatch is correct: `0x381`/`0x3A1` mask to the `RESTART` service and are checked before falling through, so basic `03 80` still routes to `Restart`. The APCI bit-layout docstring is accurate and explains the 0x3A1 response value neatly. `pytest`/`ruff`/`mypy` all clean.

A few optional notes, none blocking:

- **Range validation in `to_knx`**, for consistency with the recent SystemNetworkParameter / FunctionPropertyExt classes: `process_time=0x10000` (and out-of-range `erase_code`/`channel_number`/`error_code`) raises a bare `struct.error` rather than `ConversionError`. Those newer classes validate and raise `ConversionError`; these don't. Lower severity since it's the to_knx path, but worth it for family consistency.
- **Length-error message format** differs from the `"... in CEMI: {raw.hex()}"` form the FunctionPropertyExt PR converged on (`f"Invalid length for RestartMasterReset: {len(raw)} bytes."`). Trivial, just noting it if you want the family uniform.
- **Optional robustness:** `from_knx` requires the response low byte to be exactly `0xA1`; Calimero uses a looser `apdu[1] & 0x32` "response bit set" check. Requiring the canonical value is spec-correct and I wouldn't change it - just flagging the difference.

farmio (approved):

PR #1862, opened 2026-07-23
feat: add A_PropertyExtValue*

kewde:
## Description

This PR implements the remainder of the ADC service, all missing A_PropertyExtValue*

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,7 +17,7 @@ nav_order: 2
 - Add A_SystemNetworkParameter_Read, A_SystemNetworkParameter_Response and A_SystemNetworkParameter_Write APCI service parsing.
 - Add A_FunctionPropertyExtCommand, A_FunctionPropertyExtState_Read and A_FunctionPropertyExtState_Response APCI service parsing.
 - Fix A_Restart parsing to distinguish Basic Restart from Master Reset (restart_type bit was ignored, silently dropping erase_code/channel_number on relay). Add A_Restart_Response parsing for the Master Reset confirmation.
-- Add A_PropertyExtValue_Read and A_PropertyExtValue_Response APCI service parsing. Recognize A_PropertyExtValue_WriteCon/WriteConRes/WriteUnCon/InfoReport and A_PropertyExtDescription_Read/Response APCI services; payload parsing not implemented yet for those (raises `NotImplementedError`).
+- Add A_PropertyExtValue_Read, A_PropertyExtValue_Response, A_PropertyExtValue_WriteCon and A_PropertyExtValue_WriteConRes APCI service parsing. Recognize A_PropertyExtValue_WriteUnCon/InfoReport and A_PropertyExtDescription_Read/Response APCI services; payload parsing not implemented yet for those (raises `NotImplementedError`).
 
 ### DPT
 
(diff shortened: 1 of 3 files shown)


farmio (changes requested):
Claude Code  🤖:

Nice, big chunk of the property-ext surface. The field layouts check out against the spec, but there's one blocker that affects all 8 services, plus a couple of small notes.

## Blocker: `calculated_length` is off by one in all 8 classes

Every new class returns `len(APDU) - 2` where it should be `len(APDU) - 1`. `calculated_length()` becomes the NPDU length octet in `CEMILData.to_knx()` (`cemi_frame.py:187`), and `from_knx` enforces `len(apdu) == npdu_len + 1` (`cemi_frame.py:218-222`). So a round-trip through `CEMILData` fails - xknx can't even parse its own serialized frame:

```python
from xknx.cemi import CEMILData
from xknx.cemi.cemi_frame import ...  # L_Data
from xknx.telegram.apci import PropertyExtValueRead
from xknx.telegram.address import IndividualAddress
from xknx.telegram.tpci import TDataIndividual

f = CEMILData(flags=0, src_addr=IndividualAddress("1.1.1"),
              dst_addr=IndividualAddress("1.1.2"),
              payload=PropertyExtValueRead(17, 1, 51), tpci=TDataIndividual())
CEMILData.from_knx(bytes(f.to_knx()))
# CouldNotParseCEMI: "APDU LEN should be 8 but is 9 ..."
```

A real device would likewise see a length field one octet short. All the existing spec-verified classes (`SystemNetworkParameter*`, `FunctionPropertyExt*`, etc.) satisfy `calculated_length == len(to_knx()) - 1`; these 8 are the only ones that don't. Correct values:

| class | current | correct |
|---|---|---|
| PropertyExtValueRead | 8 | 9 |
| Response / WriteCon / WriteUnCon / InfoReport | 8 + len(data) | 9 + len(data) |
| PropertyExtValueWriteConRes | 9 | 10 |
| PropertyExtDescriptionRead | 7 | 8 |
| PropertyExtDescriptionResponse | 15 | 16 |

The tests miss this because `test_calculated_length` asserts the buggy values and nothing round-trips through `CEMILData`. Please fix the eight values + their assertions, and consider adding a `CEMILData` round-trip test (or simply `assert obj.calculated_length() == len(obj.to_knx()) - 1`) so it can't regress.

## Verified correct against 03_03_07

- Header 16-bit IOT + 12-bit object_instance + 12-bit PID (reuses the already-verified `_unpack_function_property_ext_header`).
- Value services (Fig 49/50, §3.4.5.1): `nr_of_elem` 8-bit + `start_index` 16-bit, data at raw[10:]. Correct.
- Description Read (Fig 44, §3.4.3.2): description_type 4-bit + property_index 12-bit. Correct.
- Description Response (Fig 45): DPT main 2B + sub 2B, then writable(bit7)/reserved(bit6)/PDT(6-bit), max_nr_of_elem 2B, read_level/write_level (4+4). Matches the figure, and the writable-at-bit-7 split matches the classic Fig 43 convention. Heads-up: Calimero stubs these (`useExtPropertyServices = false`), so the spec figure is the only cross-reference for the PDT byte - probably worth a second look there, but I believe it's right.

## Minor

- Inconsistent `start_index` defaults: `PropertyExtValueRead` and `WriteConRes` default `start_index=1`, while Response/WriteCon/WriteUnCon/InfoReport default `0`. Per spec, index 0 = "read the current number of elements" and normal reads start at 1 - worth picking one and applying it consistently across the siblings.
- The good habits from the earlier PRs are all here (range validation -> `ConversionError`, `ReturnCode` `ValueError`->`ConversionError` wrap in `WriteConRes`, `__str__` uses `.name`, `"... in CEMI: {raw.hex()}"` messages). Nice.

PR #1867, opened 2026-07-23
feat: implement parsing for all APCI's / APDU's 

kewde:
## Description

I've ran down the application layer spec, added all the APCIs and added parsing for all the packet layouts. It's quite a big PR, I'm ok with splitting it up if you want.

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,6 +27,7 @@ nav_order: 2
 - Add A_Link_Read, A_Link_Response and A_Link_Write APCI service parsing.
 - Add A_GroupPropValue_Read, A_GroupPropValue_Response, A_GroupPropValue_Write and A_GroupPropValue_InfoReport APCI service parsing (KNX Logical Tag Extended specification).
 - Add A_FileStream_InfoReport APCI service parsing.
+- Fix A_DomainAddressSerialNumber_Read parsing to raise ConversionError instead of a bare struct.error for a truncated frame.
 
 ### DPT
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -18,6 +18,15 @@ nav_order: 2
 - Add A_FunctionPropertyExtCommand, A_FunctionPropertyExtState_Read and A_FunctionPropertyExtState_Response APCI service parsing.
 - Fix A_Restart parsing to distinguish Basic Restart from Master Reset (restart_type bit was ignored, silently dropping erase_code/channel_number on relay). Add A_Restart_Response parsing for the Master Reset confirmation.
 - Add A_PropertyExtValue_Read, A_PropertyExtValue_Response, A_PropertyExtValue_WriteCon, A_PropertyExtValue_WriteConRes, A_PropertyExtValue_WriteUnCon, A_PropertyExtValue_InfoReport, A_PropertyExtDescription_Read and A_PropertyExtDescription_Response APCI service parsing.
+- Add A_UserMemoryBit_Write and A_MemoryBit_Write APCI service parsing.
+- Add A_FilterTable_Open, A_FilterTable_Read, A_FilterTable_Response and A_FilterTable_Write APCI service parsing.
+- Add A_RouterMemory_Read, A_RouterMemory_Response and A_RouterMemory_Write APCI service parsing. Recognize A_RouterStatus_Read/Response/Write APCI services; these are a legacy EIB/BCU1-era coupler status service with no PDU definition in the current Application Layer spec and are not planned for implementation - configure couplers via the Router Object properties (`A_PropertyValue_Read`/`A_PropertyValue_Write`) instead.
+- Add A_Key_Write and A_Key_Response APCI service parsing.
+- Add A_NetworkParameter_Read, A_NetworkParameter_Response and A_NetworkParameter_Write APCI service parsing.
+- Add A_DomainAddress_Write, A_DomainAddress_Read, A_DomainAddress_Response, A_DomainAddressSelective_Read, A_DomainAddressSerialNumber_Read, A_DomainAddressSerialNumber_Response and A_DomainAddressSerialNumber_Write APCI service parsing.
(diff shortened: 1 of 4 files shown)


farmio (changes requested):
Claude Code  🤖:

Huge effort, and mechanically it holds up really well - but there's one receive-path regression to fix, and I'd like to talk about splitting this before it goes in.

## What I verified mechanically (all clean)

- **`calculated_length` invariant holds everywhere.** I harvested all 105 wire frames the tests feed to `APCI.from_knx` and every parseable one satisfies `calculated_length() == len(to_knx()) - 1` - no repeat of the #1862 off-by-one. The only non-identical round-trips are intentional reserved-bit normalizations (SystemNetworkParameter + the new LinkRead).
- **CEMI round-trips: 64/64 clean** through `CEMILData.to_knx -> from_knx`.
- **Existing classes are pure relocations.** The apci.py deletions are Authorize / PropertyValue / PropertyDescription / IndividualAddressSerial moved into spec-table order - identical text, each class present exactly once, no semantic change. The cemi test change is correct too (moved the "APDU not supported" probe off `0x3C0`, now FilterTable_Open, to the reserved gap `0x3C4`).
- **Spot-checked layouts against 03_03_07:** NetworkParameter correctly uses an 8-bit PID (vs SystemNetworkParameter's 12-bit, §3.2.6); DomainAddressWrite validates 2/6-octet PL110/RF addresses; Link Read/Response reserved-nibble + bounds handling look right.

## Blocker: RouterStatus stubs raise `NotImplementedError` into the parse path

The three `A_RouterStatus_*` classes are dispatched in `from_knx`, but their `from_knx` raises `NotImplementedError`. `CEMILData.from_knx` only catches `ConversionError` (`cemi_frame.py:257-258`), so a real (legacy) RouterStatus frame now escapes as an **uncaught exception** instead of the pre-PR clean `CouldNotParseCEMI`:

```python
raw = bytes.fromhex("0000" "1101" "1102" "01" "03cd")  # L_Data carrying A_RouterStatus_Read
CEMILData.from_knx(raw)
# before this PR: CouldNotParseCEMI   ->   now: uncaught NotImplementedError
```

That can take down a receive loop when a legacy coupler is on the bus - strictly worse than leaving the code unrecognized. The docstring says the stub exists "so dispatch doesn't misroute it," but it misroutes to a raise type the parse path doesn't handle. Easy fix: raise `ConversionError` in the stubs (keep the explanatory message), or drop the dispatch entries and let them fall through to the existing `raise ConversionError` at the end of `from_knx`. I'd also add a test asserting a `0x3CD` APDU parses to `CouldNotParseCEMI` at the CEMI layer.

## Please split it

4481 lines / ~20 services in one PR. The invariant sweeps above give good coverage of lengths and round-trips, but they can't catch a self-consistent-yet-spec-wrong layout - and realistically neither a reviewer nor I can independently cross-check 20 PDUs against 03_03_07 in a single pass with real confidence. You already offered to split; I'd take that offer. Natural seams: FilterTable, RouterMemory, NetworkParameter, DomainAddress (+SerialNumber), Link, GroupPropValue, FileStream, the bit-write services. That also lets the RouterStatus decision stand on its own.

Thanks again - the per-class quality (range validation, `ReturnCode` ValueError->ConversionError wrapping, consistent `__str__` / error messages) is clearly carried over from the earlier PRs.

farmio:
Claude Code  🤖:

I did a full class-by-class pass over every service in this PR, checking each byte layout against the 03_03_07 figures (and the LTE spec for GroupPropValue). The RouterStatus fix looks good. Layouts are spec-accurate throughout - NetworkParameter (8-bit PID), the DomainAddressSerialNumber_Write 21-octet secure split (mcast 4 + security version 1 + backbone key 16), Link flags (d=bit1/s=bit0), FileStream (handle+seq packed in one octet), FilterTable/RouterMemory (number+addr+data), the MemoryBit AND/XOR semantics, etc. `calculated_length == len(to_knx()) - 1` holds for all 105 test frames and CEMI round-trips are 64/64 clean. GroupPropValue is correctly attributed to LTE §7.6.4 (not 03_03_07) and its IOT2/OI1/PID1 header matches the cEMI LTE convention.

One new issue slipped through, same class as the RouterStatus one you just fixed:

## `DomainAddressSerialNumberRead` (0x3EC) leaks a bare `struct.error`

Unlike its siblings, `from_knx` has no length guard before `struct.unpack("!6s", raw[2:])`. A truncated frame raises `struct.error`, which `CEMILData.from_knx` doesn't catch (it only catches `ConversionError`), so it propagates **uncaught** out of the receive path:

```python
CEMILData.from_knx(bytes.fromhex("0000 1101 1102 03 03ec0102"))
# UNCAUGHT struct.error: unpack requires a buffer of 6 bytes
```

One-line fix, matching the other DomainAddressSerialNumber_* classes:

```python
if len(raw) != 8:
    raise ConversionError(f"Invalid length for A_DomainAddressSerialNumber_Read in CEMI: {raw.hex()}")
```

...and a truncated-frame test asserting it parses to `CouldNotParseCEMI`/`UnsupportedCEMIMessage` at the CEMI layer.

## Not blocking, just noting

Fuzzing all services turned up ~29 pre-existing classes with the same bare-`struct.error`-on-truncation behaviour (ADC, Memory*, PropertyValue*, Authorize, IndividualAddressSerial*, DeviceDescriptorResponse, ...). That's a longstanding trait, not something this PR introduced - worth a separate hardening pass sometime, out of scope here. Two cosmetic nits in the new code: `DomainAddressSelectiveRead` accepts an empty ASDU (no min-length check) and `DomainAddressRead` silently ignores trailing bytes.

Nice work overall - the per-PDU accuracy across ~28 services is impressive.

kewde:
@farmio fixed 👍

farmio (approved):

farmio:
Thank you very much 👍

PR #1869, opened 2026-07-24
Recognize A_RouterStatus_Read/Response/Write APCI services

kewde:
## Summary

- Adds `ROUTER_STATUS_READ`/`RESPONSE`/`WRITE` (APCI 0x3CD-0x3CF) to `APCIExtendedService` and wires them into `APCI.from_knx`'s dispatcher.
- Deliberately leaves them unimplemented: this is a legacy EIB/BCU1-era line coupler status register (BCU alias "LcGroupWrite" for the write) with no PDU definition in the current Application Layer spec (03_03_07 only reserves the APCI codes). Modern ETS configures couplers exclusively through the Router Object (Interface Object Type 6) properties via `A_PropertyValue_Read`/`Write` instead (see AN161 Coupler Model 2.0).
- `from_knx` raises `ConversionError` so a real legacy frame on the bus is rejected as `UnsupportedCEMIMessage`, matching how every other unrecognized service is handled - not an uncaught crash. `calculated_length`/`to_knx` raise `NotImplementedError` since there's no legitimate way to construct one of these objects to send.

This is the first of a stack of small PRs splitting up a previous oversized branch that tried to add ~20 APCI services at once. This one stands alone since it doesn't add any new dispatch entries the other PRs would conflict with.

## Test plan

- [x] Added `TestRouterStatusRead/Response/Write` covering dispatch (`ConversionError`), direct construction (`NotImplementedError` for `to_knx`/`calculated_length`), and `__str__`.
- [x] Added 3 cases to the existing `test_unsupported_tpci_apci` CEMI-layer parametrization confirming a full L_Data frame using these APCI codes parses to `UnsupportedCEMIMessage`, not a crash.
- [x] Full test suite (3427 passed), ruff, mypy clean.


farmio:
done by #1867

PR #1870, opened 2026-07-24
Implement A_UserMemoryBit_Write and A_MemoryBit_Write APCI parsing

kewde:
## Summary

- Implements `A_UserMemoryBit_Write` (§3.5.6.4, APCI 0x2C4) and `A_MemoryBit_Write` (§3.5.5, APCI 0x3D0).
- Both modify 1-40 bits in a contiguous block of up to 5 octets: a 1 byte `number` (octet count), a 2 byte address, and `and_data`/`xor_data` (each `number` bytes). Each result bit is computed as `(and_data_bit AND block_bit) XOR xor_data_bit` - the two share an identical wire layout, differing only in which memory space they target (user memory vs. general memory).

Second in the stack splitting up the oversized APCI branch - based on #1869 (`kewde/apci-routerstatus`).

## Test plan

- [x] `TestUserMemoryBitWrite`/`TestMemoryBitWrite`: calculated_length invariant, from_knx (valid + too-short + mismatched-number-length), to_knx, round-trip, range validation (address/memory_address out of range, `and_data` >255 octets, `and_data`/`xor_data` length mismatch), `__str__`.
- [x] Full test suite (3446 passed), ruff, mypy clean.


farmio:
done by #1867

PR #1876, opened 2026-07-26
Add host-agnostic MCP tool functions (xknx.mcp)

martinhoefling:
## Summary

Adds an `xknx.mcp` subpackage of plain **async** tool functions for the KNX bus and data point types, so consumers (SpectrumKNX, Home Assistant, …) can expose bus and DPT operations over the Model Context Protocol **without pulling an MCP SDK into this library**. Each consumer wraps these functions into its own MCP transport and decides which write tools to expose.

Part of a wider KNX MCP effort spanning the XKNX libraries; tracking issue: martinhoefling/SpectrumKNX#331 (epic martinhoefling/SpectrumKNX#328). Companion PRs add the same `mcp` subpackage to `xknxproject` (XKNX/xknxproject#625) and `knx-telegram-store`.

## Tools (transport-agnostic, JSON-serialisable dataclass I/O)

DPT catalogue (static, no connection):
- `list_dpts` — browse DPTs with a main-number / text filter and pagination
- `describe_dpt` — resolve a DPT number (`"9.001"`) or value-type name to its number, value type, unit and numeric bounds

Bus (require a connected `XKNX`):
- `get_connection_status` — connection state, type, local individual address
- `read_group_value` — read and DPT-decode a group value
- `send_group_value_read` — queue a `GroupValueRead`
- `send_group_value_write` — queue a `GroupValueWrite` (a **write** op; consumers gate it behind their read-write mode)

## Notes

- Thin wrappers over `xknx.tools`; frozen dataclass inputs/outputs, no MCP SDK / Home Assistant / web-framework imports.
- `async` throughout for a uniform calling convention across the XKNX MCP tool libraries.
- Covered by tests (`pytest-asyncio`; `ValueReader.read` mocked for reads). Passes `ruff` and strict `mypy`.


farmio (commented):
Claude Code  🤖:

Reviewed with a focus on LLM usefulness, readability, and Home Assistant MCP integration. Cross-checked the wrappers against `xknx.tools`, `DPTBase`, and `connection_manager`.

## Overall

Genuinely well-crafted. The core decision — a base-library subpackage of plain `async` functions with frozen, JSON-native dataclass I/O and **zero** MCP-SDK / HA / web-framework dependency — is the right call and is executed cleanly. The wrappers use the existing `xknx.tools` primitives correctly, the enum/attribute accesses line up (`XknxConnectionType.NOT_CONNECTED.value is None`, so the `connection_type` test is right), and coverage hits the meaningful paths. Close to mergeable; the notes below are refinements, not blockers.

## Strengths for LLM use

- **Docstrings read like tool descriptions** — one clear imperative line per function, field semantics spelled out (`"9.001"` vs `"temperature"`, the raw-payload fallback, the write-gating note). Exactly what an MCP layer wants to surface.
- `dataclasses.asdict` + JSON-native `GroupValue` means outputs serialize with no custom encoder. `_jsonify` correctly flattens tuples→lists and uses `is not None` for `responded` (so a decoded `0`/`False` still counts as a response).
- Write safety is delegated to the consumer, the correct boundary for a library.

## Issues, ranked

**1. Per-field descriptions are lost (highest impact for LLM quality).** Frozen dataclasses give a schema generator *types* but not per-parameter `description`s. The rich prose lives in class docstrings, which most generators won't map onto individual JSON-Schema properties — so the tool schema for `send_group_value_write` shows `value_type: string | null` with no hint that it's `"9.001"`-or-`"temperature"`. Given the deliberate no-Pydantic stance, consider `typing.Annotated[str | None, "DPT number or value-type name"]` on the fields, or ship a documented mapping recipe for consumers. Biggest lever on downstream tool quality.

**2. Inconsistent error contract.** `describe_dpt` returns structured `found=False`, but `read_group_value` / `send_group_value_*` *raise* on an unparseable address (and on encoding errors). LLMs handle a uniform contract far better. Either return a structured error for bad addresses too, or explicitly document "these raise `ValueError`/`CouldNotParseAddress`; wrap as a tool error."

**3. Pagination isn't self-describing.** `DptListResult` reports `total_count` and `limit_reached` but not the `offset` used or a `next_offset`, so a paging LLM must track offset itself. Echoing `offset` / `next_offset` makes it stateless for the caller.

**4. `SendResult.queued` is a constant.** It defaults `True` and is never set otherwise, and the send is fire-and-forget (queued ≠ delivered ≠ acknowledged). Worth stating in the tool description so the model doesn't report "written" as confirmed — and either give `queued` real meaning or drop it.

## Readability / code

- **Redundant address validation** (`tools.py`): `send_group_value_read`/`_write` call `parse_device_group_address(...)` and then the underlying `group_value_*` parse again. The primitive already parses before `put_nowait`, so nothing is queued on a bad address regardless — the `# validate before queueing` comment describes a guarantee you already have.
- **Dead branches in `_jsonify`**: `read_group_value` already unwraps to the raw `.value.value`, so the `DPTArray`/`DPTBinary` `isinstance` branches never execute on the read path. Remove them, or note that `_jsonify` is intentionally general.
- **Input-type naming drift**: `ReadGroupValueInput` vs `GroupValueWriteInput` vs `GroupAddressInput`. Align to one pattern (e.g. `GroupValueReadInput`).
- **Subpackage name**: `xknx.mcp` contains no MCP code. The docstring clarifies intent, but since it lands in `__all__` it's a new stable public surface — worth a deliberate decision (something like `agent_tools` describes it more honestly). Low priority / debatable.
- **Test nit**: `pytest.raises(Exception)` with `# noqa: B017` is too broad; asserting `CouldNotParseAddress` would catch a regression that changes the failure mode.

## HA MCP-server integration

Integration mechanics are clean: the bus tools take an injected `XKNX`, and HA already holds one per KNX config entry, so wiring is straightforward. Two flags:

- **Abstraction level**: these operate on raw group addresses + DPTs, *below* HA's entity/device model — powerful for a KNX-literate LLM but bypassing HA's entity permissions and exposure settings. The library correctly punts write-gating to the consumer; HA's side of that gate is where the real safety story lives.
- **Schema mapping**: HA doesn't use Pydantic for its LLM tools either, so issue #1 lands there too — it'll need to re-author parameter descriptions unless the library carries them.

## Project hygiene

- **No changelog entry** — `docs/changelog.md` should get a line for this new public subpackage.
- No docs page for the new public API entering `__all__`; consider whether it's documented/versioned as stable.

martinhoefling:
Thanks for the thorough review — pushed changes addressing it (commit `f9e4801`; the sibling `xknxproject` PR XKNX/xknxproject#625 got the parallel treatment for consistency):

**Addressed**
1. **Per-field descriptions** — input dataclass fields now carry their description as `typing.Annotated[T, "…"]` metadata (plain string, no Pydantic). A consumer can surface these via `get_type_hints(..., include_extras=True)`; documented the convention in the module docstring.
3. **Self-describing pagination** — `DptListResult` now reports the `offset` used and a `next_offset` (`None` when exhausted). Mirrored to the three xknxproject list results.
4. **`SendResult.queued`** — clarified in the docstring that it means "enqueued for transmission", not delivered/acknowledged.
- **Redundant address validation** — dropped; the `xknx.tools` primitives already parse before `put_nowait`, so a bad address raises and nothing is queued. Documented the raised `CouldNotParseAddress`/`ConversionError` instead.
- **Dead `_jsonify` branches** — already removed in an earlier commit (`read_group_value` unwraps to `.value.value`).
- **Naming drift** — renamed `ReadGroupValueInput`/`Result` → `GroupValueReadInput`/`Result` to match `GroupValueWrite*`.
- **Test nit** — the invalid-address test now asserts `CouldNotParseAddress`.
- **Changelog** — added an entry under a new *Unreleased* section.

**Deferred / kept as-is (happy to change)**
2. **Uniform error contract** — kept raising for invalid addresses/encoding (documented) rather than returning a structured error, since raising is the natural library boundary and the consumer wraps it as a tool error; `describe_dpt`'s `found=False` is a lookup miss rather than bad input. Can switch to structured errors if you'd prefer consistency.
- **Subpackage name `mcp`** — kept for consistency with the already-merged `knx_telegram_store.mcp` and the cross-library naming; open to `agent_tools` if the org prefers.
- **Per-field descriptions in the HA/consumer schema** — that's consumer-side; the `Annotated` metadata is there for them to read.

xknx/mcp/tools.py
```diff
@@ -0,0 +1,228 @@
...
+    transcoder = DPTBase.get_dpt(request.value_type)
+    encoded = transcoder.to_knx(request.value)
+    if isinstance(encoded, DPTArray):
+        payload = list(encoded.value)
+    else:
+        payload = [encoded.value]
+    return EncodeValueResult(payload=payload, value_type=request.value_type)
+
+
+async def decode_payload(request: DecodePayloadInput) -> DecodePayloadResult:
```
kewde: Maybe decode_dpt_payload?

xknx/mcp/tools.py
```diff
@@ -0,0 +1,228 @@
...
+    :exc:`~xknx.exceptions.ConversionError` if ``value`` does not fit ``value_type``
+    (nothing is queued in either case).
+    """
+    group_value_write(
+        xknx, request.group_address, request.value, value_type=request.value_type
+    )
+    return SendResult(group_address=request.group_address, apci="GroupValueWrite")
+
+
+async def encode_value(request: EncodeValueInput) -> EncodeValueResult:
```
kewde: Maybe encode_dpt_payload?

xknx/mcp/tools.py
```diff
@@ -0,0 +1,228 @@
...
+    Queue a GroupValueRead telegram to trigger a response on the bus.
+
+    Raises :exc:`~xknx.exceptions.CouldNotParseAddress` for an invalid address
+    (nothing is queued in that case).
+    """
+    group_value_read(xknx, request.group_address)
+    return SendResult(group_address=request.group_address, apci="GroupValueRead")
+
+
+async def send_group_value_write(xknx: XKNX, request: GroupValueWriteInput) -> SendResult:
```
kewde, line 180: nit maybe also an option to send a raw cemi?

xknx/mcp/tools.py
martinhoefling: Done — renamed to `encode_dpt_payload` (and its dataclasses `EncodeDptPayloadInput`/`Result`) in 0fbc181.

xknx/mcp/tools.py
martinhoefling: Done — renamed to `decode_dpt_payload` (and `DecodeDptPayloadInput`/`Result`) in 0fbc181.

xknx/mcp/tools.py
martinhoefling: Good idea. I'd like to keep this PR to DPT-typed group-value I/O and add a raw-CEMI send as a follow-up: it's a broader surface (arbitrary APCI/destination, not just GroupValueWrite) with a bigger safety footprint, so it belongs behind the consumer's read-write gate with its own tests. Happy to open a tracking issue if you'd like it in scope here.

kewde (approved):
utACK 👍

xknx/mcp/tools.py
```diff
@@ -0,0 +1,216 @@
...
+
+    ``read_group_value`` returns Python natives (a DPT transcoder's ``from_knx``
+    output, or the raw payload value). Tuples become lists; anything not already
+    JSON-native (e.g. an enum or dataclass from a complex DPT) is stringified.
+    """
+    if isinstance(value, tuple):
+        return [_jsonify(item) for item in value]
+    if value is None or isinstance(value, bool | int | float | str | list | dict):
+        return value
+    return str(value)
```
farmio, line 151: For `DPTComplex` we should probably use `as_dict()` instead of `str()` when we want json-native https://github.com/XKNX/xknx/blob/c99aa412899503d0c3faaec3860a59ade2b5b389/xknx/dpt/dpt.py#L407
Not sure about DPTEnum - we could maybe just use `value.name` here.

xknx/mcp/tools.py
```diff
@@ -0,0 +1,216 @@
...
+def _summarize_dpt(dpt: type[DPTBase]) -> DptSummary:
+    value_min, value_max, resolution = _numeric_bounds(dpt)
+    return DptSummary(
+        dpt=dpt.dpt_number_str(),
+        value_type=dpt.value_type,
+        unit=dpt.unit,
+        value_min=value_min,
+        value_max=value_max,
+        resolution=resolution,
+    )
```
farmio, line 72: for `DPTEnum` we can use `dpt.get_valid_values()` or `[o.name for o in dpt.get_valid_values()]` https://github.com/XKNX/xknx/blob/c99aa412899503d0c3faaec3860a59ade2b5b389/xknx/dpt/dpt.py#L375

for `DPTComplex` we have `get_schema_dict()` for the json-serializable data version https://github.com/XKNX/xknx/blob/c99aa412899503d0c3faaec3860a59ade2b5b389/xknx/dpt/dpt.py#L411

for `DPTString` the `payload_length` might be useful - it defines the max. length of the string.

xknx/mcp/tools.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def _concrete_dpts() -> list[type[DPTBase]]:
+    """Every concrete DPT transcoder (those with a main number), ordered by main then sub."""
+    seen: dict[str, type[DPTBase]] = {
+        dpt.dpt_number_str(): dpt
+        for dpt in DPTBase.dpt_class_tree()
+        if dpt.dpt_main_number is not None
```
farmio, line 80: `dpt_class_tree()` already checks for concrete classes. All concrete classes must have a dpt_main_number so this check is redundant.

xknx/mcp/tools.py
```diff
@@ -0,0 +1,216 @@
...
+    matches: list[DptSummary] = []
+    for dpt in _concrete_dpts():
+        if filters.main is not None and dpt.dpt_main_number != filters.main:
+            continue
+        summary = _summarize_dpt(dpt)
+        if needle is not None and needle not in (
+            f"{summary.dpt}\n{summary.value_type or ''}\n{summary.unit or ''}".lower()
+        ):
+            continue
+        matches.append(summary)
```
farmio, line 102: I'd omit `_concrete_dpts()` for `dpt_class_tree()` here and sort the filtered list then.

xknx/mcp/tools.py
```diff
@@ -0,0 +1,216 @@
...
+        xknx, request.group_address, request.value, value_type=request.value_type
+    )
+    return SendResult(group_address=request.group_address, apci="GroupValueWrite")
+
+
+async def encode_dpt_payload(request: EncodeDptPayloadInput) -> EncodeDptPayloadResult:
+    """Encode a value using a specific DPT into its raw payload bytes."""
+    transcoder = DPTBase.get_dpt(request.value_type)
+    encoded = transcoder.to_knx(request.value)
+    payload = list(encoded.value) if isinstance(encoded, DPTArray) else [encoded.value]
```
farmio, line 199: should the docstring mention that `ValueError` is raised on not found value_type? And `ConversionError` for non-encodable values.

xknx/mcp/tools.py
martinhoefling: Done in #1880: _jsonify now uses as_dict() for DPTComplexData and .name for DPTEnumData.

xknx/mcp/tools.py
martinhoefling: Done in #1880: documented the raised ValueError (unknown value_type) and ConversionError (non-encodable value) on both encode/decode.

xknx/mcp/tools.py
martinhoefling: Done in #1880: DptSummary now exposes enum_values (get_valid_values), schema (get_dict_schema) for complex, and payload_length (max length for DPTString).

xknx/mcp/tools.py
martinhoefling: Done in #1880: dropped the redundant dpt_main_number guard (dpt_class_tree already yields only concrete classes).

xknx/mcp/tools.py
martinhoefling: Done in #1880: list_dpts now iterates dpt_class_tree() directly and sorts the filtered list; _concrete_dpts() removed.

PR #1877, opened 2026-07-27
fix: add explicit length checks to every APCI from_knx

kewde:
## Description

While #1871 closed the crash risk with a broad `except (IndexError, struct.error, ValueError)` wrapper around the whole `APCI.from_knx` dispatcher, this adds explicit `if len(raw) != N: raise ConversionError(...)` (or minimum-length) checks to every remaining `from_knx` method that lacked one, as defense-in-depth on top of it.

Each service now raises a specific `Invalid length for A_X in CEMI: ...` message for a malformed frame instead of relying solely on the generic dispatcher-level catch-all, matching the convention already established for the services that had this from prior PRs. Every added line has a matching regression test.

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -6,6 +6,12 @@ nav_order: 2
 
 # Changelog
 
+# Unreleased changes
+
+### Protocol
+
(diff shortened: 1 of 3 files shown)


farmio (approved):
Nice, thank you!

PR #1880, opened 2026-07-28
mcp: enrich DPT summaries/decoding (follow-up to #1876)

martinhoefling:
## Summary

Follow-up to #1876 addressing @farmio's post-approval review notes (the "after-merge" DPT refinements). Kept as a separate PR so #1876 can merge as approved.

> ⚠️ **Stacked on #1876.** It branches off `#1876`'s head, so until that merges this PR's diff also shows #1876's commits — **review the last commit only** (`mcp: enrich DPT summaries/decoding and simplify list_dpts`). Once #1876 merges, this collapses to just that commit. Happy to rebase/retarget then.

## Changes (top commit)

- **Richer `DptSummary`** — adds `payload_length`, `options` (enum DPTs, via `get_valid_values()`, lower-cased names) and `schema` (complex DPTs, via `get_dict_schema()`). `describe_dpt("1.001")` now lists `["off", "on"]`, `describe_dpt("2.001")` carries its field schema, and `describe_dpt("16.000")` reports the 14-char max length.
- **Better value decoding** — `_jsonify` renders complex DPT values with `as_dict()` and enum values with their lower-cased member `name` (e.g. `"on"` not `"Switch.ON"`), instead of a bare `str()`.
- **Simpler `list_dpts`** — iterates `dpt_class_tree()` directly (already only concrete transcoders, each with a main number) and sorts the filtered list, dropping the redundant `_concrete_dpts()` helper and main-number guard; only the returned page is summarised.
- **Docstrings** — `encode_dpt_payload` / `decode_dpt_payload` now document the raised `ValueError` (unknown `value_type`) and `ConversionError` (bad value/payload).

Maps to @farmio's inline comments on `tools.py` lines 72/151, 80/102, and 199. Passes ruff / mypy / pylint / codespell; new tests cover the enum/complex/string summaries and complex/enum decoding — `xknx.mcp` stays at 100% coverage.

### DIFF
--- a/xknx/mcp/tools.py
+++ b/xknx/mcp/tools.py
@@ -15,7 +15,16 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, TypeVar
 
 from xknx.core.connection_state import XknxConnectionState
-from xknx.dpt import DPTArray, DPTBase, DPTBinary, DPTNumeric
+from xknx.dpt import (
+    DPTArray,
+    DPTBase,
+    DPTBinary,
+    DPTComplex,
+    DPTComplexData,
+    DPTEnum,
+    DPTEnumData,
+    DPTNumeric,
+)
 from xknx.tools import (
     group_value_read,
     group_value_write,
@@ -69,20 +78,23 @@ def _summarize_dpt(dpt: type[DPTBase]) -> DptSummary:
         value_min=value_min,
         value_max=value_max,
         resolution=resolution,
+        payload_length=dpt.payload_length,
+        enum_values=(
+            [member.name for member in dpt.get_valid_values()]
+            if issubclass(dpt, DPTEnum)
+            else None
+        ),
+        schema=(
+            [dict(field) for field in dpt.get_dict_schema()]
+            if issubclass(dpt, DPTComplex)
+            else None
+        ),
     )
 
 
-def _concrete_dpts() -> list[type[DPTBase]]:
-    """Every concrete DPT transcoder (those with a main number), ordered by main then sub."""
-    seen: dict[str, type[DPTBase]] = {
-        dpt.dpt_number_str(): dpt
-        for dpt in DPTBase.dpt_class_tree()
-        if dpt.dpt_main_number is not None
-    }
-    return sorted(
-        seen.values(),
-        key=lambda dpt: (dpt.dpt_main_number or 0, dpt.dpt_sub_number or -1),
-    )
+def _dpt_haystack(dpt: type[DPTBase]) -> str:
+    """Return the lower-cased text a ``list_dpts`` text filter matches against."""
+    return f"{dpt.dpt_number_str()}\n{dpt.value_type or ''}\n{dpt.unit or ''}".lower()
 
 
 async def list_dpts(filters: DptFilter | None = None) -> DptListResult:
@@ -90,20 +102,19 @@ async def list_dpts(filters: DptFilter | None = None) -> DptListResult:
     filters = filters or DptFilter()
     needle = filters.text.lower() if filters.text else None
 
-    matches: list[DptSummary] = []
-    for dpt in _concrete_dpts():
-        if filters.main is not None and dpt.dpt_main_number != filters.main:
-            continue
-        summary = _summarize_dpt(dpt)
-        if needle is not None and needle not in (
-            f"{summary.dpt}\n{summary.value_type or ''}\n{summary.unit or ''}".lower()
-        ):
-            continue
-        matches.append(summary)
+    # dpt_class_tree() already yields only concrete transcoders (each with a
+    # main number); filter, then order by DPT number.
+    matches = [
+        dpt
+        for dpt in DPTBase.dpt_class_tree()
+        if (filters.main is None or dpt.dpt_main_number == filters.main)
+        and (needle is None or needle in _dpt_haystack(dpt))
+    ]
+    matches.sort(key=lambda dpt: (dpt.dpt_main_number or 0, dpt.dpt_sub_number or -1))
 
     window, limit_reached = _paginate(matches, filters.limit, filters.offset)
     return DptListResult(
-        dpts=window,
+        dpts=[_summarize_dpt(dpt) for dpt in window],
         total_count=len(matches),
         offset=filters.offset,
         next_offset=filters.offset + len(window) if limit_reached else None,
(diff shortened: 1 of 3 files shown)


xknx/mcp/tools.py
```diff
@@ -141,9 +152,14 @@ def _jsonify(value: object) -> GroupValue:
...
-    output, or the raw payload value). Tuples become lists; anything not already
-    JSON-native (e.g. an enum or dataclass from a complex DPT) is stringified.
+    output, or the raw payload value). Complex DPT values become their
+    ``as_dict()`` form and enum values their member name; tuples become lists;
+    anything else JSON-native passes through, and the rest is stringified.
     """
+    if isinstance(value, DPTComplexData):
+        return value.as_dict()
+    if isinstance(value, DPTEnumData):
+        return value.name
```
farmio: just fyi: in HA we use `value.name.lower()`. That's also what you see in Group Monitor. Sending enum values works with both, the name (case insensitive) or the actual value. Or the Enum class instance of course.

xknx/mcp/types.py
```diff
@@ -43,6 +43,9 @@ class DptSummary:
     value_min: float | None
     value_max: float | None
     resolution: float | None
+    payload_length: int | None  # DPTArray byte / DPTBinary bit length; e.g. string max length
+    enum_values: list[str] | None  # for enum DPTs: the valid value names
```
farmio: can we call it `options` to be in line with what a `schema` for DPTComplex would yield for an enum type?
https://github.com/XKNX/xknx/blob/c99aa412899503d0c3faaec3860a59ade2b5b389/xknx/dpt/dpt.py#L387

xknx/mcp/tools.py
martinhoefling: Good call — switched the enum branch to `value.name.lower()` so standalone decodes match HA, the Group Monitor, and the `as_dict()` enum form (`decode_dpt_payload("1.001")` → `"on"`, consistent with the complex `{"value": "on"}` case). Done in 7beb896.

xknx/mcp/types.py
martinhoefling: Renamed `enum_values` → `options` to match the complex-DPT enum-field schema. Also lower-cased the values (`m.name.lower()`) so `options` is byte-for-byte what `get_dict_schema()` emits for an enum field (dpt.py L469). Done in 7beb896.

farmio (commented):
Claude Code  🤖:

Verified locally on the PR head (`7beb896f`, 2 commits on top of `main`): 16/16 mcp tests pass, mypy and pylint clean, and `list_dpts` output is **byte-identical to main** in order and content (230 entries) — the `_concrete_dpts()` removal is a true no-op refactor.

The decoding change is a real fix, not cosmetics: `str()` produced `"Switch.ON"`, which `to_knx` cannot parse back, so decode→encode round-trips were broken. Now they work:

```
1.001   [1]      -> "on"                                   -> [1]     ✓
2.001   [3]      -> {"control": true, "value": "on"}        -> [3]     ✓
3.007   [9]      -> {"control": "increase", "step_code": 1} -> [9]     ✓
232.600 [1,2,3]  -> {"red":1,"green":2,"blue":3}            -> [1,2,3] ✓
```

## Should fix: `payload_length` is ambiguous without `payload_type`

`DPTBase.payload_length` means **bytes** for `DPTArray` and **bits** for `DPTBinary`, and `DptSummary` exposes no `payload_type`. The emitted JSON is therefore indistinguishable for two very different wire shapes:

| DPT | payload_type | `payload_length` in JSON |
|---|---|---|
| 1.001 | DPTBinary | `1` |
| 5.001 | DPTArray | `1` |
| 2.001 | DPTBinary | `2` |
| 9.001 | DPTArray | `2` |

This matters precisely for the consumer this package exists for: `DecodeDptPayloadInput.payload` documents "list of byte integers, **or a single integer** for 6-bit DPTs like DPT 1/2/3" — the field an LLM would use to pick between those two forms is exactly the one that is ambiguous. Reading `payload_length: 2` on 2.001, the natural move is `payload=[0, 0]`, which raises.

Suggest emitting `payload_type: "array" | "binary"` alongside it, or splitting into `payload_bytes` / `payload_bits`. The `# DPTArray byte / DPTBinary bit length` comment in `types.py` documents the ambiguity for a Python reader, but the JSON consumer never sees it.

## Consider

**Token cost of the always-present fields in `list_dpts`.** Serialised full catalogue grows 29 750 → 51 660 chars (**+74 %**); 171 of 230 entries carry `"options": null, "schema": null`. Since `describe_dpt` already exists for the drill-down, the enum/complex detail arguably belongs there rather than in every listing row — either move `options`/`schema` onto `DptDetail`, or drop `None` fields at serialisation. Default `limit` is 200, so this hits the common unfiltered call.

**`schema` is a slightly misleading name.** It is a field list (`name`/`type`/`required`/`options`), not JSON Schema — an LLM may well try to read it as one. `fields` would match what `get_dict_schema()` actually returns.

**Dead fallback in the new sort key** (`tools.py`, `matches.sort(key=...)`). Dropping the `dpt_main_number is not None` filter is fine — I confirmed all 230 tree entries have a main number and there are no duplicate DPT numbers — but the key still carries `dpt.dpt_main_number or 0` for the `None` case the guard used to handle. Either drop that fallback (the `or -1` on `dpt_sub_number` is still needed — 11 main-only classes like `DPT2ByteFloat`), or keep the guard. Right now the invariant is asserted in the comment and hedged in the code.

**`.name.lower()` is now in four places** — twice in `_summarize_dpt`/`_jsonify` here, twice in `DPTComplexData.get_dict_schema()`. A property on `DPTEnumData` would make the "serialisable enum name" rule single-sourced.

**`test_jsonify_shapes` no longer exercises what it reads like.** It asserts `_jsonify(_Mode.AUTO) == str(_Mode.AUTO)` with a plain `Enum` — still correct, but the new `DPTEnumData` / `DPTComplexData` branches are only covered transitively through `decode_dpt_payload`. Worth adding both directly there.

## Adjacent, pre-existing (not this PR)

`encode_dpt_payload` emits `{"payload": [true]}` for `1.001`, though `EncodeDptPayloadResult.payload` is typed `list[int]` — `DPTBinary.value` is a `bool` for DPT 1.x and `[encoded.value]` passes it straight through. Same class of "JSON-native output an LLM has to interpret" problem this PR is cleaning up, so it might be worth folding in an `int()` coercion.

## Notes

- Nothing needed for the changelog: `xknx.mcp` is still under *Unreleased*, so the added `DptSummary` fields are not breaking. Optionally worth a mention in the existing feature bullet.
- `ruff format` wants to rewrap the three new trailing comments in `types.py` into the parenthesised `payload_length: (\n    int | None\n)  # ...` form — but it already wants that for the pre-existing `next_offset` line on `main`, and tox runs `ruff format` without `--check`, so CI will not flag it. Consistent with the file as it stands; only relevant if you would rather move those comments above the fields.

farmio:
- for `payload_length`: In HA we use `0` for every `DPTBinary` DPT. They are either DPTEnum or DPTComplex so a consumer would have encoding guidelines anyway.
`payload_type: "array" | "binary"` also seems to be a reasonable way. Your call.
- Dropping `None` fields (options, schema) seems to be a good choice to me.

martinhoefling:
Addressed in #1891:

- **`payload_type: "array" | "binary"`** added to `DptSummary` — disambiguates `payload_length` (bytes vs bits).
- **`schema` → `fields`** renamed to avoid JSON Schema confusion.
- **`options`/`fields` moved to `DptDetail` only** — `list_dpts` no longer emits them at all (35k chars for 230 DPTs, down from ~52k). `describe_dpt` provides the drill-down.
- **`encode_dpt_payload` `int()` coercion** — binary payloads now emit `[1]` instead of `[true]`.
- **Sort key cleanup** — dropped dead `or 0` fallback on `dpt_main_number`.

PR #1882, opened 2026-07-30
Send L_Data_Extended frames for APDUs longer than 15 octets

### LINKED ISSUE #1868: xknx with secure tunneling does crash if payload is 2 bytes or bigger
**xknx with secure tunneling does crash if payload is 2 bytes or bigger:**
Everything works when I deactivate KNX-Secure on the group address.

- [x] using xknx standalone

**Version information:**
- xknx / Home-Assistant release with the issue: 3.16.0
- last working xknx / Home-Assistant release (if known): don't know

**Python snippet I tested:**
```python
import asyncio
from xknx import XKNX
from xknx.tools import group_value_write
from xknx.io import ConnectionConfig, ConnectionType, SecureConfig

secure_config = SecureConfig(
    knxkeys_file_path="./Keyfile.knxkeys",
    knxkeys_password="password",
)
connection_config = ConnectionConfig(
    connection_type=ConnectionType.TUNNELING_TCP_SECURE,
    gateway_ip="192.168.5.50",
    individual_address="1.1.253",
    secure_config=secure_config,
)

async def main():
    async with XKNX(connection_config=connection_config) as xknx:
        # following 1 bit payload works
        group_value_write(xknx, "1/1/78", True) # Works as intended
        # following >2 byte payload crashes
        group_value_write(xknx, "1/1/78", {"red": 255, "green": 0, "blue": 5}, value_type="232.600") # crashes

asyncio.run(main())
```

**Traceback:**
```
Fatal error: protocol.data_received() call failed.
protocol: <xknx.io.transport.tcp_transport.TCPTransport.TCPTransportFactory object at 0x7fbd24c070c0>
transport: <_SelectorSocketTransport fd=7 read=polling write=<idle, bufsize=0>>
  File "/usr/lib/python3.14/asyncio/selector_events.py", line 1026, in _read_ready__data_received
    self._protocol.data_received(data)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/io/transport/tcp_transport.py", line 61, in data_received
    self.data_received_callback(data)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/io/transport/tcp_transport.py", line 112, in data_received_callback
    self.handle_knxipframe(knxipframe, self.remote_hpai)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/io/ip_secure.py", line 357, in handle_knxipframe
    super().handle_knxipframe(knxipframe, source)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/io/transport/ip_transport.py", line 74, in handle_knxipframe
    callback.callback(knxipframe, source, self)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/io/tunnel.py", line 374, in _request_received
    self._tunnelling_request_received(knxipframe.body)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/io/tunnel.py", line 384, in _tunnelling_request_received
    self.cemi_received_callback(tunneling_request.raw_cemi)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/io/knxip_interface.py", line 477, in cemi_received
    self.xknx.cemi_handler.handle_raw_cemi(raw_cemi)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/cemi/cemi_handler.py", line 100, in handle_raw_cemi
    cemi = CEMIFrame.from_knx(raw_cemi)
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/cemi/cemi_frame.py", line 626, in from_knx
    data=CEMILData.from_knx(remainder),
         ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/cemi/cemi_frame.py", line 257, in from_knx
    payload = APCI.from_knx(_apdu)
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/telegram/apci.py", line 223, in from_knx
    return SecureAPDU.from_knx(raw)
           ~~~~~~~~~~~~~~~~~~~^^^^^
  File "/home/lukas/Documents/xknx-test/.venv/lib/python3.14/site-packages/xknx/telegram/apci.py", line 1788, in from_knx
    scf=SecurityControlField.from_knx(raw[2]),
                                      ~~~^^^
IndexError: index out of range
```

farmio:
## Description

`CEMILData.init_from_telegram()` hardcoded `CEMIFlags.FRAME_TYPE_STANDARD` and `to_knx()` never revisited that bit. An L_Data_Standard frame encodes the NPDU length in 4 bit only, so it can carry at most 15 octets after the TPCI octet (3/2/2 §2.2.4).

KNX Data Secure adds a fixed 12 octets to the plain APDU (`2 + 6 seq nr + plain APDU + 4 MAC`), so every secured group write with 2 or more octets of payload was serialized as a **standard** frame with an illegal length octet:

| plain payload | secured `npdu_len` | fits standard? |
|---|---|---|
| `GroupValueRead` / `DPTBinary` | 14 | yes |
| `DPTArray` 1 byte | 15 | yes (at the limit) |
| `DPTArray` 2 bytes (DPT 9.x) | 16 | **no** |
| `DPTArray` 3 bytes (DPT 232.600) | 17 | **no** |

The interface truncates such a frame to the 4 bit length field (`0x11 & 0x0F = 1`), so it came back as a 2 octet stub - a bare `03 f1` Secure APCI - which crashed `SecureAPDU.from_knx()` with an `IndexError`. That crash itself is already contained by f0d6eb94, but the illegal frame on the wire is not and the telegram never arrives. Plain (unsecured) frames were affected the same way for any APDU longer than 15 octets.

### Changes

- **Frame Type is derived from the NPDU length in `CEMILData.to_knx()`** - standard frame if the NPDU fits in 15 octets, extended frame otherwise. This has to happen at serialization time rather than in `init_from_telegram()`, because Data Secure replaces the payload *after* the `CEMILData` is created and `copy()`s its flags - which is exactly the case in #1868. `self.flags` is left untouched, so `to_knx()` stays free of side effects.
- **`ConversionError` for an NPDU longer than 254 octets** (255 is the escape code - 3/2/2 §2.2.5.6). This previously escaped as a bare `OverflowError` from `int.to_bytes`, which neither `CEMIHandler.send_telegram` nor `Tunnel` catches.
- **Parsing rejects a non-zero Extended Frame Format field** with `UnsupportedCEMIMessage`. LTE-HEE frames (EFF `01xxb`) use zone addressing, so their address fields can not be parsed as `GroupAddress` - they were silently mis-parsed as group telegrams before. The Frame Type flag itself is deliberately *not* validated against the NPDU length: "any receiver shall be tolerant towards the use of the Frame Format" (3/6/3 §4.1.5.2.3).
- **`CEMIFlags.EXTENDED_FRAME_FORMAT` removed.** Its value `0x0001` is a reserved EFF value, not an extended frame indicator - `0x0000` is used for standard frames *and* for long extended frames, `01xxb` denotes LTE. It was dead code. `LTE_FRAME_FORMAT` and `EXTENDED_FRAME_FORMAT_MASK` are added instead.

### Notes for reviewers

**The Data Secure MAC is not affected.** `block_0()` masks the 16 bit cEMI flags with `B0_AT_FIELD_FLAGS_MASK = 0b10001111`, which only reaches Ctrl2 (Address Type bit + EFF nibble) - Ctrl1, and therefore the Frame Type, is excluded from the CCM input. Two independent vectors already in the test suite confirm this empirically: the real-device group frame fixture has Ctrl1 `0x3C` (FT = 0) and the AN158 v07 Annex A example has Ctrl1 `0xB0` (FT = 1), and both verify today under a mask that drops the bit. `test_data_secure_group_send_extended` now also round-trips encrypt -> serialize -> parse -> decrypt across a flipped Frame Type.

**Two existing fixtures changed.** `test_valid_command` and `test_valid_tpci_control` used flags with FT = 0 together with a short payload, which is not a frame xknx would ever emit; their `to_knx() == raw` round trip now requires `FRAME_TYPE_STANDARD`. `test_data_secure_authentication_only` compares telegrams instead of whole `CEMILData` objects, because the pre-serialization frame still carries the default standard flag while the serialized frame went out as extended.

Verified end-to-end against the reported case - the boundary sits exactly where the spec puts it:

```
GroupValueRead         ctrl1=0xbc standard  npdu_len= 14
1 byte  (DPT 5.001)    ctrl1=0xbc standard  npdu_len= 15
2 bytes (DPT 9.001)    ctrl1=0x3c EXTENDED  npdu_len= 16
3 bytes (DPT 232.600)  ctrl1=0x3c EXTENDED  npdu_len= 17
14 bytes (DPT 16.000)  ctrl1=0x3c EXTENDED  npdu_len= 28
```

Fixes #1868

## Type of change

## Checklist

### DIFF
--- a/xknx/cemi/cemi_frame.py
+++ b/xknx/cemi/cemi_frame.py
@@ -27,7 +27,13 @@ from xknx.telegram import GroupAddress, IndividualAddress, Telegram
 from xknx.telegram.apci import APCI
 from xknx.telegram.tpci import TPCI, TDataBroadcast
 
-from .const import CEMIErrorCode, CEMIFlags, CEMIMessageCode
+from .const import (
+    MAX_NPDU_LENGTH,
+    STANDARD_FRAME_MAX_NPDU_LENGTH,
+    CEMIErrorCode,
+    CEMIFlags,
+    CEMIMessageCode,
+)
 
 
 class CEMIInfo:
@@ -138,6 +144,7 @@ class CEMILData(CEMIData):
     ) -> CEMILData:
         """Return CEMILData from a Telegram."""
         flags = (
+            # default; the Frame Type is derived from the NPDU length in `to_knx()`
             CEMIFlags.FRAME_TYPE_STANDARD
             | CEMIFlags.DO_NOT_REPEAT
             | CEMIFlags.BROADCAST
@@ -191,8 +198,24 @@ class CEMILData(CEMIData):
             tpdu[0] |= self.tpci.to_knx()
             npdu_len = self.payload.calculated_length()
 
+        if npdu_len > MAX_NPDU_LENGTH:
+            raise ConversionError(
+                f"APDU too long for a single frame: {npdu_len} octets; "
+                f"maximum is {MAX_NPDU_LENGTH}"
+            )
+        # The Frame Type is a function of the NPDU length: an L_Data_Standard frame
+        # can carry at most 15 octets after the TPCI octet, and an L_Data_Extended
+        # frame shall not be used when a standard frame suffices - 3/2/2 §2.2.5.1.
+        # Derived here instead of when the frame is created because the payload may
+        # be replaced afterwards - eg. wrapped in a SecureAPDU by Data Secure.
+        flags = (
+            self.flags | CEMIFlags.FRAME_TYPE_STANDARD
+            if npdu_len <= STANDARD_FRAME_MAX_NPDU_LENGTH
+            else self.flags & ~CEMIFlags.FRAME_TYPE_STANDARD
+        )
+
         return (
-            self.flags.to_bytes(2, "big")
+            flags.to_bytes(2, "big")
             + self.src_addr.to_knx()
             + self.dst_addr.to_knx()
             + npdu_len.to_bytes(1, "big")
@@ -211,6 +234,18 @@ class CEMILData(CEMIData):
         flags = int.from_bytes(raw[0:2], "big")
 
         src_addr = IndividualAddress.from_knx(raw[2:4])
+
+        # The Extended Frame Format defines how the address fields are to be
+        # interpreted - 3/2/2 §2.2.5.3. Only 0000b (standard and long frames) is
+        # supported; 01xxb are LTE-HEE zone addressed frames, the rest is reserved.
+        # The Frame Type flag itself is deliberately not validated against the NPDU
+        # length: "any receiver shall be tolerant towards the use of the Frame
+        # Format" - 3/6/3 §4.1.5.2.3.
+        if _eff := flags & CEMIFlags.EXTENDED_FRAME_FORMAT_MASK:
+            raise UnsupportedCEMIMessage(
+                f"Extended Frame Format not supported: {_eff:#06b} "
+                f"from {src_addr} in CEMI: {raw.hex()}"
+            )
 
         _dst_is_group_address = bool(flags & CEMIFlags.DESTINATION_GROUP_ADDRESS)
         dst_addr: GroupAddress | IndividualAddress = (
--- a/xknx/cemi/cemi_frame.py
+++ b/xknx/cemi/cemi_frame.py
@@ -129,28 +129,6 @@ class CEMILData(CEMIData):
         # Setting new hops
         self.flags |= val << 4
 
-    def npdu_length(self) -> int:
-        """Get number of NPDU octets; the TPCI octet is not included."""
-        if self.tpci.control:
-            return 0
-        if not isinstance(self.payload, APCI):
-            raise ConversionError(
-                f"Invalid payload set for data TPDU: {type(self.payload)}"
-            )
-        return self.payload.calculated_length()
-
-    @property
-    def frame_type_standard(self) -> bool:
-        """
-        Return True if this is serialized as L_Data_Standard frame.
-
-        The Frame Type is not stored in `flags` for frames created by XKNX; it is a
-        function of the NPDU length. An L_Data_Standard frame can carry at most 15
-        octets after the TPCI octet and an L_Data_Extended frame shall not be used
-        when a standard frame suffices - 3/2/2 §2.2.4 and §2.2.5.1.
-        """
-        return self.npdu_length() <= STANDARD_FRAME_MAX_NPDU_LENGTH
-
     def calculated_length(self) -> int:
         """Get length of KNX/IP body."""
         if not self.tpci.control and self.payload is not None:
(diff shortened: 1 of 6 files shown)


xknx/cemi/cemi_frame.py
```diff
@@ -191,8 +198,24 @@ def to_knx(self) -> bytes:
...
+            raise ConversionError(
+                f"APDU too long for a single frame: {npdu_len} octets; "
+                f"maximum is {MAX_NPDU_LENGTH}"
+            )
+        # The Frame Type is a function of the NPDU length: an L_Data_Standard frame
+        # can carry at most 15 octets after the TPCI octet, and an L_Data_Extended
+        # frame shall not be used when a standard frame suffices - 3/2/2 §2.2.5.1.
+        # Derived here instead of when the frame is created because the payload may
+        # be replaced afterwards - eg. wrapped in a SecureAPDU by Data Secure.
+        flags = (
```
kewde, line 211: We make a copy of the flags, but don't update the internal state, I'm not sure if that's ideal. I wonder if we can mutate self.flags directly..

kewde (approved):
Hah, I tried to fix the issue as well! 

One of the concerns I have is that CEMILData accepts flags, so a consumer of the class can instantiate it explicitly with the standard frame format flag, but to_knx() silently upgrades it to an extended frame format.

That's a bit of an unexpected behavior, typically solved with factories to prevent such an invalid state.

But it's a reasonable fix imho, one thing that I did different in my PR was mutate self.flags rather than copy-and-edit-before-serialization. That at least reflects the change in the internal state too

xknx/cemi/cemi_frame.py
farmio: logging of the cemi frame happens before calling `to_knx`, after it the instance is discarded anyway. So it wouldn't show in current implementation. Let's keep `to_knx` side-effect free for this fix and refactor flags in a follow up for the next major release.

PR #1884, opened 2026-07-31
Refactor CEMILData.flags into a typed CEMIFlags dataclass

farmio:
## Description

Follow-up to #1882, targeted at the next major release. `CEMILData.flags` was a bare 16 bit `int` and `CEMIFlags` a bag of bit constants, so callers had to know the layout of Ctrl1 and Ctrl2 to read or set anything, and `hops` needed a hand written mask/shift property.

`CEMIFlags` is now a slotted dataclass holding the control field values that are actually independent state:

```python
@dataclass(slots=True)
class CEMIFlags:
    priority: CEMIPriority = CEMIPriority.LOW
    # `repeat_on_error` and `system_broadcast` are inverted on the wire
    repeat_on_error: bool = False
    system_broadcast: bool = False
    acknowledge_request: bool = False
    confirm_error: bool = False
    hop_count: int = 6
    # as received; not used when serializing
    frame_type: CEMIFrameType = CEMIFrameType.STANDARD
    frame_format: CEMIFrameFormat = CEMIFrameFormat.STANDARD
```

`enum.Flag` was considered and does not fit: priority is a 2 bit value and the hop count a 3 bit counter, neither of which are flags. A plain slotted dataclass gives free `__eq__`, keyword construction and mutability for the hop count, which is what a router would need.

### Each part serializes itself

The three fields that are not plain flags get their own type and place their own bits, instead of being masked in and out by hand:

| field | type | where it comes from |
|---|---|---|
| Frame Type (Ctrl1 b7) | `CEMIFrameType` | derived when serializing: NPDU length - `<= 15` is a standard frame |
| Address Type (Ctrl2 b7) | `CEMIAddressType` | derived when serializing: `CEMILData.address_type`, ie. the type of the destination address |
| Extended Frame Format (Ctrl2 b3..0) | `CEMIFrameFormat` | held by `CEMIFlags`; `STANDARD` for every frame xknx supports |

`CEMIFlags.to_knx()` leaves the two derived bits clear and the frame ORs them in from the values only it knows:

```python
self.flags.to_knx() | frame_type.to_knx() | self.address_type.to_knx()
```

This means `flags` can no longer disagree with what is put on the wire - the class of bug that made #1868 possible - and there is no stale Frame Type left to go wrong when Data Secure replaces the payload of a frame it copied.

Frame Type and Extended Frame Format are still *parsed* into the dataclass, but they are not the same kind of thing and are not treated the same:

- `frame_type` is informational. It holds what was received, for inspection, and is ignored when serializing, since the Frame Type of an outgoing frame follows from its NPDU length.
- `frame_format` is real state - it does not follow from anything else - so it is serialized as held. Data Secure protects the Extended Frame Format, so the value on the wire and the one fed to the MAC have to come from the same place; a constant in both spots would only be kept in sync by a comment. It is `STANDARD` throughout today: LTE-HEE frames are rejected while parsing and cannot be constructed.

### On the `int | None` idea

Storing raw incoming flags in a private attribute was considered and rejected. Parsing straight into the dataclass loses nothing but the reserved Ctrl1 b6 - the Frame Type is kept, and everything else round-trips, including for a received frame. Letting a stored raw value take precedence when serializing would reintroduce exactly the stale-flags problem, since a `copy()`d frame with a swapped payload would emit the old Frame Type. If you want byte-exact fidelity for debugging, the raw octets of a received frame are already available in `CEMIHandler.handle_raw_cemi` - keeping a second, partly authoritative copy on the object is the part that does not pay for itself.

Happy to add it as a debug-only field if you disagree - it is a small addition on top of this.

### Also in this change

- `CEMILData.hops` is replaced by `flags.hop_count`, validated to `0..7` when serializing instead of silently corrupting Ctrl2. `CEMILData.address_type` is added.
- The raw bit masks move to `xknx.cemi.flags` as module level constants. Ctrl1 and Ctrl2 stay one 16 bit value (`Ctrl1 << 8 | Ctrl2`), so `CEMIFlags.to_knx()` returns an `int` and `from_knx()` takes one; the bits that have a type of their own are not among the constants.
- `block_0()`, `SecureData.init_from_plain_apdu()` and `SecureData.get_plain_apdu()` take `address_type: CEMIAddressType` and `frame_format: CEMIFrameFormat` instead of `frame_flags: int`. Only those two fields of Ctrl2 ever survived `B0_AT_FIELD_FLAGS_MASK`, so the mask is gone and the B0 octet is spelled out as the `A000EEEEb` of KNX v02.01.01 - Application Layer 03.03.07 - §5.1.3.2 Figure 100. The CCM input is unchanged.
- `CEMILData(flags=...)` is optional and defaults to `CEMIFlags()`.
- `CEMIFlags` has a compact `__str__` used in `CEMILData.__repr__` - eg. `LOW STANDARD hop_count=6`, listing only the boolean flags that are set.

### Verification

The Data Secure byte vectors are the strongest check that nothing moved: the AN158 v07 Annex A example and the real-device frame still verify, and the secured outgoing frames are byte-identical to what `main` produces, including the extended-frame vector from #1882:

```
3ce0500104001103f11000254ae1cb67cd98e577b519be47bb
```

New `test/cemi_tests/flags_test.py` covers parsing, the priority encoding from 3/2/2 Figure 28, the Extended Frame Format including its reserved values, the composition of the derived bits, a Ctrl1/Ctrl2 round trip and hop count validation. Full suite passes (3875); `test/integration_tests/routing_multicast_test.py` fails on my machine for lack of multicast routing, identically on `main`.

Rebased onto `main` after the 3.20.0 release - the code is byte-identical to the previously reviewed version, only the changelog entries moved into the new `# Unreleased changes` section.

## Type of change

## Checklist

### DIFF
--- a/xknx/cemi/cemi_frame.py
+++ b/xknx/cemi/cemi_frame.py
@@ -34,7 +34,7 @@ from .const import (
     CEMIMessageCode,
 )
 from .flags import (
-    DESTINATION_GROUP_ADDRESS,
+    CEMIAddressType,
     CEMIFlags,
     CEMIFrameFormat,
     CEMIFrameType,
@@ -123,9 +123,13 @@ class CEMILData(CEMIData):
         self.payload = payload
 
     @property
-    def dst_is_group_address(self) -> bool:
-        """Return True if the destination address is a group address."""
-        return isinstance(self.dst_addr, GroupAddress)
+    def address_type(self) -> CEMIAddressType:
+        """Return the Address Type of the destination address."""
+        return (
+            CEMIAddressType.GROUP
+            if isinstance(self.dst_addr, GroupAddress)
+            else CEMIAddressType.INDIVIDUAL
+        )
 
     def calculated_length(self) -> int:
         """Get length of KNX/IP body."""
@@ -195,15 +199,15 @@ class CEMILData(CEMIData):
         # frame shall not be used when a standard frame suffices - 3/2/2 §2.2.5.1.
         # Derived here instead of when the frame is created because the payload may
         # be replaced afterwards - eg. wrapped in a SecureAPDU by Data Secure.
+        frame_type = (
+            CEMIFrameType.STANDARD
+            if npdu_len <= STANDARD_FRAME_MAX_NPDU_LENGTH
+            else CEMIFrameType.EXTENDED
+        )
         return (
-            self.flags.to_knx(
-                frame_type=(
-                    CEMIFrameType.STANDARD
-                    if npdu_len <= STANDARD_FRAME_MAX_NPDU_LENGTH
-                    else CEMIFrameType.EXTENDED
-                ),
-                dst_is_group_address=self.dst_is_group_address,
-            )
+            (
+                self.flags.to_knx() | frame_type.to_knx() | self.address_type.to_knx()
+            ).to_bytes(2, "big")
             + self.src_addr.to_knx()
             + self.dst_addr.to_knx()
             + npdu_len.to_bytes(1, "big")
@@ -221,8 +225,9 @@ class CEMILData(CEMIData):
         src_addr = IndividualAddress.from_knx(raw[2:4])
 
         # Control field 1 and Control field 2 - first 2 octets
+        _control_field = int.from_bytes(raw[0:2], "big")
         try:
-            flags = CEMIFlags.from_knx(raw[0:2])
+            flags = CEMIFlags.from_knx(_control_field)
         except ConversionError as err:
             raise UnsupportedCEMIMessage(
                 f"{err} from {src_addr} in CEMI: {raw.hex()}"
@@ -240,7 +245,9 @@ class CEMILData(CEMIData):
                 f"from {src_addr} in CEMI: {raw.hex()}"
             )
 
-        _dst_is_group_address = bool(raw[1] & DESTINATION_GROUP_ADDRESS)
+        _dst_is_group_address = (
+            CEMIAddressType.from_knx(_control_field) is CEMIAddressType.GROUP
+        )
         dst_addr: GroupAddress | IndividualAddress = (
             GroupAddress.from_knx(raw[4:6])
             if _dst_is_group_address
--- a/xknx/cemi/cemi_frame.py
+++ b/xknx/cemi/cemi_frame.py
@@ -35,8 +35,9 @@ from .const import (
 )
 from .flags import (
     DESTINATION_GROUP_ADDRESS,
-    EXTENDED_FRAME_FORMAT_MASK,
     CEMIFlags,
+    CEMIFrameFormat,
+    CEMIFrameType,
     CEMIPriority,
 )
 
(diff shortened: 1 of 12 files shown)


xknx/cemi/cemi_frame.py
```diff
@@ -241,13 +224,13 @@ def from_knx(cls, raw: bytes) -> CEMILData:
         # The Frame Type flag itself is deliberately not validated against the NPDU
         # length: "any receiver shall be tolerant towards the use of the Frame
         # Format" - 3/6/3 §4.1.5.2.3.
-        if _eff := flags & CEMIFlags.EXTENDED_FRAME_FORMAT_MASK:
+        if _eff := raw[1] & EXTENDED_FRAME_FORMAT_MASK:
```
farmio: I don't like this this check runs on raw. Should we add an enum / dataclasses that holds frametype? 
I'd like to be able to hold (log) it for incoming frames while auto-setting it for outgoing. It would be merged with the flags `.to_knx()` and the address type when sending. 

For address type we can see the received value from the type of `dst_addr`, so that's fine. Original frame format however is lost.

farmio: Claude Code  🤖: Done in c072f10a — both the typed frame type and the parsed check.

`CEMIFlags` gains two informational fields that hold what was received and are ignored when serializing:

```python
frame_type: CEMIFrameType = CEMIFrameType.STANDARD    # EXTENDED / STANDARD
frame_format: CEMIFrameFormat = CEMIFrameFormat.STANDARD  # STANDARD / LTE_HEE
```

`to_knx()` takes the Frame Type as a `CEMIFrameType` from `CEMILData`, which derives it from the NPDU length, so `self.frame_type` is never consulted when sending — exactly the "hold it for incoming, auto-set for outgoing, merge in `to_knx()`" split you described. `test_to_knx_ignores_received_frame_type` pins that.

The EFF check no longer touches raw bits:

```python
if flags.frame_format is not CEMIFrameFormat.STANDARD:
    raise UnsupportedCEMIMessage(
        f"Extended Frame Format not supported: {flags.frame_format.name} ..."
    )
```

`CEMIFrameFormat._missing_` resolves the whole `01xxb` range to `LTE_HEE` (the 2 lsb are the LTE extended address type, so enumerating them separately would be noise). Reserved encodings raise `ConversionError` out of `CEMIFlags.from_knx`, which `CEMILData.from_knx` converts to `UnsupportedCEMIMessage` with the source address and raw frame — the same pattern already used for an unsupported TPCI. Both paths are parametrised in `cemi_frame_test.py`.

One thing I did **not** do: serialize `self.frame_format`. `to_knx()` always writes `STANDARD`, because a `CEMILData` with any other format cannot exist — it is rejected during parsing. Storing it is still worth it for the typed check and for logging, but writing it back would only matter if we ever supported LTE.

Address type is unchanged, readable from `dst_addr` and also exposed as `CEMILData.dst_is_group_address`.

Also added the compact `__str__` you asked for: `LOW STANDARD hop_count=6`, listing only the booleans that are set, used by `CEMILData.__repr__` so debug log lines stay short. The full dataclass `__repr__` still shows every field.

The Data Secure byte vectors are unchanged, including `3ce0500104001103f11000254ae1cb67cd98e577b519be47bb`.

xknx/secure/data_secure_asdu.py
```diff
@@ -159,7 +162,7 @@ def init_from_plain_apdu(
                 block_0=block_0(
                     sequence_number=sequence_number_bytes,
                     address_fields_raw=address_fields_raw,
-                    frame_flags=frame_flags,
+                    address_type=address_type,
```
kewde, line 169: nice, much easier to see what gets included in the MAC 👍

kewde (approved):
utACK

pdgendt (approved):
I like this! Could maybe be added in a backwards compatible way with some getter/setter functions, but not sure if it's worth it.

farmio:
I think we can cut a new major version with this... Python 3.10 is reaching EOL soon too, so we could also cut that alongside.

PR #1894, opened 2026-08-02
mcp: describe input fields via dataclass field metadata instead of Annotated

farmio:
## Summary

Moves the per-field descriptions of the `xknx.mcp` **input** dataclasses from `Annotated[T, "..."]` to `field(metadata={"description": "..."})`. Output dataclasses are untouched.

## Why

Both forms are stdlib-only and keep the package free of any schema/validation dependency, as intended. The difference is what a consumer gets for free.

The MCP SDKs build a tool's input schema from the function's type hints, through Pydantic — from the [Python SDK's tools documentation](https://py.sdk.modelcontextprotocol.io/servers/tools/):

> From those type hints the SDK generates a JSON Schema and sends it to the client during `tools/list`

> If you've used FastAPI or Pydantic, you already know all of this. It's the same `Field`, the same `Annotated`, the same validation.

And Pydantic applies the metadata of a **stdlib** dataclass field as that field's `Field()` arguments — see ["You can use both the Pydantic's `Field()` and the stdlib's `field()` functions"](https://pydantic.dev/docs/validation/latest/concepts/dataclasses/) with its `metadata={'title': ..., 'description': ...}` example. A bare string inside `Annotated`, by contrast, is metadata Pydantic does not recognise and silently ignores.

So today's descriptions never reach the wire unless every consumer writes its own extraction loop. Same tool defined both ways against the official SDK (`mcp` 2.0.0), reading back `tool.input_schema`:

```
with_annotated -> {"group_address": {"title": "Group Address", "type": "string"}}
with_metadata  -> {"group_address": {"description": "Destination group address, e.g. \"1/2/3\".",
                                     "title": "Group Address", "type": "string"}}
```

After this change, registering the real `send_group_value_write` with the SDK yields:

```json
"GroupValueWriteInput": {
  "description": "Input for :func:`~xknx.mcp.tools.send_group_value_write`.",
  "properties": {
    "group_address": {
      "description": "Destination group address, e.g. \"1/2/3\".",
      "title": "Group Address", "type": "string"
    },
    "value_type": {
      "anyOf": [{"type": "string"}, {"type": "null"}],
      "default": null,
      "description": "DPT number (\"9.001\") or value-type name (\"temperature\"). Without it, an int is sent as a 6-bit payload and a list of ints as a byte array.",
      "title": "Value Type"
    }
  },
  "required": ["group_address", "value"]
}
```

Zero glue code in the consumer. This is not new or version-fragile behaviour — I verified identical results on Pydantic 2.5.3, 2.9.2, 2.11.7 and 2.13.4. The mapping is not limited to `description`:

| declared | resulting JSON schema |
|---|---|
| `metadata={"description": "d"}` | `"description": "d"` |
| `metadata={"title": "T", "examples": ["x"]}` | `"title": "T", "examples": ["x"]` |
| `metadata={"ge": 1}` | `"minimum": 1` |
| `metadata={"foo": "bar"}` | ignored, no error |
| `Annotated[str, "d"]` | dropped |
| `Annotated[str, Doc("d")]` (PEP 727) | dropped |
| `Annotated[str, Field(description="d")]` | `"description": "d"` — but needs a Pydantic import |

`Annotated[..., Field(...)]` is the only other form that works, and it would put a Pydantic dependency in `xknx`, which is exactly what this package avoids. Adding constraints (`ge` on `offset`, etc.) is now possible without one — deliberately not done here to keep this change about descriptions only.

Consumers that build their schema by hand — Home Assistant's `mcp_server` derives its tool schemas from voluptuous via `voluptuous_openapi`, so it needs a mapping loop either way — benefit as well: `dataclasses.fields(cls)` exposes the metadata directly, with no `get_type_hints(..., include_extras=True)` call. That also removes a fragile coupling: the current approach obliges every consumer to keep every annotation in `xknx.mcp.types` resolvable at runtime from a module they do not own, so adding a single `TYPE_CHECKING`-only import here would break their extraction with a `NameError`.

## Trade-offs

- Noisier declarations: each documented field needs a `field(...)` call and its default moves inside it.
- The description leaves the type annotation, so it no longer shows on IDE hover of the type. It stays visible in the source and via `dataclasses.fields()`.
- `metadata` is `Mapping[str, Any]`, so a typo in the key is silent — the same weakness the bare-string `Annotated` form had. The new `types_test.py` guards the `description` key for every input field.

## Testing

`ruff`, `ruff format`, `mypy`, `pylint`, `codespell` and `pytest test/mcp_tests` (22 passed) all clean. No changelog entry: `xknx.mcp` is still under *Unreleased*, so this is not a breaking change for any release.

### DIFF
--- a/test/mcp_tests/types_test.py
+++ b/test/mcp_tests/types_test.py
@@ -1,0 +1,35 @@
+"""Tests for the xknx MCP input/output dataclasses."""
+
+from dataclasses import Field, fields
+
+import pytest
+
+from xknx.mcp import (
+    DecodeDptPayloadInput,
+    DptFilter,
+    EncodeDptPayloadInput,
+    GroupAddressInput,
+    GroupValueReadInput,
+    GroupValueWriteInput,
+)
+
+
+@pytest.mark.parametrize(
+    "input_type",
+    [
+        DecodeDptPayloadInput,
+        DptFilter,
+        EncodeDptPayloadInput,
+        GroupAddressInput,
+        GroupValueReadInput,
+        GroupValueWriteInput,
+    ],
+)
+def test_input_fields_carry_a_description(input_type: type) -> None:
+    """Every input field is described, so a consumer can build its tool schema."""
+    field: Field  # type: ignore[type-arg]
+    for field in fields(input_type):
+        description = field.metadata.get("description")
+        assert isinstance(description, str) and description, (
+            f"{input_type.__name__}.{field.name} has no description metadata"
+        )
--- a/xknx/mcp/types.py
+++ b/xknx/mcp/types.py
@@ -6,16 +6,20 @@ arguments and serialise outputs with :func:`dataclasses.asdict` without custom
 encoders. DPTs are rendered as ``"main"`` or ``"main.sub"`` strings and
 timestamps as ISO-8601.
 
-Input fields carry their human-readable description as :data:`typing.Annotated`
-metadata (a plain string). This keeps the library free of any schema/validation
-dependency while letting a consumer surface per-parameter descriptions in its
-tool schema via ``typing.get_type_hints(..., include_extras=True)``.
+Input fields carry their human-readable description in the dataclass field
+metadata - ``field(metadata={"description": ...})``. This keeps the library free
+of any schema/validation dependency while letting a consumer surface
+per-parameter descriptions in its tool schema: Pydantic applies the metadata of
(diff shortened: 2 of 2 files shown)


farmio:
@martinhoefling not sure how you are looking to implement MCP server in SpectrumKnx. If you use are planning to use the sdk linked in the description, this PR might be worth a look. Otherwise feel free to close it.

kewde (approved):

PR #1897, opened 2026-08-06
fix(mcp): reorder GroupValue union types to prevent object schema generation

martinhoefling:
## Description
Reorders  type alias union in  so  precedes .

Because Python's  is a subclass of , having  first led Pydantic / FastMCP schema generators to output  type schemas for MCP tool arguments (e.g. ), rejecting integer/float payloads.

## Changes
- Update  in 
- Bump version to  for patch release


martinhoefling:
obsolete, updated the HA pr with a fix.

PR #1898, opened 2026-08-07
Parse M_PropInfo.ind cEMI frames

pdgendt:
## Description
A KNXnet/IP server sends M_PropInfo.ind to announce a property changing on its own, such as the device state of the KNXnet/IP parameter object when the KNX bus fails. The message code was already known and already excluded from carrying additional info, and serializing a frame with it worked, but from_knx() rejected it with UnsupportedCEMIMessage.

Its payload is the same as M_PropRead.con: the property, then either its value or, when the number of elements is zero, an error code. Parse it into a CEMIMPropReadResponse.

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -8,6 +8,10 @@ nav_order: 2
 
 # Unreleased changes
 
+### Features
+
+- Parse `M_PropInfo.ind` cEMI frames. A KNXnet/IP server uses them to announce a property changing on its own, for example the KNXnet/IP parameter object's device state when the KNX bus fails, and `CEMIFrame.from_knx()` raised `UnsupportedCEMIMessage` for them. The payload is the same as `M_PropRead.con`, so it is parsed into a `CEMIMPropReadResponse`. Serializing already worked.
+
 # 3.18.0 Protocol Droid 2026-08-02
 
 ### Bugfixes
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -8,9 +8,9 @@ nav_order: 2
 
 # Unreleased changes
 
-### Features
+### Protocol
 
-- Parse `M_PropInfo.ind` cEMI frames. A KNXnet/IP server uses them to announce a property changing on its own, for example the KNXnet/IP parameter object's device state when the KNX bus fails, and `CEMIFrame.from_knx()` raised `UnsupportedCEMIMessage` for them. The payload is the same as `M_PropRead.con`, so it is parsed into a `CEMIMPropReadResponse`. Serializing already worked.
+- Parse `M_PropInfo.ind` cEMI frames into a `CEMIMPropReadResponse` - its payload has the same layout as `M_PropRead.con`. A KNXnet/IP server sends these to announce a property changing on its own, e.g. the KNXnet/IP parameter object's device state when the KNX bus fails; they previously raised `UnsupportedCEMIMessage` and were counted as incoming errors. Serializing already worked.
 
 # 3.18.0 Protocol Droid 2026-08-02
 
--- a/test/cemi_tests/cemi_frame_test.py
+++ b/test/cemi_tests/cemi_frame_test.py
@@ -482,8 +482,8 @@ def test_valid_prop_info_ind() -> None:
     assert frame.to_knx() == raw
 
 
-def test_valid_error_prop_info_ind() -> None:
-    """Test for valid frame parsing."""
+def test_prop_info_ind_with_error_code() -> None:
+    """Test lenient parsing of an off-spec error payload in M_PropInfo.ind."""
     raw = get_prop(0xF7, 0x000B, 1, 69, 0, 1, [0x07])
     frame = CEMIFrame.from_knx(raw)
     assert frame.code == CEMIMessageCode.M_PROP_INFO_IND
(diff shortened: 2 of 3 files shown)


farmio (commented):
Claude Code  🤖:

Adds one dispatch branch in `CEMIFrame.from_knx` so message code `0xF7` parses into a `CEMIMPropReadResponse` instead of raising `UnsupportedCEMIMessage`, plus two tests and a changelog entry. The code was already in `CEMIMessageCode` and already excluded from `_has_info()`, so this is genuinely the one missing piece.

Correct and worth merging. Three small things to fix first.

## Correctness

The wire format checks out against 03_06_03 EMI/IMI §4.1.7.3.6:

```
Message  Interface Object  Object    Property  number of  start index  Data
 Code          Type        Instance     ID     elements
 F7h            ...           1        ...       > 0         SI        ...
```

Identical to `M_PropRead.con`'s positive form, so reusing `CEMIMPropReadResponse` is right, and Calimero does the same (one `CEMIDevMgmt` for all the M_Prop codes). The test vectors are valid too — `0x000B` is `OBJECT_KNXNETIP_PARAMETER` and PID 69 is `PID_KNXNETIP_DEVICE_STATE`.

**One spec deviation.** §4.1.7.3.6 defines `NoE > 0` for `M_PropInfo.ind` — unconditionally. There is no negative form; §4.1.7.3.7.1 scopes error responses to the *confirmed* services (`M_PropRead.req` / `M_PropWrite.req`) only, and it would be odd for an unsolicited event notification to report an error in the first place. So:

- `test_valid_error_prop_info_ind` calls an off-spec frame shape "valid" and pins it as expected behavior.
- The changelog and the inline comment both state the payload *is* "either its value or, when the number of elements is zero, an error code" — true for `.con`, not for `.ind`.

Accepting it anyway is the right call — §4.1.5.2.3 says a receiver shall be tolerant about frame format, and it falls out of the class reuse for free. But say what it is:

```python
if code == CEMIMessageCode.M_PROP_INFO_IND:
    # Same payload as M_PropRead.con. The spec defines only the positive
    # form (NoE > 0, see 3/6/3 EMI_IMI §4.1.7.3.6); a NoE == 0 error
    # payload is parsed leniently as well.
    return CEMIFrame(code=code, data=CEMIMPropReadResponse.from_knx(raw[1:]))
```

and rename the test to something like `test_prop_info_ind_with_error_code` so it doesn't read as a spec-conformant case.

## The behavior change isn't documented

This is the part the PR description undersells. Before, an `M_PropInfo.ind` hit the `UnsupportedCEMIMessage` handler in `CEMIHandler.handle_raw_cemi`: logged at info as "CEMI not supported" and incremented `cemi_count_incoming_error`. After, it parses and falls into the `not isinstance(cemi.data, CEMILData)` branch of `handle_cemi_frame` — debug log, no counter.

That's the actual user-visible benefit: an interface that pushes device-state notifications when the KNX bus drops no longer inflates the incoming-error count. Worth a sentence in the changelog. It's also worth being explicit that nothing *consumes* these yet — there is no device-management client in xknx, and `CEMIMPropReadResponse` has no usage outside `cemi_frame.py`, so the frames are parsed and dropped.

## Changelog

**Section is wrong by project convention.** Both 3.18.0 and 3.17.0 file cEMI/APCI parsing additions under `### Protocol` ("Add A_DomainAddress_Write … APCI service parsing", "Reject incoming CEMI L_Data frames with a non-zero Extended Frame Format field"). `### Features` is used for user-facing capability like `xknx.mcp`. Move it to `### Protocol`.

It is also about twice the length of neighbouring entries. Something like:

> - Parse `M_PropInfo.ind` cEMI frames into a `CEMIMPropReadResponse` — its payload has the same layout as `M_PropRead.con`. A KNXnet/IP server sends these to announce a property changing on its own, e.g. the KNXnet/IP parameter object's device state when the KNX bus fails; they previously raised `UnsupportedCEMIMessage` and were counted as incoming errors. Serializing already worked.

Note there is now a `### Bugfixes` block under "Unreleased changes" on main, so on rebase put `### Protocol` after it to match the 3.18.0 ordering.

## Latent issue this widens (pre-existing, out of scope)

`CEMIMPropReadResponse.error_code` resolves the enum lazily and `__repr__` calls it, so an unrecognized error octet raises from the *repr*, not from parsing:

```python
raw = bytes([0xF7, 0x00, 0x0B, 0x01, 0x45, 0x00, 0x01, 0x99])
frame = CEMIFrame.from_knx(raw)   # parses fine
repr(frame.data)                  # ValueError: 153 is not a valid CEMIErrorCode
```

`CEMIErrorCode` only covers `0x00`–`0x0A`. The `%s` formatting in `handle_cemi_frame` sits *outside* the broad-except guard around `from_knx`, so with debug logging enabled a device sending an out-of-range error code raises inside the datagram callback. This already applies to `M_PROP_READ_CON`, so it is not introduced here — but `M_PropInfo.ind` is unsolicited, meaning a device can now push this at you unprompted rather than only in response to a request the library made. Worth a separate PR resolving the enum at parse time (or falling back to the raw octet); I would not hold this one for it.

## Minor

- Both new tests reuse the file's `"""Test for valid frame parsing."""` docstring, consistent with every other test there — fine as-is, though the error-case one would read better with its own line.
- Good that `test_valid_prop_info_ind` asserts `calculated_length()` and the `to_knx()` roundtrip, matching the sibling tests.

farmio:
Thank you very much 👍

PR #1899, opened 2026-08-07
Add a device management connection client

pdgendt:
## Description

Adds a client for the KNXnet/IP Device Management connection type (03.08.03) — the
connection used to read and write the Properties of a server's own Interface Objects
(its KNXnet/IP Parameter Object and Device Object), separate from tunnelling and
unrelated to the KNX bus behind the server.

### New classes

- **`UDPDeviceManagementConnection`** / **`TCPDeviceManagementConnection`** /
  **`SecureDeviceManagementConnection`** — connection clients structured like
  `UDPTunnel` / `TCPTunnel` / `SecureTunnel`: an abstract base holds the shared
  connect/heartbeat/disconnect/request logic, the subclasses supply the
  transport-specific behaviour. `read_property()` and `write_property()` cover the
  Property services, `request()` sends any other cEMI frame, and the `M_PropInfo.ind`
  frames a server sends unprompted (e.g. the evented `PID_KNXNETIP_DEVICE_STATE`) go
  to an `indication_callback`. Not an `Interface`: it carries no telegrams and leaves
  `xknx.current_address` and the connection manager alone.
- **`DeviceManagement`** — the receiving half on its own: acknowledges
  `DeviceConfigurationRequest` frames arriving on a caller-owned transport and tracks
  their sequence counters, as the counterpart to the existing `DeviceConfiguration`.

### Spec conformance

- A sent request waits `DEVICE_CONFIGURATION_REQUEST_TIMEOUT` (10 s) for its ACK, is
  repeated 3 times, then the connection is terminated with a `DisconnectRequest`
  (03.08.03 §2.3.2). `DeviceConfiguration` now defaults to this timeout instead of the
  1 s inherited from `RequestResponse`.
- A failed heartbeat is repeated 3 times, then the connection is terminated
  (Core 03.08.02 §5.4). A `DisconnectRequest` sent by the server is answered with a
  `DisconnectResponse` and ends the connection (§5.5).
- Over TCP no acknowledgements are sent, received ones are ignored, and sequence
  counters are not evaluated (03.08.03 §2.3.2 "On TCP", Core §8.4.3.4.1); a lost TCP
  connection ends the device management connection. The Secure variant runs the same
  logic over a `SecureSession`.
- Sequence counters are independent per direction and reset on connect (Core §5.3.4).
  Received expected-1 repetitions are re-acknowledged and discarded — 03.08.03 only
  states the server side, so the client follows the rules Tunnelling spells out for
  both ends (see the comment in `DeviceManagement`).

### Robustness

Answers are correlated to the requesting property, so the late answer to a timed-out
request cannot resolve a later one. `disconnect()` fails a pending request immediately
instead of letting it time out. Unparseable cEMI payloads and exceptions from user
callbacks are contained — both run inside the transport's datagram callback. An error
octet outside `CEMIErrorCode` raises `CommunicationError` instead of leaking
`ValueError`.

### Notes

- `request()` is effectively limited to the Property services for now: xknx doesn't
  parse the `M_FuncProp*` message codes yet, and `M_Reset.req` has no confirmation at
  all (EMI_IMI §4.1.7.5.2 note 1), so nothing consumes their answers.
- Resolving `CEMIErrorCode` at parse time instead of lazily in
  `CEMIMPropReadResponse.error_code` would simplify the error handling here — worth a
  separate PR.

🤖 Generated with Claude Code

## Type of change

## Checklist


farmio (commented):
Claude Code  🤖:

Checked out and verified locally: 48 tests pass, mypy/ruff/pylint clean. The structure mirrors `tunnel.py` (`_Tunnel`/`UDPTunnel`/`TCPTunnel`/`SecureTunnel`) closely enough that it reads as one family — nice.

## Spec conformance

I checked the claims in the PR description against the v3.0.2 PDFs, and they hold:

- **03.08.03 §2.3.2** (p. 7): 10 s `DEVICE_CONFIGURATION_REQUEST_TIMEOUT`, *"repeat the DEVICE_CONFIGURATION_REQUEST three (3) times and then terminate the connection by sending a DISCONNECT_REQUEST"*. `range(REPETITIONS + 1)` = 4 transmissions ✔, the repetition keeps the sequence counter ✔, and the spec's *"or the status of a received DEVICE_CONFIGURATION_ACK frame signals any kind of error condition"* is covered because `DeviceConfiguration.success` is already false on a non-`E_NO_ERROR` ack ✔.
- **Core §8.4.3.4.1** reads verbatim *"For connections inside a TCP connection, the sequence counter shall not be evaluated by the receiver"* — matches `TCPDeviceManagementConnection` ✔.
- The `DeviceManagement` sequence rules match **Tunnelling §2.6** client-side wording exactly (expected → ack + process; expected−1 → ack + discard; anything else → no reply, discard). The comment justifying why the client can't take the Device Management server sentence literally is correct.
- Operator precedence in `x + 1 & 0xFF` / `x - 1 & 0xFF` is right.

Also verified by probe: the *ack lost but answer arrived* branch in `_send_request` works and advances the sequence counter; the `disconnect()` ordering (`_stop()` cancels `_pending` before `communication_channel = None`, with no `await` in between) is sound; and the `CancelledError` discrimination in `request()` correctly separates `_stop()` from external task cancellation.

## One real finding

**`write_property()` blocks for 10 s and reports the wrong cause when the server returns an undefined error octet.**

`CEMIMPropReadResponse.error_code` is lazy, so `read_property` can catch the `ValueError` and report it accurately. `CEMIMPropWriteResponse.from_knx` resolves `CEMIErrorCode(raw[6])` **eagerly**, so the same server response never becomes a frame: the `ValueError` is caught by `_cemi_received`'s `except (CouldNotParseCEMI, UnsupportedCEMIMessage, ValueError)`, the frame is dropped with a warning, and the pending request runs into the answer timeout.

Confirmed with a probe (`M_PropWrite.con`, `number_of_elements=0`, error octet `0xF7`):

```
CommunicationError: No answer to CEMIMessageCode.M_PROP_WRITE_REQ within 10 seconds.
```

So a server that answered promptly is diagnosed as unresponsive. The PR description mentions the lazy/eager inconsistency as a cosmetic follow-up — it's actually a behavioural asymmetry on the write path. Since `write_property`'s `assert isinstance(...)` is unreachable for this case anyway, the fix is the one you already propose, just pulled into this PR: resolve the error octet uniformly (and tolerantly) at parse time.

## Minor

- **`CEMIMPropReadResponse.__repr__` raises `ValueError`** for an undefined error octet (`f'error_code="{self.error_code}"'`). Pre-existing, but this PR adds the two `logger.debug(..., cemi)` calls that route such a frame into a repr (`_cemi_received` "unexpected cEMI frame", `request()` "not answering the request"). Lazy `%s` formatting means it only bites with debug logging on, and `logging` swallows it into `--- Logging error ---`. Same root cause as above — the same fix removes it.
- **`request()` reuses `DEVICE_CONFIGURATION_REQUEST_TIMEOUT` for the *answer* wait.** The spec's 10 s is the ACK timeout only; this is a sensible choice but not a spec requirement, and sharing the constant makes it look like one. Worth a one-line comment. Consequence: worst-case UDP `read_property` is 4×10 s (ack) + 10 s (answer) ≈ 50 s, with no per-call override.
- **`connect()` leaks `communication_channel` on a late failure.** It is assigned in `_connect_request()` before `_start_receiving()` / `register_callback`; if either raises, the `except (OSError, CommunicationError)` stops the transport but leaves the channel set, so the next `connect()` raises "already open" on a dead connection. Narrow, but `self.communication_channel = None` in that handler is free.
- **The correlation guarantee is slightly weaker than the description states.** "The late answer to a timed-out request cannot resolve a later one" holds across *different* properties; `_same_property` will happily match a retry of the same read. Harmless in practice (same property, same value), but the claim reads stronger than the code.
- **`DeviceManagement` diverges from `UDPTunnel` on out-of-sequence frames** — the tunnel schedules a disconnect (`_invalid_sequence_number_reconnect_schedule`), this only logs. Defensible since the class doesn't own the connection, and the comment says the server will repeat then drop. But the tunnel's shortcut exists precisely because *"some interfaces don't initiate a disconnect here"* — if those same interfaces misbehave on a device management connection, this hangs instead. Worth a deliberate decision either way.
- `TCPDeviceManagementConnection` sets `local_hpai` in both `__init__` and `_setup_connection()`. Redundant, but `TCPTunnel` does the same, so it is at least consistent.

Test coverage is genuinely thorough — heartbeat recovery/shutdown, route-back, external cancellation, the defensive guards. The one gap that matters is the write-side error-code case above.

pdgendt:
I'll update and test on Monday 🙂 

EDIT: Done

kewde (approved):

kewde:
Mercikes @pdgendt!

AGENTS.md
```diff
@@ -48,9 +48,12 @@ if they aren't already established elsewhere in the codebase.
...
 | 03.03.04 | Transport Layer | v01.02.02 | `xknx/telegram/tpci.py` |
 | 03.07.02 | Datapoint Types | v02.02.01 | `docs/changelog.md` (historical entry) |
-| 10.1 | Logical Tag Extended | v01.02.02 | `xknx/telegram/apci.py` |
-| 3.8.5 | Routing | v01.05.02 | `xknx/io/routing.py` |
-| 3.2.6 | Communication Medium KNX IP | v01.01.02 | `xknx/io/routing.py` |
+| 10.01 | Logical Tag Extended | v01.02.02 | `xknx/telegram/apci.py` |
+| 03.08.05 | Routing | v01.05.02 | `xknx/io/routing.py` |
+| 03.02.06 | Communication Medium KNX IP | v01.01.02 | `xknx/io/routing.py` |
+| 03.08.02 | Core | v01.06.02 | `xknx/io/device_management_connection.py` |
+| 03.08.03 | Device Management | v01.07.03 | `xknx/io/device_management.py`, `xknx/io/device_management_connection.py`, `docs/changelog.md` |
```
pdgendt, line 55: changelog.md?

kewde: I'll clean up the AGENTS.md in a follow up PR - the references to these files are somewhat useless anyways. Mostly trying to tell the LLMs to reference the specifications in a consistent manner.

PR #1900, opened 2026-08-07
Don't swallow the cause of CEMI and APDU parse errors

farmio:
### Description

`CouldNotParseCEMI` and `UnsupportedCEMIMessage` are logged with `%s`, so the exception chain established by `raise ... from err` never reaches the log - only the outermost, most generic message does. Reported from a Home Assistant log:

```
WARNING (MainThread) [xknx.cemi] CEMI Frame failed to parse: <CouldNotParseCEMI description="APDU invalid from 1.0.252 to 0/0/0 with TPCI TDataBroadcast() in CEMI: 30e010fc00001803f192003f1414e8e5000a46492919b3498640a7655948919e" />
```

Two layers each dropped their cause, so acting on a report like this means re-parsing the hex by hand in a REPL.

- `APCI.from_knx` now repeats the `IndexError` / `struct.error` / `ValueError` it wraps in its `ConversionError` message. These are raised deep inside the service parsers (slicing, `struct.unpack`, enum lookups) and carry the only hint about what exactly was wrong with the payload.
- `CEMILData.from_knx` includes the description of the wrapped `ConversionError` / `UnsupportedAPCIService` / TPCI error. The "APDU not supported" case gains the most - it used to drop the name of the offending service entirely, even though that is the whole point of the message.

The same frame now logs as

```
<CouldNotParseCEMI description="APDU invalid: Error parsing APCI 0b1111110001 from raw data: 03f192003f1414e8e5000a46492919b3498640a7655948919e: ValueError('2 is not a valid SecurityALService') - from 1.0.252 to 0/0/0 with TPCI TDataBroadcast() in CEMI: 30e010fc00001803f192003f1414e8e5000a46492919b3498640a7655948919e" />
```

The APDU hex appears twice - once on its own, once as part of the cEMI frame. `APCI.from_knx` is also used standalone, so it keeps its own context.

Note that the frame from that report is *not* actually invalid - it is a Data Secure S-A_Sync_Req with a mis-decoded S-A-Service. That is fixed separately in `fix-data-secure-scf-sync-req`; the tests here use SCF `0x94` (S-A-Service `0b100`, reserved) instead, so the two changes are independent.

### Checklist

- [x] Tests added/updated
- [x] `docs/changelog.md` updated

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,10 @@ nav_order: 2
 
 - Fix the encoding of `SecurityALService.S_A_SYNC_REQ`: the S-A-Service field of the Security Control Field encodes an S-A_Sync_Req-PDU as `0b010`, not `0b001` - KNX v02.01.01 - Application Layer 03.03.07 - §5.1.1 Table 5. Incoming Data Secure sync requests were rejected with `CouldNotParseCEMI` "APDU invalid" instead of being parsed and then ignored as an unsupported S-AL service.
 
+### Internals
+
+- Keep the underlying error message when a CEMI or APDU parse error is re-raised as a more generic exception so it no longer hides what was actually wrong with the frame.
+
 # 3.18.0 Protocol Droid 2026-08-02
 
 ### Bugfixes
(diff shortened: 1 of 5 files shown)


kewde (approved):
LGTM - I wonder if there is a way to have the logger also contain the cause text (of `from err`) without having to re-include the message again.

farmio:
`logger.exception()` would log the whole stack trace. But this seems a bit verbose.

PR #1904, opened 2026-08-12
Extract shared data connection mechanics for tunnelling and device management

pdgendt:
## Description

Refactor only — no functional change intended (two narrow exceptions listed below).

Tunnelling and device management connections each carried their own copy of the
mechanics KNXnet/IP Core 03.08.02 §5 "Communication Channels" defines for every
data connection. This extracts them into `xknx.io.data_connection`, shared by both:

- **`ConnectionHeartbeat`** — heartbeat monitoring (§5.4): send a
  `ConnectionStateRequest` every `HEARTBEAT_RATE` seconds, repeat a failed one
  three times, then log and report the failure once. Owns its worker task
  (`start()`/`stop()`, safe to stop from within the failure callback).
- **`IncomingSequenceCounter`** — evaluates the sequence counter (§5.3.4) of a
  received frame ("evaluate" per §8.4.3.4.1) and returns a `SequenceVerdict`:
  `EXPECTED` (acknowledge and process), `REPEATED` (acknowledge again, discard)
  or `OUT_OF_ORDER` (discard without acknowledgement — the `E_SEQUENCE_NUMBER`
  wording, §7.3.4), with 8-bit wraparound. Rules per Tunnelling 03.08.04 §2.6.1,
  which Device Management clients follow as well.

Both components implement the *mechanics* only; the failure *policy* stays with
the owner — which is where the two connection types genuinely differ, and why
this is composition rather than a shared base class: a tunnel reconnects
(`_tunnel_lost()`) and schedules a disconnect for out-of-order frames, a device
management connection closes and only observes.

## Adoption

- `DeviceManagement` and `_DeviceManagementConnection` use the components in
  place of their inlined copies (net deletion).
- `_Tunnel`'s `do_heartbeat`/`_do_heartbeat_failed` are replaced;
  `start_heartbeat()`/`stop_heartbeat()` delegate. `UDPTunnel` evaluates
  incoming `TunnellingRequest` counters through the component, keeping its
  reconnect-schedule shortcut for out-of-order frames.
- Public API is unchanged: `DeviceManagement.expected_sequence_number` and
  `UDPTunnel.expected_sequence_number` remain as delegating properties.

## Behaviour

The tunnel test suite passes **unmodified**. Two subtle differences come with
the component, both hardening previously untested edges:

1. The old tunnel heartbeat loop did not end on failure — it kept running until
   the reconnect cancelled it, and with `auto_reconnect=False` it would warn and
   re-trigger `_tunnel_lost()` every `HEARTBEAT_RATE` forever. The component
   reports the failure once and ends; a successful reconnect starts a fresh one.
2. The old `start_heartbeat()` overwrote the task reference without cancelling,
   so a double start leaked a running heartbeat task. The component stops a
   previous heartbeat first.

## Tests

The components get direct unit tests (verdicts, wraparound — including the
255-is-a-repetition-of-expected-0 edge, heartbeat retry/recovery/failure,
`CommunicationError` handling, stop-from-failure-callback, restart). The
existing tunnel and device management tests cover the adoption; only two device
management tests changed, both because they poked the private heartbeat task
attribute.

## Notes

- Naming was checked against the spec PDFs: "data connection" (Core Figure 5,
  §5.3.1, §5.3.4), "evaluate" (§8.4.3.4.1), "out of order" (§7.3.4), "expected" /
  "repeated" (Tunnelling §2.6.1). `SequenceVerdict` has no spec term — the spec
  never names the three-way outcome.
- Deliberately **not** included: backporting the device management connection's
  communication-channel filter to `UDPTunnel`, and ACK correlation
  (channel/sequence) in `RequestResponse` — both are behaviour changes that now
  have a single place to land, but they belong in a separate PR so this one
  stays a pure refactor.

## Type of change

## Checklist


xknx/io/tunnel.py
```diff
@@ -82,7 +86,10 @@ def __init__(
         self.sequence_number = 0
```
kewde, line 86: Tunnel & data management connection keep a copy of `sequence_number` - and another copy of the sequence number seems to be stored in `IncomingSequenceCounter`. I wonder if it makes sense to make a single source of truth for the sequence numbers?

https://github.com/XKNX/xknx/pull/1904/changes#diff-d3a682d26b70727ccf3a892fce4c45e2789b0e65c76da3388e9cdbf1a32d2a77R65

farmio:
Claude Code  🤖:

Reviewed the diff, ran the full suite (3826 passed) and ruff/mypy (clean). The refactor is sound — the sequence-counter semantics are preserved exactly (`(expected ± 1) & 0xFF` precedence is right, wraparound covered), and the composition-over-base-class argument holds up. Findings, most important first:

## 1. Spec citations in the new module don't follow AGENTS.md's mandated format

`AGENTS.md` requires `KNX v<version> - <Document Title> <document number> - §<paragraph>`. The new `xknx/io/data_connection.py` uses none of it:

| in the new file | required form |
|---|---|
| `KNXnet/IP Core 03.08.02 §5.4` | `KNX v01.06.02 - Core 03.08.02 - §5.4` |
| `KNXnet/IP Tunnelling 03.08.04 §2.6.1` | `KNX v01.07.01 - Tunnelling 03.08.04 - §2.6.1` |
| `Device Management 03.08.03 §2.3.2` | `KNX v01.07.03 - Device Management 03.08.03 - §2.3.2` |

This is a regression rather than a fresh omission: the comment block deleted from `device_management.py` cited both documents *in the correct form*, and the move stripped the versions. All three versions are in AGENTS.md's confirmed-versions table, so no guessing is needed. The changelog entry in this same PR uses the correct format, so the PR is internally inconsistent too. Six sites: module docstring, `ConnectionHeartbeat` docstring, the inline §5.4 comment, `SequenceVerdict.OUT_OF_ORDER`'s `§7.3.4`, and two in the `IncomingSequenceCounter` docstring.

## 2. The heartbeat log line no longer says *which* connection failed

Both callers now emit `Connection heartbeat failed …`. Previously they were distinguishable:

- tunnel: `Heartbeat failed. Tunnelling server repeatedly responded with status: X`
- device management: `Device management connection heartbeat failed …. Disconnecting.`

Since 3.20 makes a tunnel and a device management connection in one process the normal case, a user report of "connection heartbeat failed" is now ambiguous. Cheapest fix is a `name: str` ctor argument folded into the message; alternatively drop the `logger.warning` from the component and let each owner log inside `on_failure` — which is where the policy already lives.

## 3. Undocumented behaviour change: heartbeat on a cleared channel

`_DeviceManagementConnection._connectionstate_request()` now raises `CommunicationError` when `communication_channel is None`, where the old heartbeat did `if (channel := …) is None: return` and ended silently. The new path logs a warning and calls `disconnect()` (which then does a redundant `transport.stop()`). Every production path that clears the channel also stops the heartbeat, so this is unreachable in practice — but it turns a defensive no-op into a user-visible warning, and it isn't in the PR description's "Behaviour" section alongside the two changes that are.

## 4. A test assertion was weakened without need

`test/io_tests/device_management_connection_test.py` replaced `send_mock.assert_not_called()` with a loop asserting merely that no `ConnectionStateRequest` was among the sends. I put the strict assertion back and ran it — it still passes, because nothing at all is sent on that path. Worth reverting; as written the test would no longer notice if the new failure path started emitting frames.

## Minor

- `test_stop_from_on_failure` reaches in and reassigns the private `_on_failure` after construction. A closure over a one-element list, or `AsyncMock(side_effect=…)` assigned before `start()`, would test the same thing through the public surface.
- `await self._on_failure()` sits outside the `try` in `_run`, so an exception from the owner's callback escapes as an unretrieved task exception. Both old implementations had the same exposure, so this is not a regression — just the one rough edge left in an otherwise well-encapsulated component.
- The `expected_sequence_number` setters on `DeviceManagement` and `UDPTunnel` have no production writers (tests only). Keeping them is defensible since the attribute was public before; just noting they're compatibility shims, not live API.

**Verdict:** the refactor itself is correct and I'd take it. Item 1 should be fixed before merge (it's a documented repo rule and the information was already in the deleted code); items 2–4 are cheap and worth doing in the same pass.

xknx/io/device_management.py
```diff
@@ -61,16 +62,26 @@ def __init__(
...
 
+    @property
+    def expected_sequence_number(self) -> int:
+        """Return the sequence counter expected on the next request."""
+        return self._sequence.expected
+
+    @expected_sequence_number.setter
+    def expected_sequence_number(self, value: int) -> None:
+        """Set the sequence counter expected on the next request."""
+        self._sequence.expected = value
```
farmio: are those only for unit-test purposes? I think we could just reach into `_sequence.expected` then.

xknx/io/tunnel.py
farmio: That's outgoing sequence number. It's separate from incoming sequence number.

xknx/io/device_management.py
farmio: same for `UDPTunnel`, but it's used for logging purposes too there - which could also just reach into `self._sequence`. Guess the setter is only for unit-tests there too.

xknx/io/device_management_connection.py
```diff
@@ -285,10 +286,11 @@ def _disconnect_request_received(
         )
         self._connection_lost()
 
-    async def _connectionstate_request(self) -> tuple[bool, str | None]:
+    async def _connectionstate_request(self) -> tuple[bool, str | None] | None:
         """Send a ConnectionStateRequest and return its outcome. Heartbeat callback."""
         if self.communication_channel is None:
-            raise CommunicationError("No active device management connection.")
+            # The connection is already gone - end the heartbeat quietly.
+            return None
```
farmio, line 293: We have the same for `_Tunnel`. Do we want to change this there too?

pdgendt: Updated, but it is a small behavior change.

xknx/io/tunnel.py
kewde: Would it make sense to create a `SequenceCounter` class to track both in a single place? Can add that in a follow up PR if it makes sense

test/io_tests/tunnel_test.py
```diff
@@ -611,6 +611,51 @@ async def test_tunnel_heartbeat_no_answer(
...
+        heartbeat_task = self.tunnel._heartbeat._task
+        assert heartbeat_task is not None
+
+        self.tunnel.communication_channel = None
+        await time_travel(HEARTBEAT_RATE)
+
+        # no ConnectionStateRequest for a dead channel, no _tunnel_lost
+        mock_transport_send.assert_not_called()
+        mock_tunnel_lost.assert_not_called()
+        assert heartbeat_task.done()
```
farmio: this test seems very synthetic - `tunnel.communication_channel` is only set to `None` when a disconnect was sent or received. In both cases `self.stop_heartbeat()` (latter through `self._tunnel_lost`) is called. 
Is the `communication_channel = None` branch realistically unreachable code?

farmio (approved):
Very nice! Thank you 👍

PR #1908, opened 2026-08-16
Make Telegram generic over its payload's APCI type

farmio:
## Summary
- Make `Telegram` generic over `payload` (`Telegram[GroupValueWrite]`, etc.), defaulting to today's behavior when left unparametrized. Parametrized as `Telegram[SomeAPCI]`, `payload` is `SomeAPCI` - never `None` - while the default/unparametrized `Telegram` keeps `payload: APCI | None` for control telegrams (ACK, Disconnect, ...) that carry none.
- Add `ExpectedResponse`, a mixin marking that an APCI request service (`DeviceDescriptorRead`, `AuthorizeRequest`, `PropertyValueRead`) has a KNX-spec-defined response type, set once on the class as `RESPONSE_TYPE`. `P2PConnection.request()` no longer takes an `expected=` argument at all - it infers and verifies the expected response from `type(payload).RESPONSE_TYPE` and returns the correspondingly typed `Telegram[ResponseType]`.
- `Device.process()` now hands `process_group_write()`/`process_group_response()`/`process_group_read()` a `Telegram` narrowed to the APCI type it already verified via `isinstance`, propagated through `RemoteValue.process()` and every device's `process_group_*` override.
- No behavior change: `RemoteValue.process()` keeps its own `isinstance` check, since - unlike the management case - nothing enforces the payload type before it's called directly (a test constructs a telegram with an arbitrary payload to exercise exactly that).

## Test plan
- [x] `mypy xknx` (matches the project's pre-commit hook scope) - clean
- [x] `ruff check` / `ruff format --check` - clean
- [x] `pylint xknx examples` - clean
- [x] `pytest test/` - 3839 passed


xknx/management/management.py
```diff
@@ -392,8 +394,17 @@ async def _receive(self, expected_payload: type[APCI] | None) -> Telegram:
...
+    async def request(self, payload: ExpectedResponse[APCIT]) -> Telegram[APCIT]: ...
+    @overload
+    async def request(self, payload: APCI) -> Telegram: ...
+    async def request(self, payload: Any) -> Telegram[Any]:
+        """
+        Send a payload to the KNX device and wait for the response.
+
+        The response is verified against `type(payload).RESPONSE_TYPE` if the
+        payload defines one - see `ExpectedResponse`.
+        """
```
farmio, line 407: @kewde do we need a custom `expected` payload type attribute? Or is it enough to only define expected responses in the request APCI?

pyproject.toml
```diff
@@ -18,7 +18,7 @@ classifiers = [
 ]
 dependencies = [
   "async_timeout>=4.0.0; python_version<'3.11'",
-  "typing_extensions; python_version<'3.11'",
+  "typing_extensions",
```
kewde: Intended?

xknx/management/management.py
kewde: You mean as an overload? I think "Any" is a good enough escape hatch from the type system, giving little use to "expected" from the typing side.

But "expected" is currently doing a little more than that, should we assert that a payload has a response class is set as well? What's the intended behavior when `expected` is None?

Also does the KNX specs define a singular response class per request, or can a single request be answered with different response classes (e.g an error class is a valid response and the type system shouldnt fight us on that?)

I'll look through the specs and see if I can find any such cases. A 1 to 1 mapping would be awesome.

Alternatively, I wonder if the response class can be looked up based on apci bits of the header. So it always parses responses correctly, but a mechanism to verify that the response matches our expectation at runtime should be kept imho.

pyproject.toml
farmio: Yes, `TypeVar(..., default=...)` was introduced in 3.13
So we couldn't use `tg: Telegram` but needed `tg: Telegram[APCI | None]` under mypy strict rules (`disallow_any_generics`) wherever a telegram type was used.

xknx/management/management.py
farmio: not as an overload, but the additional argument - like it was before. Like, if `None` (or some sentinel constant), it's ignored, if set it was used rather than the APCIs `RESPONSE_TYPE`.

> I wonder if the response class can be looked up based on apci bits of the header

I guess that won't work with static type checking.

xknx/telegram/apci.py
```diff
@@ -449,6 +457,57 @@ def from_knx(cls, raw: bytes) -> APCI:
...
+                continue
+            response_type: type[APCIResponseT] = (
+                getattr(sys.modules[owner.__module__], args[0].__forward_arg__)
+                if isinstance(args[0], ForwardRef)
+                else args[0]
+            )
+            return response_type
+        raise AttributeError(
+            f"{owner.__name__} does not parametrize APCIRequest with a response type"
+        )
```
farmio: This is a bit harsh...

If we reordered every APCI subclass to 
- write
- response
- read

we could drop this - and drop the forward reference strings. So it would look like
`class ADCRead(APCIRequest[ADCResponse]):`
instead of
`class ADCRead(APCIRequest["ADCResponse"]):`
since `ADCResponse` would already be defined before it was used in `ADCRead` class header then.

... but we needed to keep that order, otherwise we'd get typing errors, and more importantly, import-time `NameError`s when typing errors were ignored. 

Any preferences @kewde ?

kewde: Re-order is cleaner 👍

uv.lock
kewde, line 1: intended?

farmio: no. amended the commit

kewde (approved):

PR #1911, opened 2026-08-18
Discard unencrypted frames on KNX IP Secure transports

farmio:
## Description

Resolves the three `# TODO: disallow unencrypted frames ...` comments in `xknx/io/ip_secure.py`.

`SecureSession.handle_knxipframe()` passed **every** non-`SecureWrapper` frame straight to the callbacks. Since `RequestResponse` and `_Tunnel` register their callbacks on the transport, an on-path attacker on the TCP connection could inject a plain `TunnellingRequest` (telegram injection), a plain `DisconnectRequest` (forced disconnect) or a plain `ConnectResponse`, and it was acted on unauthenticated. `SecureGroup` dropped plain `RoutingIndication` but still forwarded plain `RoutingBusy` into `Routing._flow_control.handle_routing_busy()`, so any sender on the multicast group could throttle or stall our transmissions.

### What the spec allows in plain

All references: `KNX v01.01.02 - KNX IP Secure 03.08.09`.

**Secure session** — exactly two frames, and only before the session is initialized:

- §2.2.3.6.1 / §2.2.3.7 — `SessionRequest` out, `SessionResponse` in.
- §2.2.3.5.2.3 — the session state machine knows only two wire events: *Received SESSION_REQUEST* and *Received valid SECURE_WRAPPER*.
- §2.4.1 — "For all connections created within a secure session, the server shall accept only SECURE_WRAPPER frames […] and plain frames shall be ignored."

The old TODO mentioned "exceptions for discovery etc. eg. DescriptionResponse", but that turns out **not** to be an exception for *plain* frames: §2.2.1.4.2 exempts discovery from *authorization*, not from encryption, and §2.4.1 puts connectionless services like `DESCRIPTION_REQUEST` "within the context of the session" — still wrapped. That matches what xknx already does: `_Tunnel.request_description()` runs over the established session, and pre-connection discovery uses its own plain `UDPTransport`.

**Secure routing** — `TimerNotify` (carries its own MAC, §2.2.2.4.3) plus discovery and self description, which are never secured (§2.6.2, §2.2.1.4.2) and share the endpoint: the routing multicast group is by default the discovery group, and `create_multicast_sock()` binds the listener to `INADDR_ANY:3671` on macOS and Windows, so unicast `DescriptionRequest`s addressed to us arrive there too. Everything else plain is discarded — which now also covers the rest of the routing service family (§2.2.1.4.5 speaks of the *service family*, not just `Routing.ind`), `TimerNotify` on a session, and `SESSION_*` on the multicast group.

**Never encapsulated at all** — rejected after decryption, in `decrypt_frame()` so both transports share the rule:

- a nested `SecureWrapper` (§2.2.1.2.2.2, §2.2.3.5.1);
- the Remote Configuration and Diagnosis family, which §2.2.1.4.7 forbids outright in Security Mode and which `03.08.07` transmits by multicast — straight onto the `SecureGroup` socket. Those four service types have no body class yet, so the guard is what keeps the rule true once they get one.

### Summary of behaviour changes

| | Before | After |
|---|---|---|
| `SecureSession.handle_knxipframe` | every plain frame dispatched | only pre-init `SessionResponse`; rest discarded with a warning |
| `SecureSession.send` | anything sent plain pre-init | raises `IPSecureError` for anything but `SessionRequest` |
| `SecureGroup.handle_knxipframe` | only plain `RoutingIndication` dropped | allow-list of `SEARCH_*` + `DESCRIPTION_*`; rest discarded |
| `decrypt_frame` | any decrypted body accepted | rejects nested `SecureWrapper` and the Remote Config/Diagnosis family |
| `SecureGroup.send` | stale TODO | removed — it always encrypts |

### Notes for the reviewer

- `test_lifecycle` fed a plain `SessionStatus(STATUS_CLOSE)` marked *"unencrypted for sake of simplicity"*. §2.2.3.9 says that frame "shall only be sent wrapped in a SECURE_WRAPPER frame", so the test now wraps it via `encrypt_frame()` — the shortcut had been exercising a path the spec forbids.
- Sending raises rather than dropping: reaching that branch can only be an internal bug, never remote input, matching the `IPSecureError` already raised on sequence-number overflow. `IPSecureError` is a `CommunicationError`, which `RequestResponse.start()` already handles.
- Receiving discards rather than raises, per §2.4.1 "shall be ignored" — a raise would escape `TCPTransport.data_received_callback()`, which only guards `KNXIPFrame.from_knx()`.
- `AGENTS.md` gains the confirmed `KNX IP Secure | 03.08.09 | v01.01.02` row (taken from the document's own title page), and the one `AN159 v06`-style citation in `ip_secure.py` is converted to the documented format.

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/xknx/io/ip_secure.py
+++ b/xknx/io/ip_secure.py
@@ -24,8 +24,8 @@ from xknx.knxip import (
     HPAI,
     KNXIPFrame,
     KNXIPServiceType,
-    RoutingIndication,
     SecureWrapper,
+    SessionRequest,
     SessionResponse,
     SessionStatus,
     TimerNotify,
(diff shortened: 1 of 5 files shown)


xknx/io/ip_secure.py
```diff
@@ -490,15 +523,19 @@ def handle_knxipframe(self, knxipframe: KNXIPFrame, source: HPAI) -> None:
...
             knx_logger.debug("Decrypted frame: %s", knxipframe)
-        # handle decrypted frames and plain frames (e.g. SearchRequest for discovery)
+        elif knxipframe.header.service_type_ident not in PLAIN_MULTICAST_SERVICES:
+            ip_secure_logger.info(
+                "Discarding received unencrypted frame: %s",
+                knxipframe,
+            )
+            return
+        # handle decrypted frames and plain discovery frames (e.g. SearchRequest)
         super().handle_knxipframe(knxipframe, source)
```
kewde:
```suggestion
        elif knxipframe.header.service_type_ident in PLAIN_MULTICAST_SERVICES:
           super().handle_knxipframe(knxipframe, source)
         else:
            ip_secure_logger.info(
                "Discarding received unencrypted frame: %s",
                knxipframe,
            )
            return
        # handle decrypted frames and plain discovery frames (e.g. SearchRequest)
```

maybe more clear? dont merge this suggestion, probably f'd up lint etc

farmio: 👍 Done

kewde (approved):
LGTM, nice 👍

PR #1914, opened 2026-08-20
Manage dependencies with uv and run hooks with prek

farmio:
## Description

Preparation for 4.0: `pyproject.toml` becomes the only place dependencies are declared, and `uv.lock` the only place versions are pinned.

Today they live in four places, kept in sync by hand and by two different bots:

| Where | What |
| --- | --- |
| `pyproject.toml` `[project].dependencies` | the library's ranges — correct, published |
| `requirements/production.txt` | the *same* runtime deps as exact pins; nothing installs it on its own |
| `requirements/testing.txt` | the dev tools, exact pins — consumed by `tox.ini`, README and both workflows |
| `.pre-commit-config.yaml` | ruff pinned a **second** time, already drifted: `ruff==0.16.0` vs `rev: v0.16.3` |

Nothing pinned transitive dependencies, so CI was not reproducible, and every job installed the dev deps twice — once for `tox`/`pre-commit`, then again inside each tox env.

### What changed

- **Dev tooling moves to `[dependency-groups] dev`** in `pyproject.toml` as ranges; `uv.lock` (committed) carries the exact versions including transitives, resolved universally across 3.10–3.14 × Linux/macOS/Windows. `requirements/` is gone.
- **Dependabot moves from the `pip` to the `uv` ecosystem.** This is the load-bearing part: its `pip` ecosystem does not reliably update PEP 735 `[dependency-groups]` ([dependabot-core#12788](https://github.com/dependabot/dependabot-core/issues/12788)), while the `uv` ecosystem updates `uv.lock` ([GA since March 2025](https://github.blog/changelog/2025-03-13-dependabot-version-updates-now-support-uv-in-general-availability/)). Moving to `[dependency-groups]` while staying on pip would have silently killed the only automation keeping our pins fresh.
- **tox is removed.** CI drives the version matrix directly via `astral-sh/setup-uv`, and the four `ruff,typing,lint,pylint` envs collapse into a single `lint` job. `uv sync --locked` doubles as a guard that `uv.lock` matches `pyproject.toml`.
- **pre-commit → [prek](https://github.com/j178/prek)**, from the same `.pre-commit-config.yaml`. It ships as a platform wheel on PyPI, so it sits in the `dev` group and is locked and Dependabot-bumped like everything else — no separate installer, no extra version source. Precedent: CPython, FastAPI, Airflow and [Home Assistant](https://developers.home-assistant.io/blog/2026/01/13/replace-pre-commit-with-prek/). Reverting means putting `pre-commit` back in the dev group; the config format is unchanged.
- **ruff, ruff format, mypy and pylint are local hooks** run through `uv run --frozen`, so their versions come from `uv.lock` alone. That removes the second ruff pin, and with it `script/run-in-env.sh`, which existed only to guess which virtualenv the mypy hook should use.
- `check-json` dropped — the repository tracks no JSON files, so it has always been a no-op.

### Contributor impact

The local setup changes; `pip install -r requirements/testing.txt` and `tox` no longer exist. README and AGENTS.md are updated:

```
uv sync                        # create/update the environment
uv run pytest                  # unit tests
uv run prek run --all-files    # ruff, ruff format, mypy, pylint, codespell
uv run --python 3.10 pytest    # test against another supported version
```

### Verified locally

- `uv lock --check` and `uv sync --locked` clean
- `uv run pytest` — 3842 passed; Python 3.10 (3841 passed, 1 skipped) and 3.14 (3842 passed) both green
- `prek run --all-files` — all hooks pass, 21 s wall at 512% CPU (was a serial tox run); `--hook-stage manual` and `prek install` exercised
- `uv build` — wheel and sdist still carry `py.typed`, LICENSE and README; **published `Requires-Dist` metadata is unchanged**, no `Provides-Extra` leakage

The macOS/Windows runners and the new `setup-uv` / `~/.cache/prek` caches only prove out on a real CI run — that is what this PR is for.

### Note on ruff

Collapsing the two pins lands on ruff 0.16.4. It produced no reformatting, so there is no formatting churn to review.

### Follow-up (not in this PR)

`pythonpublish.yml` still uses `python -m build` + `twine upload` with long-lived `PYPI_USERNAME`/`PYPI_PASSWORD` secrets. Worth switching to `uv build` + PyPI Trusted Publishing (OIDC) before 4.0 is cut; that needs a one-time trusted-publisher config on PyPI.

Fixes # (issue)

## Type of change

## Checklist


.github/workflows/ci.yml
```diff
@@ -9,86 +9,88 @@ on:
...
+    name: Lint, format and type check
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - name: Set up uv and Python
+        uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+          python-version: "3.14"
+      - name: Cache prek hook environments
```
kewde, line 60: curious, why do we need a git hook engine in CI?

.pre-commit-config.yaml
kewde: didn't we bump from 3.10 to (I forgot)?

farmio: just merged it this morning... this needs to be updated. Thanks

.github/workflows/ci.yml
farmio: we don't know if a user was running it... could have just not installed or `git commit --no-verify`  so we want to run the checks on CI too.
Also it's a single entry point to the linters - which seems nice.

kewde (approved):

PR #1915, opened 2026-08-20
Publish to PyPI with trusted publishing

farmio:
> [!NOTE]
> Stacked on #1914 (`uv-packaging`) — base it against `main` once that lands. The diff shown here is this PR's commit only.

## Description

Preparation for 4.0: the release workflow authenticates to PyPI with [Trusted Publishing](https://docs.pypi.org/trusted-publishers/) (OIDC) instead of a stored username and password.

`pythonpublish.yml` uploaded with `twine` using `PYPI_USERNAME` / `PYPI_PASSWORD` — long-lived credentials sitting in repository secrets, valid until someone rotates them, and usable by anything that can run a workflow here. It now builds with `uv build` and uploads with [`pypa/gh-action-pypi-publish`](https://github.com/pypa/gh-action-pypi-publish), which trades a short-lived GitHub OIDC token for an equally short-lived PyPI token, once per run. Nothing is stored, and PyPI accepts that token only from this repository's `pythonpublish.yml` running in the `pypi` environment.

### What changed

- **`python -m build` → `uv build`**, via `astral-sh/setup-uv@v7` on a pinned Python, same as [`ci.yml`](https://github.com/XKNX/xknx/blob/uv-packaging/.github/workflows/ci.yml). Same setuptools backend, same artifacts.
- **`twine upload` → `pypa/gh-action-pypi-publish@release/v1`.** No `TWINE_*` environment, no secrets in the workflow at all.
- **Build and publish are two jobs**, so `id-token: write` is scoped to the one that needs it. The `publish` job downloads the distributions as an artifact and neither checks out the repository nor runs any project code — nothing between build and upload can reach the OIDC token. Workflow-level permissions are `contents: read`.
- **Distributions get [PEP 740](https://peps.python.org/pep-0740/) attestations**, which the action generates by default when publishing this way.

### Before the 4.0 release is cut

Ordering matters — the release upload fails if step 1 is missing:

1. On PyPI, project `xknx` → *Publishing* → add a GitHub publisher: owner `XKNX`, repository `xknx`, workflow `pythonpublish.yml`, environment `pypi`. Needs a project owner. The environment name has to match the workflow, or the `environment:` block gets dropped here instead.
2. Optionally create the `pypi` environment under repository settings with required reviewers, for a manual gate on every upload. GitHub creates it implicitly otherwise.
3. After the first successful release, delete the `PYPI_USERNAME` and `PYPI_PASSWORD` secrets.

### Verified locally

- `uv build` produces `xknx-3.20.0.tar.gz` and `xknx-3.20.0-py3-none-any.whl` from the setuptools backend, leaving no stray `build/` behind
- workflow YAML parses; `prek` clean on both changed files

The publish path itself can only be proven by a release — the first 4.0 release candidate is the test.

## Type of change

## Checklist


kewde (approved):
LGTM 👍 

One thing I think might be worth adding is to pin the actions using the commit hash instead of versions (e.g "v7")? Reduces the supply chain risk a bit, a compromised version can be published over this tags which will automatically be used. But on the downside, you don't get instant fixes for existing vulnerabilities, choose your poison I guess. All the actions here seem to be from reputable sources, but if they were from some random person, then forking or pinning can be useful

PR #1916, opened 2026-08-21
Carry the expected response type through broadcast requests

farmio:
Follow-up to #1908. That PR made `P2PConnection.request()` return a `Telegram` narrowed to the response its payload declares, removing the `assert isinstance(response.payload, ...)` every caller needed. The broadcast path was left out, because there the `isinstance` is not redundant — nothing verifies the payload before the caller sees it, so it could not simply be deleted, only moved somewhere it is written once.

### `Broadcast.request()`

Sends a request service and yields the responses to it, holding a context open for the iteration:

```python
async for result in xknx.management.broadcast.request(
    apci.IndividualAddressSerialRead(serial=serial), timeout=timeout
):
    if result.payload.serial == serial:   # payload: IndividualAddressSerialResponse
        return result.source_address
```

Previously that read:

```python
async with xknx.management.broadcast() as bc_context:
    await xknx.management.send_broadcast(apci.IndividualAddressSerialRead(serial=serial))
    async for result in bc_context.receive(timeout=timeout):
        if (
            isinstance(result.payload, apci.IndividualAddressSerialResponse)
            and result.payload.serial == serial
        ):
            return result.source_address
```

Any number of devices may answer, so it yields until `timeout` elapses rather than returning a single telegram like `P2PConnection.request()`.

`BroadcastContext.receive()` now takes the APCI class to listen for and narrows to it; called without an argument it behaves as before. Several requests sharing one context still compose from `context()`, `send()` and `receive()` directly.

### `Broadcast`

Broadcast logic was spread over four members of `Management` plus a branch of its `process()`, while `BroadcastContext` had been reduced to a filtered queue — see [this thread](https://github.com/XKNX/xknx/pull/1916#discussion_r3830763545). A `Broadcast` class now owns the sending, the open contexts and the dispatching, reached as `xknx.management.broadcast`; `Management` keeps the point-to-point connections and routes broadcast telegrams to it.

The verb-noun method names collapse into plain verbs on the attribute:

| before | after |
| --- | --- |
| `management.send_broadcast(payload)` | `management.broadcast.send(payload)` |
| `async with management.broadcast()` | `async with management.broadcast.context()` |

This renames public API that exists on main. Whether to keep it is still open in that thread.

### `APCIBroadcastRequest`

The broadcast counterpart of `APCIRequest`, carrying its response type the same way. Both derive from a shared `_APCIWithResponse` base holding the `RESPONSE_TYPE` derivation. `IndividualAddressRead`, `IndividualAddressSerialRead`, `DomainAddressRead`, `DomainAddressSerialNumberRead`, `NetworkParameterRead` and `SystemNetworkParameterRead` derive from it — `GroupValueRead` and `GroupPropValueRead` stay plain `APCI`, since group communication never reaches a `BroadcastContext`.

It is a **sibling** of `APCIRequest`, not a subclass, so neither base accepts the other's services:

```python
conn.request(apci.IndividualAddressRead())                    # -> Telegram[APCI | None], not narrowed
xknx.management.broadcast.request(apci.MemoryRead(...))       # -> error: incompatible type
```

A broadcast service sent over a point-to-point connection just times out; inheriting would have made that call type check and promise a response that never arrives.

### Notes

- The six converted services have their `Response` class moved ahead of them, following the ordering established in #1908.
- Tests cover the filtering, `Broadcast.request()`, and that leaving the iteration early still releases the broadcast context. Each fails if the behaviour it covers is removed.
- `mypy`, `ruff`, `pylint` and the full suite (3897) pass.


kewde (approved):

xknx/management/management.py
```diff
@@ -157,22 +178,67 @@ async def broadcast(self) -> AsyncIterator[BroadcastContext]:
...
+    ) -> AsyncGenerator[Telegram[APCIResponseT], None]:
+        """
+        Broadcast a request service and yield the responses to it.
+
+        `send_broadcast()` and `receive()` in one - the response type to listen
+        for is the one `payload` declares. Any number of devices may answer, so
+        this yields until `timeout` elapses instead of returning a single
+        telegram like `P2PConnection.request()` does.
+        """
+        await self._management.send_broadcast(payload)
```
kewde: nit: move send broadcast logic into the broadcast context, so we don't need to pass down management

xknx/management/management.py
```diff
@@ -157,22 +178,67 @@ async def broadcast(self) -> AsyncIterator[BroadcastContext]:
 class BroadcastContext:
     """Class providing broadcast contexts."""
 
-    __slots__ = ("queue",)
+    __slots__ = ("_management", "queue")
 
-    def __init__(self) -> None:
+    def __init__(self, management: Management) -> None:
```
kewde: passing down management is a bit meh, clouds the ownership model imho

xknx/management/management.py
```diff
@@ -157,22 +178,67 @@ async def broadcast(self) -> AsyncIterator[BroadcastContext]:
 class BroadcastContext:
```
kewde, line 207: could make this an encapsulation of all broadcast related logic, kinda moving away from its original purpose now that this will land

xknx/management/management.py
farmio: its gone 😃

xknx/management/management.py
farmio: Claude Code  🤖: Prototyped this in a5aa4028.

`Broadcast` now owns the sending, the open contexts and the dispatching. `Management` keeps the point-to-point connections and just routes broadcast telegrams to it. The verb-noun names collapse into plain verbs on the attribute:

- `management.send_broadcast(payload)` → `management.broadcast.send(payload)`
- `management.request_broadcast(payload)` → `management.broadcast.request(payload)`
- `async with management.broadcast()` → `async with management.broadcast.context()`

+53/−32 across 8 files. Worth weighing: it renames public API that already exists on main (`send_broadcast()` and the `broadcast()` context manager), on top of what this PR changes anyway.

I kept `BroadcastContext` as the subscription primitive rather than folding it in too. It could be dropped so that `receive()` manages its own subscription, but an async generator does not subscribe until its first `__anext__`, so a manual `send()` followed by `async for receive()` could miss responses arriving in between — `request()` is only safe because it sends *inside* the subscription.

Keep it, or drop the commit again?

xknx/management/management.py
```diff
@@ -125,15 +129,37 @@ async def disconnect(self, address: IndividualAddress) -> None:
...
+    Class for broadcast communication.
+
+    Broadcast telegrams are sent to every device on the bus and answered by any
+    number of them, so this owns the sending, the open receive contexts and the
+    dispatching of incoming telegrams to them.
+    """
+
+    __slots__ = ("_contexts", "xknx")
+
+    def __init__(self, xknx: XKNX) -> None:
```
kewde, line 152: maybe pass down just self.xknx.cemi_handler ?

farmio: Claude Code  🤖: You're right that it is available at all times - it just isn't yet at the point `Management` is constructed. `xknx.py` builds `Management(self)` on line 84 and `cemi_handler` on line 87, so injecting it today raises `AttributeError` during `XKNX()`. That's a one-line move though, and I checked it is safe: `CEMIHandler.__init__` only stores `self.xknx` and doesn't reach for `xknx.management` until runtime, same for `TelegramQueue`.

So the ordering isn't the real objection. Two things that are:

**Tests.** 39 tests use `xknx.cemi_handler = AsyncMock()` as their seam. If `Broadcast` captures the handler at construction, the mock arrives too late - 16 of them fail, sending into the real unconnected interface. They'd have to move to `patch.object(CEMIHandler, "send_telegram")` (which `test_broadcast_message` already does, so there is precedent) across ~5 files, none related to this PR.

**It's only half a win.** `Management` still needs `xknx` for `task_registry` and `current_address`, so we'd have `Management(xknx)` constructing `Broadcast(cemi_handler)` - two classes in the same module with different injection styles. Narrowing the inner one while the outer still takes the whole object doesn't buy much.

Worth noting the reason the codebase passes `xknx` everywhere: `CEMIHandler` needs `management.process()` and `Management` needs `cemi_handler.send_telegram()`, a genuine cycle that only resolves because both defer the lookup. `Broadcast` is one-directional so it doesn't need that escape hatch - your instinct is right in the abstract.

I'd suggest keeping `xknx` here and doing the injection as its own change if you want it, so the test migration isn't mixed into a PR about typed broadcast responses. Happy to do it either way.

PR #1921, opened 2026-08-21
Remove RemoteValue.__eq__ and make remote values slotted

farmio:
## Description

Follow-up to the `Device.__eq__()` removal, for `RemoteValue`.

`RemoteValue.__eq__()` compared `self.__dict__` against `other.__dict__`, skipping `after_update_cb`. It is the counterpart of `Device.__eq__()` - a leftover of the YAML config handling removed in 1.0, where a device parsed from `xknx.yaml` had to compare equal to a hand constructed one. The only caller left was `test_eq()`, testing the operator for its own sake.

Two things it caused on the way:

- Defining `__eq__` without `__hash__` sets `__hash__ = None`, so remote values were **unhashable** - they could not go into a set or be used as a dict key. `StateUpdater` keys its workers by `id(remote_value)` precisely because it could not rely on this.
- It raised `AttributeError` when compared to any object without a `__dict__` (`rv == None`, `rv == 1`, ...) instead of evaluating to `False` - the same bug that was fixed and then removed for `Device`.

Remote values compare by identity now, which is the natural semantics for these objects, and are hashable again.

With `__eq__` gone, nothing reads `RemoteValue.__dict__` any more, so every class in `xknx/remote_value/` declares `__slots__`. Remote values are the most numerous objects in a running xknx instance - a handful per device - so this drops the per instance `__dict__` and speeds up attribute access. It follows what most of the library already does (`StateUpdater`, `Devices`, `TelegramQueue`, `Telegram`, `GroupAddress`, the `io` transports, the `apci` classes).

Notes for review:

- `dpt_class` stays a plain class attribute on `RemoteValue` - a class body may not both list a name in `__slots__` and assign it. `_RemoteValueGeneric` is the one class that assigns `dpt_class` per instance, so it declares `__slots__ = ("dpt_class",)`; the slot descriptor shadows the base's `dpt_class = None` for that branch of the hierarchy.
- Subclasses that add no attributes get `__slots__ = ()` so they stay dict-less.
- Subclasses outside this package that do not declare `__slots__` themselves keep getting a `__dict__` automatically - so this is not breaking for downstream subclasses, only for setting undeclared attributes on xknx's own remote values.
- Tests no longer patch methods or `dpt_class` onto RemoteValue *instances*: a concrete `_RemoteValue2ByteFloat` subclass and `patch.object()` on the class take over.

```python
rv_1 = RemoteValueSwitch(xknx, group_address="1/2/3")
rv_2 = RemoteValueSwitch(xknx, group_address="1/2/3")

rv_1 == rv_2  # before: True (equal attributes); now: False (distinct objects)
hash(rv_1)  # before: TypeError: unhashable type; now: works
rv_1.my_own_attribute = 1  # before: works; now: AttributeError
```

Fixes # (issue)

## Type of change

## Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,6 +12,17 @@ nav_order: 2
 
 - Drop support for Python 3.10. XKNX requires Python 3.11 or newer now.
 - Remove `xknx.util`. Its only member was `asyncio_timeout` - a backport of `asyncio.timeout` for Python versions not providing it - so use `asyncio.timeout` directly.
+- Remove `RemoteValue.__eq__()`. It compared `__dict__` attributes - a leftover of the YAML config handling removed in 1.0 - and raised `AttributeError` when compared to an object without a `__dict__`. Remote values compare by identity now, which also makes them hashable again, so they can be used in sets and as dict keys.
+- All classes in `xknx.remote_value` define `__slots__` now. Setting an attribute that isn't declared by the class raises `AttributeError` instead of silently creating it.
+
+  ```python
+  rv_1 = RemoteValueSwitch(xknx, group_address="1/2/3")
+  rv_2 = RemoteValueSwitch(xknx, group_address="1/2/3")
+
+  rv_1 == rv_2  # before: True (equal attributes); now: False (distinct objects)
+  hash(rv_1)  # before: TypeError: unhashable type; now: works
+  rv_1.my_own_attribute = 1  # before: works; now: AttributeError
+  ```
 
 ### Connection
 
--- a/test/remote_value_tests/remote_value_test.py
+++ b/test/remote_value_tests/remote_value_test.py
@@ -12,6 +12,13 @@ from xknx.telegram import GroupAddress, Telegram, TelegramDecodedData
 from xknx.telegram.apci import GroupValueWrite
 
 
+class _RemoteValue2ByteFloat(RemoteValue[float]):
+    """Concrete RemoteValue to test the base class implementation."""
+
+    __slots__ = ()
+    dpt_class = DPT2ByteFloat
+
+
 @patch.multiple(RemoteValue, __abstractmethods__=set())
 class TestRemoteValue:
     """Test class for RemoteValue objects."""
@@ -19,8 +26,7 @@ class TestRemoteValue:
     async def test_get_set_value(self) -> None:
         """Test value getter and setter."""
         xknx = XKNX()
-        remote_value = RemoteValue(xknx)
-        remote_value.to_knx = DPT2ByteFloat.to_knx
+        remote_value = _RemoteValue2ByteFloat(xknx)
         remote_value.after_update_cb = Mock()
 
         assert remote_value.value is None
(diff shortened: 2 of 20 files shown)


farmio:
Claude Code  🤖:

Measured this instead of assuming. Setup: synthetic installation of 600 devices (100 each of `Light`, `Switch`, `Cover`, `BinarySensor`, `Sensor`, `Climate`) = **3700 RemoteValue instances**; `tracemalloc` for memory, `timeit` (min of 7 repeats) for speed; `main` vs this branch, same machine (M1 Pro), alternating runs.

## Memory — real and consistent

Bytes allocated for the whole device tree:

| Python | before | after | delta |
|---|---|---|---|
| 3.11.14 | 1 579 552 | 1 387 896 | **−191 656 (−12.1 %)** |
| 3.12.12 | 1 528 352 | 1 366 296 | **−162 056 (−10.6 %)** |
| 3.14.4 | 1 567 464 | 1 374 480 | **−192 984 (−12.3 %)** |

That is **−44 to −52 bytes per RemoteValue**. Isolated, a `RemoteValueSwitch` with two group addresses goes from 384.7 to 336.7 bytes on 3.11/3.14 (376.7 → 336.7 on 3.12).

So ~190 KB on a large installation, ~30 KB on a typical one. Real, but nobody will notice it next to Home Assistant's footprint.

## Speed — negligible, and worth saying out loud

Python 3.12, min of 7 × 2 M iterations, loop overhead subtracted:

| operation | before | after |
|---|---|---|
| attribute read | 2.81–2.90 ns | 2.55–2.71 ns |
| attribute write | 2.80–2.87 ns | 2.50–2.60 ns |
| `group_addresses()` | 0.210 µs | 0.211 µs |
| `RemoteValue.process()` | 0.815 µs | 0.802–0.806 µs |

~8–10 % on the raw attribute ops — but that is 0.3 ns, and end-to-end `process()` moves ~1 %. Same shape on 3.11 (0.631 → 0.628 µs) and 3.14 (0.807 → 0.802 µs).

The reason is worth knowing: since CPython 3.11 a dict-less instance already keeps its attributes in an inline values array, so `__slots__` no longer buys the attribute-access speedup it did on 3.10 and earlier. **Do not merge this expecting throughput.** The memory saving is the whole benefit.

## The one sharp edge — and it belongs to `__eq__`, not to `__slots__`

There is a real performance trap here, and removing `__eq__` is what closes it. Touching an instance's `__dict__` **materializes** the managed dict and permanently drops that object off the fast path. Two identical `RemoteValueSwitch` objects on `main`, one of which had `obj.__dict__` read exactly once:

| | untouched | after one `__dict__` access |
|---|---|---|
| attribute read | 2.87 ns | **15.1 ns (5.3× slower)** |
| `RemoteValue.process()` | 0.813 µs | **1.01 µs (+25 %)** |

Re-timing the untouched object afterwards still gives 0.810 µs, so this is not an ordering artifact — it is per-instance and permanent.

`RemoteValue.__eq__()` read `self.__dict__` *and* `other.__dict__`, so a single comparison de-optimized **both** operands for the rest of their lifetime. Nothing in xknx called it, so this was latent rather than actively costing anyone anything — but downstream code comparing two remote values would have paid 25 % on every subsequent telegram those objects processed. With `__slots__` the failure mode is structurally impossible.

<sub>(This also cost me a wrong first measurement: my initial script probed `hasattr(rv, "__dict__")` on the object it was about to time, which made `main` look 20 % slower than it is. The tables above probe a throwaway instance.)</sub>

PR #1929, opened 2026-08-30
Add DPT 275.100/275.101 room temperature setpoint sets

### LINKED ISSUE #1912: Feature request: RTSM setpoint arrays: DPT 275.100 and 275.101
## Summary

xknx is one datapoint away from mandatory-DPT coverage of the **Room Temperature Setpoint Manager (RTSM)** function block. The missing piece is DPT 275.100 (`DPT_TempRoomSetpSetF16[4]`), plus its optional companion DPT 275.101 (`DPT_TempRoomSetpSetShiftF16[4]`).

> KNX v01.03.01 - HVAC S-Mode FBs 07.19.20 - §5 defines RTSM. §5.7 ("FB Profiles") is the Mandatory/Optional matrix per profile.

## Mandatory DPT coverage — RTSM, all profiles

| RTSM datapoint | Mandatory in | DPT | xknx |
|---|---|---|---|
| `HVACMode` (input) | all profiles | 20.102 | ✅ `DPTHVACMode` (`dpt_20.py`) |
| `HVACModeUser` (input) | all profiles | 20.102 | ✅ same |
| `TempRoomSetpUserAbs` (input) | all profiles | 9.001 | ✅ `DPTTemperature` (`dpt_9.py`) |
| `TempRoomSetpUserOffset` (input) | all profiles | 9.002 | ✅ `DPTTemperatureDifference2Byte` (`dpt_9.py`) |
| `HVACModeEff` (output) | all profiles | 20.102 | ✅ `DPTHVACMode` |
| `TempRoomSetpHeatEff` / `TempRoomSetpCoolEff` (output) | (M) per profile¹ | 9.001 | ✅ `DPTTemperature` |
| **`TempRoomSetpSetHeat` / `TempRoomSetpSetCool` (input)** | **M, matching profile direction** | **275.100** | **❌ not implemented** |
| `TempRoomSetpSetHeatShift` / `...CoolShift` (input) | O, all profiles | 275.101 | ❌ not implemented |

¹ Marked `(M)^C1` in §5.7 — footnote: "can be internal, when combined with RTC" (i.e. still mandatory as a datapoint concept, may be wired internally rather than on the bus if RTSM and RTC are combined in one device).

## TempRoomSetpSet*

> KNX v01.03.01 - HVAC S-Mode FBs 07.19.20 - §5.3.1 ("Setpoint temperatures"): "The Inputs TempRoomSetpSetHeat (DPT_TempRoomSetpSetF16[4], 275.100) and TempRoomSetpSetCool (DPT_TempRoomSetpSetF16[4], 275.100) allow a supervisory system to set the current consistent set of setpoint temperatures conveniently in a single Telegram and are therefore mandatory as well."

Full per-datapoint definitions (field layout, ranges, defaults):

- §5.8.4.3 "Input TempRoomSetpSetHeat" — DPT_ID 275.100
- §5.8.4.4 "Input TempRoomSetpSetCool" — DPT_ID 275.100
- §5.8.4.5 "Input TempRoomSetpSetHeatShift" — DPT_ID 275.101
- §5.8.4.6 "Input TempRoomSetpSetCoolShift" — DPT_ID 275.101

## Why I'm asking

Running a KNX HVAC setup (Siemens WIDE UP 258) and ultimately need a way to set the room setpoint from Home Assistant for both Comfort and Building Protection modes, preferably without a template/automation workaround.  

The alternative to 275.100 is one discrete target_temperature_<mode>_address per preset. It isn't implemented in the current `knx` integration schema; `climate` only exposes a single `target_temperature_address`/`target_temperature_state_address` pair.

## Suggested implementation shape

Not verified, claude: The shape looks straightforward given existing precedent:

- `DPTComplex`/`DPTComplexData` (see `xknx/dpt/dpt_251.py`, the RGBW implementation) is the existing abstraction for a fixed-layout multi-field payload.
- Each 2-byte field is the identical KNX-float encoding `DPT2ByteFloat` already implements in `xknx/dpt/dpt_9.py` (`from_knx`/`to_knx`) — the four fields can reuse that codec directly instead of reimplementing float packing.
- 275.101 (Shift) is the same layout/codec with a different semantic range (deltas in K, can be negative) — cheap to add alongside 275.100 once the container exists, even though it's optional per the FB profile.

Rough sketch of what that would look like, following the `DPTColorRGBW` pattern:

```python
@dataclass(slots=True)
class TempRoomSetpSetF16Data(DPTComplexData):
    """4 setpoints per DPT_TempRoomSetpSetF16[4] (275.100)."""

    comfort: float | None = None
    standby: float | None = None
    economy: float | None = None
    building_protection: float | None = None

    @classmethod
    def from_dict(cls, data: Mapping[str, float | None]) -> "TempRoomSetpSetF16Data":
        return cls(
            comfort=data.get("comfort"),
            standby=data.get("standby"),
            economy=data.get("economy"),
            building_protection=data.get("building_protection"),
        )

    def as_dict(self) -> dict[str, float | None]:
        return {
            "comfort": self.comfort,
            "standby": self.standby,
            "economy": self.economy,
            "building_protection": self.building_protection,
        }


class DPTTempRoomSetpSetF16(DPTComplex[TempRoomSetpSetF16Data]):
    """Abstraction for KNX 8 octet DPT_TempRoomSetpSetF16[4] (DPT 275.100)."""

    data_type = TempRoomSetpSetF16Data
    payload_type = DPTArray
    payload_length = 8
    dpt_main_number = 275
    dpt_sub_number = 100
    value_type = "temp_room_setp_set_f16"

    @classmethod
    def from_knx(cls, payload: DPTArray | DPTBinary) -> TempRoomSetpSetF16Data:
        raw = cls.validate_payload(payload)
        return TempRoomSetpSetF16Data(
            comfort=DPT2ByteFloat.from_knx(DPTArray(raw[0:2])),
            standby=DPT2ByteFloat.from_knx(DPTArray(raw[2:4])),
            economy=DPT2ByteFloat.from_knx(DPTArray(raw[4:6])),
            building_protection=DPT2ByteFloat.from_knx(DPTArray(raw[6:8])),
        )

    @classmethod
    def _to_knx(cls, value: TempRoomSetpSetF16Data) -> DPTArray:
        return DPTArray(
            DPT2ByteFloat.to_knx(value.comfort).value
            + DPT2ByteFloat.to_knx(value.standby).value
            + DPT2ByteFloat.to_knx(value.economy).value
            + DPT2ByteFloat.to_knx(value.building_protection).value
        )
```

(275.101 would be a near-identical second class — same layout, `dpt_sub_number = 101`, values interpreted as signed shifts in K rather than absolute °C.)

Whether/how this should also surface as a new config option on the `Climate`/`ClimateMode` device (e.g. an input consuming the composite set instead of/alongside `target_temperature_address`) feels like a separate design decision with its own tradeoffs — happy to discuss that separately once the raw DPT support question is settled, rather than bundling both into one open thread.

---

This issue was drafted with AI assistance — Claude, Sonnet 5 (model id `claude-sonnet-5`). The spec citations above were checked against the primary KNX Standard documents by me and the agent had a copy of xknx source.

ilian002:
## Description

DPT_TempRoomSetpSetF16[4] (275.100) and DPT_TempRoomSetpSetShiftF16[4] (275.101) carry the four comfort/standby/economy/building_protection setpoints (°C) or setpoint shifts (K), respectively, of the RTSM and related HVAC function blocks in one 8-octet telegram.

These DPTs are not in the Datapoint Types document as of KNX v02.02.01 - Datapoint Types 03.07.02; they are defined in KNX v01.03.01 - HVAC S-Mode FBs 07.19.20 - §9.4, which is added to the confirmed-documents table in AGENTS.md.

Each field delegates to DPTTemperature / DPTTemperatureDifference2Byte for encoding and range validation. The reserved F16 pattern 0x7FFF marks a field as "not used" (§9.4: "only the value 7FFFh shall be used to denote invalid data") and maps to None.

DPT 275.100 verified against Siemens WIDE UP 258. Masking works as expected (partial temperature update possible). It behaves identically to writing the unmasked fields via their dedicated 9.001 objects.

Note: DPTTemperature/DPT2ByteFloat (dpt_9.py) aren't fully §3.10-compliant - value_max is 670760 (or the raw 670760.96 for several sibling 9.xxx types). The spec and ETS GUI use 670433.28, which encodes to 0x7FFE. 0x7FFF is documented as "invalid data" for all of DPT 9.xxx, not just composite formats, but from_knx() raises ConversionError on it instead of returning NaN. This is why _pack_float_or_not_used/_unpack_float_or_not_used sidestep the base class for the masked case.

Fixes #1912

**Update — rebased onto `main` (4ff8ddf8).**

**Depends on #1930**, which fixes the DPT 9 issue noted above. Until it lands the `0x7FFF` guard stays. After #1930 I'll rebase, drop the then-unreachable guard, correct the ranges to `-273..670433.28` / `-671088.64..670433.28`, and add the two tests from review.

Renamed per the review — §9.4.2 is "4x set of RoomTemperature Setpoint Shift Values", a set of setpoint *shifts*: `DPTRoomTemperatureSetpointShiftSet`, `RoomTemperatureSetpointShifts`, `value_type` `room_temperature_setpoint_shift_set`.

## Type of change

## Checklist


farmio:
Hi 👋! 
Thank you for your contribution! Looking at this, I think we should fix DPT 9 - it shall use `0x7FFF` as invalid marker too. Reading the specs of DPT 9
> For all Datapoint Types 9.xxx, the encoded value 7FFFh shall always be used to denote invalid data. [-671 088,64 … 670 433,28]

It's a good opportunity to making such a breaking change now since we are on the edge of releasing xknx 4.0

farmio (commented):
Claude Code  🤖:

Reviewed at b34ac848. `pytest` 3963 passed, `mypy xknx/` clean, `ruff check` + `ruff format --check` clean.

**Spec check.** Pulled `07_19_20 HVAC S-Mode FBs v01.03.01 AS.pdf` §9.4 (pages 203-204). The citation is traceable and correct - version, document number and title all match.

- 8 octets `F16F16F16F16`, field order Comf / Stdby / Eco / BProt with octet 8 first → the `raw[0:2]`…`raw[6:8]` mapping is right.
- The quoted sentence is verbatim.
- 275.101 field range -670 760 K … 670 760 K, 0,01 K → matches `DPTTemperatureDifference2Byte` exactly.
- Hand-checked every raw tuple in the tests (21.0 → `0x0C1A`, -2.5 → `0x8706`, -5.0 → `0x860C`, …), all correct.

One deviation the PR doesn't mention: §9.4.1 gives the **275.100** fields as "F16, -273°C to 655,34°C", not 9.001's full range. `655,34` is almost certainly a table artifact (a U16 × 0,01 maximum), and the FB datapoint descriptions (§4.1.7.7.3/.4) say "full" range for these - so delegating to `DPTTemperature` is the defensible call. But `-273..670760` in the `RoomTemperatureSetpoints` docstring sits directly under a §9.4 citation; a one-line comment noting the deviation would keep the citation honest.

**1. The `_NOT_USED` comment misstates the spec.** It reads "the largest value DPT2ByteFloat can encode, repurposed as a per-field 'not used' marker". KNX v02.02.01 - Datapoint Types 03.07.02 - §3.10 says: *"For all Datapoint Types 9.xxx, the encoded value 7FFFh shall always be used to denote invalid data"*, with the range given as `[-671 088,64 ... 670 433,28]`. So `0x7FFF` isn't a valid encoding being repurposed - it is the standard invalid marker, and it only *looks* like "largest encodable" because xknx's `DPT2ByteFloat.value_max` is too wide.

**2. `get_dict_schema()` advertises a max that `to_knx()` rejects.** `_RANGE_TEMPERATURE` derives `value_max = 670760` from `DPTTemperature`, but the guard rejects everything from **670597.12** upward (that's where the mantissa rounds to 2047 → `0x7FFF`). A ~163 wide band is advertised as valid and raises. Fixed upstream in #1930, which caps `DPT2ByteFloat.value_max` at `670433.28` and makes `from_knx()` reject `0x7FFF` - exactly the point your PR description makes. Once both land, the guard in `_pack_float_or_not_used` becomes unreachable and can go, and the advertised range becomes truthful.

**3. Range constants belong in `helpers/metadata.py`.** The convention here is that field `metadata` range dicts live there - including single-use ones like `RANGE_INT32` (only `dpt_235`). Suggest moving these as `RANGE_TEMPERATURE` / `RANGE_TEMPERATURE_DIFFERENCE`. Importing `..dpt_9` there is cycle free (`dpt_9` imports only `.dpt` / `.payload`, and `helpers/relative_control.py` already imports `..dpt_1` / `..dpt_3`), so the derive-from-the-class trick survives the move.

**4. Naming - public API, so worth settling now.** `DPTRoomTemperatureSetpointSetShift` / `room_temperature_setpoint_set_shift` parses as "setpoint-set shift". The spec title is "4x set of RoomTemperature Setpoint **Shift** Values" - a set of setpoint *shifts*. `DPTRoomTemperatureSetpointShiftSet` / `room_temperature_setpoint_shift_set` (+ `RoomTemperatureSetpointShifts`) reads truer, and `value_type` renames are breaking after a release.

**5. Two small test gaps.** Nothing pins `from_knx()` on an out-of-range raw field (e.g. `0xF8 0x00` → `ConversionError` from `DPTTemperature`). That's meaningful behaviour: one bad field makes the whole 8 octet payload undecodable, and some devices use `0x8000` / `0xFFFF` rather than `0x7FFF` as "invalid". Strict is the right call, but it should be a pinned decision rather than an accident. Also missing: a round-trip at the largest encodable value, the boundary just below the guard.

**Fine as-is:** `_RoomTemperatureSetpointSet` mirrors `helpers/relative_control.py::_RelativeControlDimming` closely (`dataclass_fields` iteration, `__slots__ = ()`, `Self`) - good fit. `from_dict()` not catching `AttributeError` (unlike `RGBWColor`) is harmless, `DPTComplex.to_knx()` catches it and still yields a `ConversionError`. Changelog heading and placement match convention; `docs/sensor.md` has no exhaustive value type table. `__init__.py` import ordering is right and the lookup test picks the classes up.

Nothing blocking - 1 and 3 are quick fixes, 2 is handled in #1930, 4 is your call.

farmio:
Sorry for the long delay! Just merged https://github.com/XKNX/xknx/pull/1930 raising a ConversionError for the invalid data on DPT 9 - I think this could be used from this DPT.

If you need, feel free to rename `xknx.dpt.dpt_9._INVALID_DATA` to `xknx.dpt.dpt_9.DPT_9_INVALID_DATA` to import it instead of defining it twice.

PR #1937, opened 2026-09-03
feat(management): add Master Reset to dm_restart_r_co/dm_restart

kewde:
## Description

DM_Restart_RCo (KNX v02.01.02 - Management Procedures 03.05.02 - §3.7.3) covers both a Basic Restart and a Master Reset as one Management Procedure, distinguished by the spec's `mpp_RestartType` parameter. Only Basic Restart was implemented so far. This adds Master Reset support to `dm_restart_r_co`/`dm_restart` via new keyword-only `master_reset`, `erase_code` and `channel_number` parameters.

Unlike a Basic Restart (fire-and-forget, no application-layer confirmation), a Master Reset is confirmed by an `A_Restart_Response`-PDU before the device restarts, so both functions now return the parsed `apci.RestartMasterResetResponse` (error code + process time, the latter being the spec-mandated minimum wait before retrying) for a Master Reset, or `None` for a Basic Restart as before. A non-zero error code raises `ManagementConnectionError`, with the meaning from Table 5 (§3.7.1.2.3.1): Access denied / Unsupported Erase Code / Invalid Channel Number.

`erase_code` defaults to 01h "Confirmed Restart" and is validated to 01h-08h - Table 4 (§3.7.1.2.3.1) defines only those; 00h and 09h-FFh are reserved and the spec is explicit that a Management Client shall not send them.

Non-breaking: the new parameters are keyword-only with defaults matching prior behavior, and the return type widens from `None` to `RestartMasterResetResponse | None` (existing callers only ever saw `None`).

Targets `main` directly now that #1834 (its original base) merged — first of a stack of PRs implementing the KNX device-programming Management Procedures (`DM_Restart_RCo` → memory read/write → function property → load state machine).

## Type of change

## Checklist


farmio:
Claude Code  🤖:

Reviewed at a89234ed — one commit, 3 files, clean against `main`. Locally: 3967 passed, ruff/format/mypy clean.

Nice and small, and the success test asserting the outgoing telegram exactly (`erase_code`, `channel_number` on the wire) is the right pattern.

### Verified against the spec

- **Error codes match Table 5 (§3.7.1.2.3.1) exactly**: 01h Access denied, 02h Unsupported Erase Code, 03h Invalid Channel Number, and 04h–FFh reserved — so the `"Unknown Error"` fallback is correct rather than defensive padding. (Table 5 writes "Access denied"; the map has "Access Denied". Trivial, but the strings read as quotations.)
- The docstring's "a Master Reset should only be requested of a device confirmed to support it" is §3.7.3's own "The Management Client shall prior to calling this Management Procedure with a Master Reset verify that this feature is effectively supported by the Management Server."
- The connection-breakdown/`dmp_disconnect_r_co` note matches §3.7.3 too.
- Keyword-only parameters with prior-behaviour defaults, and the return widening from `None` to `… | None`, are non-breaking as described.

### 1. `erase_code=0` defaults to a reserved value the spec forbids clients from sending

This is the one I'd fix before merging. Table 4's final row (§3.7.1.2.3.1, "Definition of Erase Code and Channel Number") groups **00h together with 09h to FFh**:

> These values are reserved. The Management Client shall not use these Erase Codes. The Management Server shall on reception of an A_Restart-PDU with an Erase Code value in this range — neither execute a Basic Restart nor any Master Reset, and — respond with an A_Restart.res with Error Code "Unsupported Erase Code".

The defined codes are 01h–08h only. So `dm_restart_r_co(conn, master_reset=True)` with the default `erase_code=0` sends a reserved code that a conformant device *must* refuse with 02h while doing nothing — which this code then turns into `ManagementConnectionError`. The default path for a Master Reset can't succeed, and sending it at all is what the spec tells a Management Client not to do.

**Heads-up on why 00h looks defensible** — there's a genuine contradiction in the document. §3.7.1.1 (Basic Restart) says:

> The Application Layer of the Management Server shall not confirm the A_Restart-service if a Basic Restart is called; to obtain the same result with an AL-confirmation, the Management Client should instead call a Master Reset with **Erase Code 00h**.

That prose is stale. Table 4 assigns exactly that role to **01h "Confirmed Restart"**, whose own description reads "No Resource value shall be reset. This encoding shall allow using the Master Reset as a confirmed alternative to the unconfirmed Basic Restart." (01h's cell even still carries a dangling "Both Erase Code values shall have the same effect" line with only one code left in the row — the fingerprint of an earlier edition where 00h shared it.) Same category of editing slip as the §3.20.2 one you documented in #1938.

So: `1` (Confirmed Restart) is the sensible default if there is to be one, or make `erase_code` required whenever `master_reset=True`. A client-side `1 <= erase_code <= 8` guard would also stop a caller reaching a device with a code the spec says not to send — the `erase_code=0x99` case the refused-test uses deliberately.

### 2. Citation precision

`_ERROR_CODES`'s comment credits "KNX v02.01.01 - Application Layer 03.03.07 - §3.4.2.2 … as enumerated for DM_Restart_RCo's exception handling (§3.7.3)". Neither is where the values live. AL §3.4.2.2 defines the PDU and names `error_code` as a parameter but assigns no values; §3.7.3's entire "Exception handling" clause is "The general exception handling shall be applicable." Both Table 4 (erase codes) and Table 5 (error codes) sit in **MP §3.7.1.2.3.1 "General requirements"**. Given how strict AGENTS.md is about traceability, worth pointing at the table that actually defines them.

### 3. The realistic failure mode isn't documented

§3.7.3 footnote 11: *"Existing implementations may not check bit 0 of octet 7 and may not react as expected. They may ignore the service entirely, only perform a Basic Restart if a Master Reset is called or exhibit another behaviour."*

Against such a device `conn.request()` waits for an `A_Restart_Response` that never arrives and surfaces `ManagementConnectionTimeout` — quite possibly after the device already performed a Basic Restart and dropped the connection. That's the likely outcome of a Master Reset on older hardware, it isn't the `ManagementConnectionError` the `:raises:` advertises, and it's the concrete reason §3.7.3 requires checking support first. One line in the docstring would connect the "verify support first" sentence to what happens if you don't.

### 4. `process_time` is returned without saying what it's for

§3.7.3 gives it a specific contract: *"The Management Client shall consider this mpp_ProcessTime as time-out after which communication attempts following a Master Reset shall be considered without success."* Right now the docstring and changelog both just call it "process time", so a caller has the one actionable number in the response and no indication that it's the mandated wait-before-retry.

### 5. Changelog: duplicate `### Features` heading

#1834 landed on `main` as 097b0c95 while this was open, and it added a `### Features` section to `# Unreleased changes`. This PR inserts a second one right after `### Bugfixes`, so the section now appears twice under one release. The entry should move into the existing section rather than opening a new one.

Two small extras: the PR description still says "Stacked on `kewde/add-dmp-interface-obj-rw`", which is no longer true now that #1834 is merged and this targets `main` cleanly; and there's no test for the `dm_restart()` master-reset path or for the `"Unknown Error"` fallback, though both are one-liners next to what's already there.

kewde:
Thanks for the thorough review — all verified against the spec text directly (Table 4/5, §3.7.1.2.3.1) and fixed in 788605c6:

1. **Reserved erase_code** — confirmed: Table 4 defines only 01h-08h, groups 00h with 09h-FFh as reserved and explicit that clients shall not send them. `erase_code` now defaults to `0x01` (Confirmed Restart) and raises `ValueError` outside 01h-08h before sending anything.
2. **Citation** — moved to Table 5/§3.7.1.2.3.1, and fixed "Access Denied" → "Access denied" to match the table's own casing.
3. **Realistic failure mode** — docstring now notes that against a non-conformant device (footnote 11), `conn.request()` times out with `ManagementConnectionTimeout` rather than raising `ManagementConnectionError`.
4. **process_time** — docstring now states it's the spec-mandated minimum wait before retrying, not just an FYI duration.
5. **Duplicate changelog heading** — merged into the `### Features` section #1834 added instead of duplicating it.

Also added tests for `dm_restart()`'s own master-reset path and the "Unknown Error" fallback, and updated the PR description (stale "stacked on" line — this now targets `main` directly since #1834 merged).

PR #1938, opened 2026-09-03
feat(management): add memory read/write/verify procedures

kewde:
## Description

Add `dmp_mem_read_r_co`, `dmp_mem_write_r_co` and `dmp_mem_verify_r_co` (KNX v02.01.02 - Management Procedures 03.05.02 - §3.18.2/§3.16.2/§3.17.2) for the 64 KiB `A_Memory_*` address space, and the `dmp_user_mem_*_r_co` siblings (§3.21.2/§3.19.2/§3.20.2) for the 1 MiB `A_UserMemory_*` space.

Both families chunk transfers to the wire APDU length or the count field's own limit (63/15 elements), whichever binds first. The header-octet constants (`MEMORY_HEADER_OCTETS = 3`, `USER_MEMORY_HEADER_OCTETS = 4`) are verified against KNX v02.01.01 - Application Layer 03.03.07 - §3.5.3/§3.5.4/§3.5.6.2/§3.5.6.3's own "Maximum APDU Length - 3/4" error-handling clauses, not just derived from the existing `apci.py` implementation. The write procedures offer an optional read-back `verify` since neither `A_Memory_Write` nor `A_UserMemory_Write` are confirmed at the application layer - a caller not verifying is expected to leave a `write_delay` instead.

**Notable finding:** `DMP_UserMemVerify_RCo`'s spec text (§3.20.2) is internally inconsistent in the KNX Standard document itself - a copy-paste artifact from `DMP_MemVerify_RCo` (§3.17.2) that left its "Used Application Layer Services" clause naming plain `A_Memory_Read` and its maximal-size bullet at 12 octets, while its own sequence diagram correctly shows `A_UserMemory_Read-PDU`/`A_UserMemory_Response-PDU`. `DMP_UserMemRead_RCo` (§3.21.2) has the identical prose slip ("subsequent A_Memory_Read-PDUs") but the correct 11 octet size, confirming the pattern rather than a one-off typo. The implementation follows the sequence diagrams (`A_UserMemory_Read`, 11 octets) with the discrepancy documented in the module's docstring.

Non-breaking: all new files/exports, no changes to existing signatures.

Stacked on `kewde/dm-restart-master-reset` (#1937) — second of the stack (`DM_Restart_RCo` → memory read/write → function property → load state machine).

## Type of change

## Checklist


farmio:
Claude Code  🤖:

Reviewed at 6b43baf5, diffed against its own base (`kewde/dm-restart-master-reset`) so this is about the two memory commits only. Locally: 4016 passed, ruff/format/mypy clean.

### Verified against the spec

I checked the chunk arithmetic rather than taking it on trust, and it lands exactly on the spec's own numbers:

- `min(MEMORY_MAX_COUNT, 15 - MEMORY_HEADER_OCTETS)` = **12** octets, which is literally what §3.16.2 and §3.17.2 state for a Management Server without L_Data_Extended.
- `min(USER_MEMORY_MAX_COUNT, 15 - USER_MEMORY_HEADER_OCTETS)` = **11**, matching §3.19.2 and §3.21.2.
- The `verify` / `write_delay` if-else mirrors §3.16.2's sequence one-for-one ("if verify = enabled → A_Memory_Read … else delay for programming the memory in the device"), footnote 12 included.

**Your "notable finding" checks out** — I reproduced it independently. §3.20.2 lists `A_Memory_Read` under "Used Application Layer Services" and puts the maximal size at 12 octets, while its own sequence diagram shows `A_UserMemory_Read-PDU`/`A_UserMemory_Response-PDU`. §3.21.2 has the mirror-image slip: prose says "subsequent A_Memory_Read-PDUs" but the size (11) and the service list (`A_UserMemory_Read`) are both right. §3.19.2 independently says 11 as well, so 11 is the majority and consistent reading. Following the sequence diagrams is the right call, and documenting it in the module docstring is the right place for it.

Also worth noting: raising `ValueError` when `max_apdu_length` leaves no room beats what the interface object procedures do today (`max(1, ...)`, which knowingly emits an over-long frame). That's the better pattern — worth backporting to #1834's family rather than diverging.

### Findings

**1. No upper-bound check on `address + size`.** All six validate the *start* address only. `MemoryRead/Write/Response.to_knx()` do bounds-check the address, so this isn't silent corruption — but a transfer that walks past `0xFFFF` (or `0xFFFFF`) fails mid-loop with `ConversionError`, which is an `XKNXException` and *not* a `ManagementConnectionError`, so it slips past callers catching the documented type. For `dmp_mem_write_r_co`/`dmp_user_mem_write_r_co` that also means a partially applied write. This is the same nit as in #1834, which you already fixed there with an upfront `start_index + count - 1 <= 0xFFF` check — the equivalent here is `address + len(data) - 1 <= 0xFFFF` / `0xFFFFF` before the first PDU goes out.

**2. Verify mismatches raise `ManagementConnectionError`, not `PropertyVerificationError`.** #1834 introduced that exception precisely so "the device's data differs" is distinguishable from "the transport failed". All four verify paths here (both `verify=True` branches and both `*_verify_r_co`) use the generic one, so a caller can't tell a mismatch from a disconnect. The name is Property-specific, so this probably wants either a shared parent or a rename — but the two families shouldn't disagree about it.

**3. The response's address isn't checked against the request.** `MemoryResponse` echoes `address`, and `conn.request()` matches on APCI type alone, so a response for a different address is accepted as the data for the one asked for. `len(data)` is checked; the address is free to check alongside it.

### Rebase

The stack below this has moved since the branch was cut, and it shows in the changelog: there are now **two `### Features` sections inside `# Unreleased changes`**, one carrying the restart/memory entries and one carrying the interface object entry. They need to collapse into one on rebase. Two smaller things that come with it: the entries say `xknx.management.procedures.device`, where the sibling entries now say `xknx.management.procedures` (that's where the procedures are re-exported), and the `Resources 03.05.01` row is still missing from the AGENTS.md table even though `PID_MAX_APDU_LENGTH` is cited in `const.py` and in all six docstrings — `| v01.10.01 | Resources | 03.05.01 |`, version verified against the document.

kewde:
Thanks — glad the chunk arithmetic and the §3.20.2 finding held up independently. Fixed in 3cea5103:

1. **Upper-bound check** — added `address + size/len(data) - 1 <= 0xFFFF`/`0xFFFFF` validation upfront in all six, before anything is sent, mirroring `dmp_interface_object_read_r`'s `start_index + count - 1` check.
2. **`VerificationError`** — added as a new parent of `PropertyVerificationError` in `xknx.exceptions`. All four verify paths (`dmp_mem_write_r_co`/`dmp_user_mem_write_r_co`'s `verify=True`, and `dmp_mem_verify_r_co`/`dmp_user_mem_verify_r_co`) now raise it for content mismatches specifically, so callers can catch verification failures uniformly across Property and Memory operations. Non-breaking — existing `PropertyVerificationError`/`ManagementConnectionError` catches are unaffected.
3. **Address echo check** — added to all six; a response for the wrong address now raises `ManagementConnectionError` instead of being silently accepted.
4. Also fixed the changelog module path (`xknx.management.procedures`, matching the sibling entries) and confirmed the duplicate `### Features` heading and the AGENTS.md Resources row are already resolved from the rebase.

Backport to #1834's family (using `ValueError` instead of `max(1, ...)`) noted but left out of scope here since that's landed on `main` already.

100% coverage maintained (added 12 new tests for the two branches each of the six files gained).

PR #1939, opened 2026-09-03
feat(management): add DM_FunctionProperty_Write_R

kewde:
## Description

Add `dm_function_property_write_r`/`dm_function_property_write_r_conn` (KNX v02.01.02 - Management Procedures 03.05.02 - §3.30.1) to invoke a Function Property on an interface object (`A_FunctionPropertyCommand`, confirmed by `A_FunctionPropertyState_Response`), and `dmp_ext_function_property_write_r`/`dmp_ext_function_property_write_r_conn` (§3.30.2 DMP_ExtFunctionProperty_Write_R), the extended-addressing sibling using `A_FunctionPropertyExtCommand`/`A_FunctionPropertyExtState_Response` (addressing by Interface Object Type + Object Instance instead of a connection-local object index).

Per the spec's own "Error handling"/"Exception handling" clause, `return_code`'s meaning and handling are Function Property specific and depend on the Configuration Procedure calling this one - so both return the full response rather than raising on a nonzero code, unlike e.g. `dmp_interface_object_write_r`'s `nr_of_elem=0` check where the Application Layer spec itself defines that as a universal failure signal.

Both take `max_apdu_length` and reject an over-long `command` upfront - unlike the Property/Memory procedures, this is not chunked, since the spec describes no way to split a Function Property command across multiple PDUs.

**No read-only counterpart added:** there is no separate DM_/DMP_ Management Procedure in the spec's catalog to read a (non-extended) Function Property's state without invoking a command - only the extended family (`A_FunctionPropertyExtState_Read`) appears as a step inside other, larger Configuration Procedures. A caller needing that should send `apci.FunctionPropertyStateRead`/`FunctionPropertyExtStateRead` via `conn.request()` directly.

**Included fixes:**
- Converts `apci.FunctionPropertyCommand` and `apci.FunctionPropertyExtCommand` to `APCIRequest[...]`, like their `*StateRead` siblings already were - both carry a spec-defined response and were missed when the other point-to-point request services were converted, mirroring the precedent `PropertyValueWrite` fix already in this stack.
- `FunctionPropertyCommand.to_knx()` now validates `object_index`/`property_id` ranges, matching what `FunctionPropertyExtCommand` already does for its own fields.

100% patch coverage. Non-breaking: all new files/exports, the `APCIRequest` conversions (additive response-type inference/validation only), and the new `max_apdu_length` parameters (keyword-only... actually positional-with-default, but additive - existing calls with fewer arguments are unaffected).

Stacked on `kewde/dmp-memory-procedures` (#1938) — third of the stack (`DM_Restart_RCo` → memory read/write → function property → load state machine).

## Type of change

## Checklist


farmio:
Claude Code  🤖:

Reviewed at 7fddfb68. Locally: 4022 passed, ruff/format/mypy clean.

### The branch has diverged from its own base

This is the main thing. GitHub reports the PR as conflicting, and the cause isn't a stray line — `git merge-base kewde/dmp-memory-procedures kewde/dm-function-property-write` is **153ef5a0**, a commit on `main`. The two branches share no stack history at all: this one carries its own duplicate copies of every commit in #1834 and #1938, in their *pre-review* form (before the `max_apdu_length` chunking, the upfront overflow check, the `start_index=0` support and the `request()` API adaptation). Conflicts land in `docs/changelog.md`, `xknx/telegram/apci.py` and both `__init__.py` files.

So `git diff` against the base branch shows 35 files and +5568, where the actual new work here is two commits — a1afd1d0 and 7fddfb68. Everything below wants rebasing onto the current `kewde/dmp-memory-procedures` before this is reviewable as a unit or mergeable. Worth doing bottom-up: #1834 has moved too (it picked up a docs commit after the last round).

While it's rebased: the AGENTS.md hunk adds the `Application Interface Layer 03.04.01` row that #1834 already carries, so it'll collapse — but the `Resources 03.05.01` row that #1938's `PID_MAX_APDU_LENGTH` citations need is still missing from the whole stack: `| v01.10.01 | Resources | 03.05.01 |`.

### On the code itself

**These are the first procedures in the stack to follow the documented naming convention** — a `_conn` variant taking an open connection plus a `(xknx, individual_address, ...)` variant that opens its own. That's exactly what `AGENTS.md` and the `procedures/__init__.py` docstring describe, and it makes the interface object procedures below (conn-only, no `_conn` suffix) the odd ones out. That's farmio's and your call to settle, but this PR is evidence for aligning them rather than leaving two patterns side by side.

**Returning the response instead of raising on a non-zero `return_code` is right.** §3.30.1's error handling does make the interpretation Configuration-Procedure-specific, so there's no universal failure signal to raise on — unlike `nr_of_elem = 0`, which the Application Layer defines as one. The docstring saying so is the right way to record it.

**The PR description doesn't mention `DMP_ExtFunctionProperty_Write_R`.** 7fddfb68 adds a second procedure pair, and the body still reads as if only the base one is in scope ("No read-only counterpart added" etc.). Worth updating so the description matches what's being merged.

### Findings

**1. `FunctionPropertyCommand.to_knx()` validates nothing, while its extended sibling validates everything.** `_pack_function_property_ext_header()` raises `ConversionError` for an out-of-range interface object type, object instance or property ID, so `FunctionPropertyExtCommand` is well covered. `FunctionPropertyCommand` just does `struct.pack("!BB{size}s", object_index, property_id, data)`, so `object_index=256` surfaces as a raw `struct.error` rather than a `ConversionError`. The procedures document "0-255" and "1-255" but don't check either. Cheap to make the two consistent.

**2. No bound on the command length.** Every other procedure in this stack takes `max_apdu_length` and sizes its PDU against it; these accept a `command: bytes` of any length and will happily build an over-long APDU. Function Property commands are short in practice, so this is a consistency point more than a live bug — but the asymmetry is worth a deliberate decision rather than an accident.

kewde:
Thanks — fixed in 42c694c3:

### Rebase
Confirmed the divergence — the branch was carrying pre-review duplicates of #1834/#1938's commits instead of stacking cleanly. Rebased onto `kewde/dmp-memory-procedures`'s current tip; `git merge-base` now equals that branch's HEAD exactly. Also fixed the `Application Interface Layer`/`Resources` AGENTS.md rows and the module-path wording in the changelog bullets while resolving the rebase conflicts.

### Findings
1. **`FunctionPropertyCommand.to_knx()` validation** — added the same range checks `_pack_function_property_ext_header()` already does for the extended family: `ConversionError` for `object_index` outside 0-255 or `property_id` outside 1-255 (matching what my own procedure docstrings document), instead of a bare `struct.pack` letting an out-of-range value surface as `struct.error`.
2. **No bound on command length** — added `max_apdu_length` to both procedure pairs, rejecting a `command` that doesn't fit via `ValueError` before sending anything. Not chunked, deliberately - the spec gives no mechanism to split a Function Property command across multiple PDUs, unlike the array-shaped Property/Memory services.

### On the rest
- PR description updated to cover `DMP_ExtFunctionProperty_Write_R` explicitly.
- Naming-convention observation (base file lacking a bare non-`_conn` name vs. the `_conn`-suffixed pattern here) - noted, but that's the interface object procedures already on `main`, out of scope for this PR to change unilaterally.
- Agreed on both "the code itself" notes (return-value-not-raise reasoning, and this being the first properly `_conn`-suffixed pair) - no changes needed there.

100% coverage maintained.

farmio:
Claude Code  🤖:

42c694c3 closes both findings from my last pass. `FunctionPropertyCommand.to_knx()` now validates its fields the way `_pack_function_property_ext_header()` already did for the extended sibling, and the `max_apdu_length` bound with an upfront rejection is the right call — a Function Property command isn't array-shaped and the spec gives no way to split one, so rejecting beats chunking. I checked both new constants against the classes: `FUNCTION_PROPERTY_HEADER_OCTETS = 3` matches `FunctionPropertyCommand.calculated_length()` (`3 + len(data)`) and `FUNCTION_PROPERTY_EXT_HEADER_OCTETS = 6` matches the extended one. Giving the `to_knx()` change its own Bugfixes entry rather than folding it into a feature bullet is also right — it's a behaviour change to existing public API. Locally: 4045 passed, ruff/format/mypy clean.

Two nits from the new code, both minor.

### 1. `FunctionPropertyCommand`'s declared default is now unserializable

The new guard is `1 <= property_id <= 0xFF`, but the dataclass still declares `property_id: int = 0`:

```
>>> apci.FunctionPropertyCommand().to_knx()
ConversionError: <ConversionError description="Property ID out of range." />
```

That makes it the odd one out in its own family — `FunctionPropertyStateRead` and `FunctionPropertyStateResponse` both default `property_id = 0` too, and both still serialize fine because neither has a guard. So the class now advertises a default it rejects. Dropping the default (making `property_id` required, as `RestartMasterReset.erase_code` is) would make the constructor say what `to_knx()` enforces. Worth a sanity check on the lower bound while you're there: rejecting PID 0 reads right to me, since there's no `property_index` in this service for a 0 to select, but it is newly rejecting something that previously encoded.

### 2. No `max_apdu_length` sanity check, unlike the memory procedures

`max_command_length` is computed as a bare subtraction, so a nonsensical `max_apdu_length` produces a negative budget and a confusing message rather than a clean rejection:

```
>>> await dm_function_property_write_r_conn(conn, 0, 1, b"", max_apdu_length=0)
ValueError: command (0 octets) does not fit max_apdu_length 0 (header is 3 octets, leaving -3 for command data)
>>> await dm_function_property_write_r_conn(conn, 0, 1, b"\x01", max_apdu_length=-5)
ValueError: command (1 octets) does not fit max_apdu_length -5 (header is 3 octets, leaving -8 for command data)
```

The six memory procedures all guard this explicitly (`if max_apdu_length <= 0: raise ValueError(f"max_apdu_length must be positive, ...")`), and these two don't, so the same bad input is reported differently depending on which procedure you reach for. It does still raise either way — this is about the message and the consistency, not a missed failure.

Nothing here blocks. The one thing that does is the rebase: the PR now targets `main`, but its merge-base is still 097b0c95, so it carries unsquashed copies of the #1937 and #1938 commits that `main` now has as 2b436e19 and 4ff8ddf8 — hence the conflicts in `docs/changelog.md`, both `__init__.py` files and `const.py`. Rebasing onto current `main` should drop those and leave your own three commits.

kewde:
Claude Code  🤖:

Both nits fixed at 797770f0.

1. `FunctionPropertyCommand.property_id` has no default anymore (reordered ahead of `object_index` in the dataclass to keep it before the two defaulted fields). All existing callers use keyword arguments, so this isn't a breaking change to any call site in the repo. `object_index`'s default of 0 stays - it's still within its own 0-255 guard.
2. Both `dm_function_property_write_r_conn` and `dmp_ext_function_property_write_r_conn` now reject `max_apdu_length <= 0` upfront with the same message the six memory/property procedures already use, instead of computing a negative budget and reporting it in a confusing "does not fit" message.

Also rebased onto current `main` per your note - `main` is now an ancestor and the branch carries only its own three commits.

Locally: 4048 passed, ruff/format/mypy clean. New/changed lines: 100% coverage.

farmio:
Claude Code  🤖:

797770f0 closes both nits, and the rebase landed — the PR is `MERGEABLE/CLEAN` against `main` with just its own four commits now, 4047 passing locally, ruff/format/mypy clean. The `max_apdu_length <= 0` guard matches the memory procedures' wording exactly, so the same bad input now reports the same way whichever procedure you reach for.

One thing came out of the `property_id` fix that I don't think was intended.

### The field reorder is a silent breaking change

Making `property_id` required meant moving it ahead of `object_index`, since a dataclass field without a default can't follow one with it. Both fields are plain `int`, so positional construction keeps working and quietly means something else:

```python
# on main
apci.FunctionPropertyCommand(1, 4)   # object_index=1, property_id=4

# on this branch
apci.FunctionPropertyCommand(1, 4)   # object_index=4, property_id=1
                                     # to_knx() -> 02c70401
```

No `TypeError`, no warning — just a different frame on the bus. Nothing inside the repo is affected: I checked every construction site in `xknx/` and `test/` and they all pass keyword arguments. But `xknx.telegram.apci` is public API, and an external caller constructing positionally gets the two swapped with no signal at all.

Two small things to square away:

1. **The changelog documents only half of it.** The Bugfixes entry says "`property_id` has no default anymore", which is true and is the safe half — an omitted `property_id` now raises `TypeError` at construction. The reorder isn't mentioned anywhere, and that's the half that fails silently.
2. **It's filed under `### Bugfixes`.** There's a `### Breaking changes` section in the same Unreleased block, and AGENTS.md asks breaking changes to carry a short before/after snippet — which this one wants more than most, precisely because the failure is invisible rather than an exception.

Worth saying the reorder itself is defensible: it makes the constructor state what `to_knx()` enforces, and it matches what happened to `RestartMasterReset.erase_code`. If you'd rather not touch the signature at all, giving `object_index` a default as well (or keeping `property_id` defaulted to a value the guard accepts) would close the original nit without moving anything. Either way is fine by me — it just shouldn't land as a Bugfix bullet that doesn't mention the swap.

kewde:
Claude Code  🤖:

Fixed at e4f95535 - by removing the reorder instead of documenting it.

Rather than keeping `property_id` required (which is what forced it ahead of `object_index`), I gave it a default that satisfies its own guard: `property_id: int = 1`, the lowest valid value. That closes the original finding the same way - a no-args construction now serializes instead of raising - without moving the fields at all:

```python
# main and this branch, unchanged:
apci.FunctionPropertyCommand(1, 4)   # object_index=1, property_id=4
```

`object_index` stays first, so positional construction means what it always did and there's nothing to file under Breaking changes - I folded the `property_id`-default detail back into the existing Bugfixes bullet instead of adding a second entry.

The real API surface (`dm_function_property_write_r`/`_conn` and their ext siblings) was never affected either way - `property_id` is a required, non-defaulted parameter there, so the raw `apci.FunctionPropertyCommand` default is purely a low-level construction convenience.

Locally: 4048 passed, ruff/format/mypy/pylint clean.

PR #1940, opened 2026-09-03
feat(management): add DMP_LoadStateMachineWrite_RCo_IO

kewde:
## Description

Add `dmp_load_state_machine_write_r_co_io` (KNX v02.01.02 - Management Procedures 03.05.02 - §3.31.3) to write a load event to an interface object's Load State Machine (`PID_LOAD_STATE_CONTROL`, `A_PropertyValue_Write`/`_Response`) and decode the resulting `LoadState` (KNX v01.10.01 - Resources 03.05.01 - §4.23.2.3.1, Table 92: `Unloaded`/`Loaded`/`Loading`/`Error`/`Unloading`/`LoadCompleting`).

An optional `expected_state` polls (re-reading the property) until reached or `poll_timeout` elapses - 30s by default, which is the spec's own minimum wait before a state transition may be considered failed (§4.23.2.1: *"The transitions between the states shall be less than 30 seconds. This time shall be the minimum delay during which a MaC shall wait after sending a load event until interpretation of the load event with the expected state transition to be failed."*) - raising immediately if the machine reports `ERROR`.

Adds the `load_state` module building the fixed 10-octet load events themselves: `start_loading`/`load_completed`/`unload`/`no_operation`, and the "Additional Load Controls" (03h) segment/task allocation subtypes - `alloc_abs_data_seg`, `alloc_abs_stack_seg`, `alloc_abs_task_seg`, `task_ptr`, `task_ctrl_1`, `task_ctrl_2`, `relative_allocation` and `data_relative_allocation` (§3.31.3.1-4).

**Every byte layout and section citation here was verified against the local KNX Standard spec text files** (`03_05_01 Resources v01.10.01 AS.txt`, `03_05_02 Management Procedures v02.01.02 AS.txt`), not derived from any third-party source or guessed - including cross-checking the Load State enum values against the canonical §4.23.2 state-transition table, separate from the abbreviated mention in §3.31.3.5.

100% patch coverage. Non-breaking: all new files/exports.

Stacked on `kewde/dm-function-property-write` (#1939) — last of the stack (`DM_Restart_RCo` → memory read/write → function property → load state machine).

## Type of change

## Checklist


farmio:
Claude Code  🤖:

Reviewed at e96b16a8, diffed against its own base so this is about the single load-state-machine commit. Locally: 4047 passed, ruff/format/mypy clean.

### Verified against the spec

You said every byte layout was checked against the local spec text rather than guessed, so I re-derived the load-bearing ones independently instead of spot-checking:

- **Table 92** (§4.23.2.3.1): Unloaded 0, Loaded 1, Loading 2, Error 3, Unloading 4 (O), LoadCompleting 5 (O). Matches `LoadState` exactly, including which two are optional — the docstring calling out `UNLOADING`/`LOAD_COMPLETING` as optional intermediates tracks the table's own M/O column.
- **Table 93** (§4.23.2.3.2): No Operation 00h, Start Loading 01h, Load Completed 02h, Additional Load Controls 03h, Unload 04h. Matches `_LoadEvent`. Both section numbers are right, which is easy to get wrong here since Table 93 sits under a different sub-clause than Table 92.
- **§4.2.5**: "The write value shall always be 10 octets", and PID = 5, matching `LOAD_EVENT_SIZE` and `ResourceGenericPropertyId.PID_LOAD_STATE_CONTROL`.
- The 30s default and its quoted justification are §4.23.2.1 verbatim.

One thing that falls out nicely and is worth stating: a 10-octet element through `dmp_interface_object_write_r` is `10 + PROPERTY_VALUE_HEADER_OCTETS = 15` — exactly the L_Data_Standard budget, so the default `max_apdu_length` accommodates a load event with nothing to spare. It works, but it's load-bearing rather than coincidental, and nothing currently records that.

### Findings

**1. `LoadState` and the event builders aren't exported.** Only `dmp_load_state_machine_write_r_co_io` reaches `procedures/device/__init__.py` and `procedures/__init__.py`. But the procedure can't be called without them — `event_data` comes from `start_loading()`/`unload()`/`alloc_abs_data_seg()` and `expected_state` takes a `LoadState`. So the public entry point is exported while its entire argument vocabulary is not, and `from xknx.management.procedures.device import load_state` silently binds the *module*. Same gap `ScannedInterfaceObject` had in #1834; `LoadState`, `SegmentType`, `LOAD_EVENT_SIZE` and the eight builders want exporting alongside it.

**2. `load_state.py` is inconsistent with itself about out-of-range input.** The address and length fields go through `.to_bytes(2, "big")`, which raises `OverflowError` for anything above `0xFFFF` (or negative) — while the fields right next to them silently truncate via `& 0xFF`. So `alloc_abs_data_seg(0x10000, 4)` explodes with a bare `OverflowError`, `alloc_abs_data_seg(4, 4, memory_type=0x100)` quietly writes `0x00`, and neither is the `ValueError` the rest of the stack raises. `relative_allocation()` (2 octets) and `data_relative_allocation()` (4 octets) have the same edge. Picking one behaviour — validate and raise `ValueError` — would make the module predictable.

**3. `_decode()` reads `data[0]` without checking the length.** `PID_LOAD_STATE_CONTROL` is `PDT_CONTROL`: 10 octets on write, 1 on read, and `dmp_interface_object_write_r`'s guard only compares `nr_of_elem`, not the data length. A device echoing the written event back would have its *load event type* decoded as a load state — `0x01` (Start Loading) reads as `LOADED`, silently. A `len(data) != 1` check closes that.

**4. Worth a docstring line:** §4.23.2.3.2 says "Illegal additional events (e.g. invalid values written to the Property) shall be ignored and shall not lead to a change of state" and "Unknown events shall be ignored". So a rejected event is indistinguishable from a slow transition — `expected_state` polling is the *only* way a caller finds out, and without it a silently ignored event returns the unchanged state as if it succeeded. That argues for saying so where `expected_state` is documented.

### Stack

This sits on #1939, which has fully diverged from its own base (its merge-base with `kewde/dmp-memory-procedures` is a commit on `main`), so this branch inherits the pre-review copies of #1834 and #1938 too. Nothing to do here beyond rebasing once the branches below are sorted out — the commit itself applies cleanly on its base.

kewde:
Thanks — glad the independent re-derivation of Tables 92/93 and §4.2.5 held up. Fixed in 4eb639bc:

### Rebase
Rebased onto `kewde/dm-function-property-write`'s current tip (used `rebase --onto` targeting the old shared commit directly, since a plain rebase was replaying already-applied history with different content and conflicting). `git merge-base` now equals that branch's HEAD exactly.

### Findings
1. **Missing exports** — `LoadState`, `SegmentType`, `LOAD_EVENT_SIZE` and all twelve load event builders now exported from `xknx.management.procedures` alongside the procedure, same fix as `ScannedInterfaceObject` got in #1834.
2. **Inconsistent out-of-range handling** — added a shared `_uint(value, octets, name)` helper that validates and raises `ValueError` for every field. No more `& 0xFF` truncation or bare `OverflowError` from `int.to_bytes()`.
3. **`_decode()` not checking length** — now requires exactly 1 octet before indexing `data[0]`, closing the "device echoes the 10 octet write event, gets misdecoded as a state" case you described. Added a test for exactly that scenario.
4. **Docstring note on §4.23.2.3.2** — added to `expected_state`'s docs: an illegal/unknown event is silently ignored without a state change, so polling for `expected_state` is the only way to detect that.

### The load-bearing observation
Added as a docstring note on the procedure itself - agreed it's worth stating explicitly rather than leaving it implicit that a load event always exactly fills the L_Data_Standard budget.

100% coverage maintained (32 tests across the two files, up from 25).

